PR #3, opened 2017-09-14
Use Polymer components instead of material-ui

dylanmccall:
This might be a terrible idea.

Happy to discuss rationale and go over details. My thinking with Polymer at the moment is there's a bit of weird code to get it working, but after that it's pretty much smooth sailing.

We should *probably* maintain React wrappers of the Polymer components we want to use (so it's easier to make them all controlled components as needed), but the effort required to add a new UI widget is, so far, happily minimal.


dylanmccall:
I guess a way to think about this is that Polymer is a component system that deals with the DOM API (which we can treat as magic), while React is a component system that sits on top. It can render finer details like maps and content areas where we need, but generally it's responsible for pushing around data and handling big pieces. To me it feels a bit more more like how a desktop UI toolkit is laid out, which is probably why I'm okay with it.

Will throw together a quick alternative with [react-mdc-web](https://www.npmjs.com/package/react-mdc-web) so we can compare approaches. Either approach should be better than what I did initially :)

dylanmccall:
Some more exploration, and the addition of our own light abstraction layer, left me feeling really good about this. It's weird having two component systems, but there's a reasonable separation of concerns, and once we get the ball rolling we probably won't even notice.

PR #27, opened 2018-03-24
Feature/issue 26 create react native app

rasmus-storjohann-PG:


rasmus-storjohann-PG:
@bensinjin please have a look.

bensinjin:
Lookin good! I'm liking the SFC approach and the TS stuff (although I wouldn't mind a run down on some of the syntax that I wasn't clear on). The structure of the models is interesting - is that something you came up with? Would you be able to elaborate on what exactly the range of responsibilities you would see a particular model having?

PR #52, opened 2018-04-20
[WIP] TSLint fixes

tnightingale:
**Not intended to be merged.**

Changes include various fixes to tslint errors and a fix to an invalid tslint configuration that I've made as I'm familiarizing myself with the codebase and architecture / components.

I have pushed this branch and created this pull request as a point of discussion rather than expecting it to be merged (though I suspect the changes will eventually be made on master anyway).

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,0 +1,30 @@
+# Getting started
+
+
+## Initial setup and build
+
+With a fresh repository, this will get the application compiled and running on Expo:
+
+    // Install dependencies
+    npm install
+    // Compile TypeScript
+    npm run tsc
+    // Run tests
+    npm run jest
+    // Build & start Expo
+    npm start
+    // Follow instructions to open app in Expo
+
+
+## Developing
+
+In seperate processes you will want to run the following:
+
+    // Watch filesystem and recompile TypeScript on changes
+    npm run tsc -- --watch
+
+    // Watch filesystem and rerun tests on changes
+    npm run jest -- --watch
+
+    // Watch filesystem and rebuild app / reload Expo on changes
+    npm start
--- a/src/stores/__tests__/counter.test.ts
+++ b/src/stores/__tests__/counter.test.ts
@@ -63,7 +63,7 @@ describe('the reducer', () => {
         const theStore = buildStore();
         const theAction = {
             type: constants.SET_COUNTER as typeof constants.SET_COUNTER,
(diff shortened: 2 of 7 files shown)


src/stores/__tests__/counter.test.ts
```diff
@@ -63,17 +63,15 @@ describe('the reducer', () => {
...
         const theNewStore = counter.reducer(theStore, theAction);
         expect(theNewStore.value).toBe(theAction.payload.value);
     });
 
     it('when called with RESET_COUNTER should return store with value of zero', () => {
         const theStore = buildStore();
-        const theAction = {
-            type: constants.RESET_COUNTER as typeof constants.RESET_COUNTER
-        };
+        const theAction = { type: constants.RESET_COUNTER as typeof constants.RESET_COUNTER };
```
tnightingale: This was a funny one; a point of contention between the `singleline` & `multiline` options in the `trailing-comma` tslint rule.

rasmus-storjohann-PG: One hesitation I have with Prettier is that it's all or nothing ... when I disagree with it, Prettier is gonna win. I can see that being a feature in a big team, but for us, I feel tslint with it's ability to turn off a warning on a case by case basis is a better fit.

INSTALL.md
```diff
@@ -0,0 +1,30 @@
...
+In seperate processes you will want to run the following:
+
+    // Watch filesystem and recompile TypeScript on changes
+    npm run tsc -- --watch
+
+    // Watch filesystem and rerun tests on changes
+    npm run jest -- --watch
+
+    // Watch filesystem and rebuild app / reload Expo on changes
+    npm start
```
rasmus-storjohann-PG: Looks good, except that I prefer to use yarn, and my preference would be that that is reflected in our docs.

tnightingale: Ya fair enough, I had this written before I 'clicked' (see: my comment about package-lock.json) and realized you were using Yarn. I will make sure to update accordingly.

INSTALL.md
```diff
@@ -0,0 +1,30 @@
+# Getting started
+
+
```
trisapeace: Hi @tnightingale ,
This is great! Very helpful. I'm working my way through it now.
Could you please add a section for pre-reqs with links to download pages, if possible?

Issue #37, opened 2018-04-17
Language switcher component


rasmus-storjohann-PG:
AC: Button component for toggling between two languages, include state where none or one language has been selected. The button just needs to work to switch the language, it needs not adhere to our visual design, that work will be done under #39

Issue #32, opened 2018-04-17
Define language state machine


rasmus-storjohann-PG:
Define language settings state machine and related data model for selecting up to two languages and toggle the UI between them. Need to handle state where the user only wants to select one language e.g. for users who just want to use English or French

There is a similar story for task states, see #86.

PR #55, opened 2018-04-28
Feature/31 Internationalization infrastructure

tnightingale:
Contains formatting changes/fixes found in #52 (and probably also overlaps a little with #53).

Adds [jsLingui](https://lingui.github.io/js-lingui/index.html) as a dependency. I had some dependency conflicts when initially integrating this library but pinning to a slightly older version (2.0.1 instead of latest 2.0.7 at time of writing) resolved this. This will likely be resolved on its own so we should just keep an eye out for new releases.

Nothing in the way of tests just yet as this is 99% integration of 3rd-party lib. The language switcher UI and language state are things that we will want coverage on. The implementations for these in this branch are just throw-away demos using Component state. Work on real implementations (& appropriate tests) should be continued in #37 and #32.


tnightingale:
@rasmus-storjohann-PG, @bensinjin FYI I've rebased this to include Rasmus' routing changes.

For better or worse, as can be seen in the examples in this pull request; i18n adds a fair amount of boilerplate & ceremony. Some of it is optional and some of it is entirely subjective, but the point is we now have a wide range of tools at our disposal for handling the nuances of multi-lingual applications. 

Looking forward to discussing this and figuring out what we should / should not use, etc.

locale/ar/messages.json
```diff
@@ -0,0 +1,91 @@
...
+    "origin": [
+      [
+        "lib/hello_world/hello_world_container.js",
+        18
+      ]
+    ]
+  },
+  "Number format:": {
+    "translation": "[AR] Number format:",
+    "origin": [
```
tnightingale: While useful for developers, this `origin` metadata is likely more confusing than helpful to translators. I'm pretty sure jsLingui has a 'minimal' output format that can be configured (it's default in the newest version of jsLingui).

locale/ar/messages.json
```diff
@@ -0,0 +1,155 @@
...
+  "Explore all": {
+    "translation": "[AR] Explore all",
+    "origin": [
+      [
+        "lib/components/navigation_bar/navigation_bar.js",
+        12
+      ]
+    ]
+  },
+  "{0, plural, zero {no (zero) items} one {# (one) item} two {# (two) item} few {# (few) items} many {# (many) items} other {# (other) items} =5 {There (#N) are exactly # items}}": {
```
tnightingale: This plural syntax is probably the most complicated/convoluted example of the CLDR/MessageFormat syntax and I've deliberately included every option in this example for demonstration. It's not surprising as MessageFormat claims to support _all possible formats required by every language_ so it's bound to be messy.

We should evaluate how much & what parts of it we need based on the subset of languages we want to support.

src/application/index.tsx
```diff
@@ -5,20 +5,52 @@ import { ConnectedNavigationBar } from '../components/navigation_bar/connected_n
...
+
+    _setLang(langCode: string): void {
+        const rtl = langCode in RTL_LANGS;
+        // In theory, I18nManager.forceRTL() will tell all RN & Expo components
+        // to render RTL.
+        // TODO: Determine whether the native bits are needed if we're setting
+        // this manually (https://facebook.github.io/react-native/blog/2016/08/19/right-to-left-support-for-react-native-apps.html#making-an-app-rtl-ready)
+        I18nManager.forceRTL(rtl);
+        // Also store in application state
+        this.setState({langCode: langCode, rtl});
```
rasmus-storjohann-PG: I couple of things here: I don't mind prototype level code in the code base, but ideally it should be pulled out into its own file, so that core files like `application/index.tsx` don't accrue tech debt. (Note that I'm reviewing the commits in order, so if you fix this later in the PR, then that's fine)

Secondly, I would like to avoid the `_` prefix for "private" functions, since they are not really private. I prefer the pattern of putting private code in its on file, see https://github.com/pg-irc/pathways-backend/tree/7614f2788441147fc2369e3588dbdc479d4a6ae0/common/testhelpers , here private means "for use in this module only, do not import from outside".

src/stores/message.ts
```diff
@@ -11,7 +13,7 @@ export const setMessage = (message: string) => (
 
 // tslint:disable-next-line:typedef
 const buildDefaultStore = () => (
-    { message: 'default text' }
+    { message: i18n.t`Default text` }
```
rasmus-storjohann-PG: On the issue of wrapping the i18m library, how hard would it be to have this be
```
{ message: ourI18NWrapper('Default text') }
```
or
```
{ message: ourI18NWrapper`Default text` }
```

src/hello_world/hello_world_container.tsx
```diff
@@ -7,8 +7,11 @@ import * as message from '../stores/message';
 import { I18n } from '@lingui/core';
 import { withI18n } from '@lingui/react';
 
-export interface Props {
+export interface I18nProps {
     i18n: I18n;
+}
```
rasmus-storjohann-PG: Why does this need to be exported?

locale/ar/messages.js
```diff
@@ -1 +1 @@
...
\ No newline at end of file
+module.exports={languageData:{"plurals":function(n,ord){var s=String(n).split("."),t0=Number(s[0])==n,n100=t0&&s[0].slice(-2);if(ord)return"other";return n==0?"zero":n==1?"one":n==2?"two":n100>=3&&n100<=10?"few":n100>=11&&n100<=99?"many":"other"}},messages:{"{0, plural, one {# item} other {# items} =3 {There are exactly # items}}":function(a){return["[AR] ",a("0","plural",{one:["#"," item"],other:["#"," items"],3:["There are exactly ","#"," items"]})]},"Number format:":"[AR] Number format:","Current date:":"[AR] Current date:","Default text":"[AR] Default text","The value is {value}":function(a){return["[AR] The value is ",a("value")]},"The double value is {double}":function(a){return["[AR] The double value is ",a("double")]},"Hello {name}!":function(a){return["[AR] Hello ",a("name"),"!"]},"Increment":"[AR] Increment","Decrement":"[AR] Decrement"}};
```
rasmus-storjohann-PG: Is there a non-minified version of this file somewhere? How is it generated?

locale/ar/messages.json
rasmus-storjohann-PG: Sounds good

src/components/hello_world/hello_world.tsx
```diff
@@ -21,33 +28,52 @@ export interface Actions {
...
-    const { navigationBarInProps, counterInProps, messageInProps, increment, decrement,
+export const HelloWorldContainer: React.StatelessComponent<I18nProps & Props & Actions> = withI18n()((props: I18nProps & Props & Actions): JSX.Element => {
+    const { i18n,
+        navigationBarInProps, counterInProps, messageInProps, increment, decrement,
         goBack, goForwards,
-        pushUserWithUrl, pushUserWithId, setMessage }: Props & Actions = props;
+        pushUserWithUrl, pushUserWithId, setMessage }: I18nProps & Props & Actions = props;
+
+    const incText = i18n.t`Increment`;
+    const decText = i18n.t`Decrement`;
```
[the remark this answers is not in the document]
rasmus-storjohann-PG: Yeah, we can tidy this up as we go

src/components/navigation_bar/navigation_bar.tsx
```diff
@@ -1,34 +1,41 @@
...
+import { I18n } from '@lingui/core';
+import { withI18n } from '@lingui/react';
+
 interface TabButtonProp {
-    onPress: () => void,
-    title: string,
-};
+    onPress: () => void;
+    title: string;
+}
```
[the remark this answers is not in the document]
rasmus-storjohann-PG: Somehow it seems to not complain as much as it should on my end. One issue was that it was ignoring tsx files, but that has been fixed. I'll keep an eye out for what might be going on there

package-lock.json
```diff
@@ -5,18 +5,18 @@
   "requires": true,
   "dependencies": {
     "@babel/code-frame": {
-      "version": "7.0.0-beta.41",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.0.0-beta.41.tgz",
-      "integrity": "sha512-omQT0n9EW38xfMCY7cCW/PAT55igUX9c9cMq6QN4EjxCcelcrocwMJ6H4JP4BGrQ+OHdJAQPM9/Eaa2Yce4Aug==",
+      "version": "7.0.0-beta.44",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.0.0-beta.44.tgz",
+      "integrity": "sha512-cuAuTTIQ9RqcFRJ/Y8PvTh+paepNcaGxwQwjIDRWPXmzzyAeCO4KqS9ikMvq0MCbRk6GlYKwfzStrcP3/jSL8g==",
```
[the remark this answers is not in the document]
tnightingale: Ya I don't have a preference. npm and yarn are essentially feature equivalent. The lock files really only become important when producing consistent builds.

However I recently read that latest npm release has made some major performance improvements on CI builds using its lock file. I'll try dig up that info.

locale/ar/messages.js
tnightingale: They're generated from their equivalent messages.json file. The compiled js contains some inlined optimizations for handling the complexities of pluralization, number/date formatting etc at runtime.

tnightingale: Generated with the `lingui compile [--strict]` command (`--strict` will complain/fail if there's missing/untranslated messages in the source json.

src/hello_world/hello_world_container.tsx
tnightingale: It probably doesn't.

rasmus-storjohann-PG:
Note that quite a few of my comments were on code that was later changed, so have a look through "hidden conversations" to see if any of those thoughts still make sense.

locale/ar/messages.js
rasmus-storjohann-PG: Thanks, that makes sense

src/stores/message.ts
tnightingale: We could though I'm not sure why we would want to?

One thing that would complicate a custom wrapper is that the i18n object relies on lang/locale context that is configured at the top of the React tree (see `<I18nProvider />` and the `withI18n()` HOC. This context is what it uses to determine the language/locale and any custom formatting overrides we might have configured. 

It's also worth noting is that these contexts can be nested and overridden for different parts of the tree.

package.json
```diff
@@ -35,6 +37,7 @@
     "lint": "tslint --rules-dir node_modules/tslint-immutable/rules/ src/**/*.ts src/**/*.tsx",
     "tsc": "tsc",
     "jest": "jest",
+    "lingui": "lingui",
```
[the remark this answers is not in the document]
tnightingale: Yeah I considered this; there are two main commands:

 - `extract`: Updates the messages.json file for all locales
 - `compile`: generates the optimized messages.js files/modules from the messages.json files that get loaded at runtime.

There is also `add-locale` which will create a messages.json file for a new locale but this will only be used once per locale we add.

I'm not sure if splitting `extract` and `compile` into their own yarn/npm script commands really improves convenience. However I suppose it would potentially improve discoverability/documentation from the perspective of someone new to this project.

rasmus-storjohann-PG:
Also need to get this building without errors or warnings. In order to really benefit from TS we need zero tolerance for build warnings and errors.

src/@types/declarations.d.ts
```diff
@@ -0,0 +1,3 @@
+// Dummy declarations
+declare module '@lingui/*';
+declare module '../../locale/*/messages';
```
tnightingale: Dealing with untyped 3rd party code with a combination of [shorthand module declarations](https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#shorthand-ambient-module-declarations) and [wildcard character in module names](https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#wildcard-character-in-module-names).

trisapeace:
Hi Tom,
Is this the error that you mentioned about jslingui not having TS definitions?
```
$ yarn tsc
/c/Program Files (x86)/Yarn/bin/yarn: line 3: readlink: command not found
yarn run v1.6.0
$ tsc
src/application/index.tsx(10,37): error TS7016: Could not find a declaration file for module '@lingui/react'. 'c:/Users/Theresa/PeaceGeeks/pathways-frontend/node_modules/@lingui/re
act/index.js' implicitly has an 'any' type.
  Try `npm install @types/lingui__react` if it exists or add a new declaration (.d.ts) file containing `declare module 'lingui__react';`
src/application/index.tsx(11,24): error TS7016: Could not find a declaration file for module '../../locale/en/messages'. 'c:/Users/Theresa/PeaceGeeks/pathways-frontend/locale/en/me
ssages.js' implicitly has an 'any' type.
src/application/index.tsx(12,24): error TS7016: Could not find a declaration file for module '../../locale/ar/messages'. 'c:/Users/Theresa/PeaceGeeks/pathways-frontend/locale/ar/me
ssages.js' implicitly has an 'any' type.
src/application/index.tsx(33,56): error TS2339: Property 'langCode' does not exist on type 'Readonly<{}>'.
src/application/index.tsx(34,25): error TS2322: Type '{ children: Element; }' is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes<Component<Pick<I18nProps & P
rops & Actions, "i18n"...'.
  Type '{ children: Element; }' is not assignable to type 'Readonly<Pick<I18nProps & Props & Actions, "i18n">>'.
    Property 'i18n' is missing in type '{ children: Element; }'.
src/application/index.tsx(35,50): error TS2322: Type '{ language: any; }' is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes<Component<Pick<I18nProps & Props
 & Actions, "i18n"...'.
  Type '{ language: any; }' is not assignable to type 'Readonly<Pick<I18nProps & Props & Actions, "i18n">>'.
    Property 'i18n' is missing in type '{ language: any; }'.
src/application/index.tsx(35,71): error TS2339: Property 'langCode' does not exist on type 'Readonly<{}>'.
src/components/hello_world/components.tsx(4,23): error TS7016: Could not find a declaration file for module '@lingui/react'. 'c:/Users/Theresa/PeaceGeeks/pathways-frontend/node_mod
ules/@lingui/react/index.js' implicitly has an 'any' type.
  Try `npm install @types/lingui__react` if it exists or add a new declaration (.d.ts) file containing `declare module 'lingui__react';`
src/components/hello_world/hello_world.tsx(8,22): error TS7016: Could not find a declaration file for module '@lingui/core'. 'c:/Users/Theresa/PeaceGeeks/pathways-frontend/node_mod
ules/@lingui/core/index.js' implicitly has an 'any' type.
  Try `npm install @types/lingui__core` if it exists or add a new declaration (.d.ts) file containing `declare module 'lingui__core';`
src/components/hello_world/hello_world.tsx(9,67): error TS7016: Could not find a declaration file for module '@lingui/react'. 'c:/Users/Theresa/PeaceGeeks/pathways-frontend/node_mo
dules/@lingui/react/index.js' implicitly has an 'any' type.
  Try `npm install @types/lingui__react` if it exists or add a new declaration (.d.ts) file containing `declare module 'lingui__react';`
src/components/navigation_bar/navigation_bar.tsx(4,22): error TS7016: Could not find a declaration file for module '@lingui/core'. 'c:/Users/Theresa/PeaceGeeks/pathways-frontend/no
de_modules/@lingui/core/index.js' implicitly has an 'any' type.
  Try `npm install @types/lingui__core` if it exists or add a new declaration (.d.ts) file containing `declare module 'lingui__core';`
src/components/navigation_bar/navigation_bar.tsx(5,26): error TS7016: Could not find a declaration file for module '@lingui/react'. 'c:/Users/Theresa/PeaceGeeks/pathways-frontend/n
ode_modules/@lingui/react/index.js' implicitly has an 'any' type.
  Try `npm install @types/lingui__react` if it exists or add a new declaration (.d.ts) file containing `declare module 'lingui__react';`
src/components/navigation_bar/navigation_bar.tsx(33,26): error TS2339: Property 'children' does not exist on type 'I18nProps & Props & Actions'.
src/stores/message.ts(4,22): error TS7016: Could not find a declaration file for module '@lingui/core'. 'c:/Users/Theresa/PeaceGeeks/pathways-frontend/node_modules/@lingui/core/ind
ex.js' implicitly has an 'any' type.
  Try `npm install @types/lingui__core` if it exists or add a new declaration (.d.ts) file containing `declare module 'lingui__core';`
error Command failed with exit code 2.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

INSTALL.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Internationalization (i18n)
+
+jsLingui provides some commands to help with i18n:
+
+    // Add a new locale(s) (generate empty messages catalog(s) for each locale)
+    yarn lingui add-locale [locales...]
+
+    // Extract all strings from codebase into catalogs (per-locale messages.json files)
+    yarn lingui extract [files...]
```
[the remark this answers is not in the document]
trisapeace: Also interesting to note is that by default, the `yarn lingui extract` command doesn't clean up unused strings. To do that, we'll need to use `yarn lingui extract --clean` (https://lingui.github.io/js-lingui/tutorials/cli.html#cleaning-up-obsolete-messages). I think that is is better to use `--clean` every time. Even if we accidentally clean up something that we want to add back later, we should be able to find it in source control. What do you think?

INSTALL.md
```diff
@@ -0,0 +1,44 @@
...
+In seperate processes you will want to run the following:
+
+    // Watch filesystem and recompile TypeScript on changes
+    yarn tsc --watch
+
+    // Watch filesystem and rerun tests on changes
+    yarn jest --watch
+
+    // Watch filesystem and rebuild app / reload Expo on changes
+    yarn start
```
trisapeace: Did you mean to remove the `yarn lint` command that is currently in the `README.md`?

.babelrc
```diff
@@ -1,5 +1,5 @@
 {
-  "presets": ["babel-preset-expo"],
+  "presets": ["babel-preset-expo", "@lingui/babel-preset-react"],
```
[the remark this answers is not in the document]
tnightingale: Babel will look for config in both package.json or .babelrc files. The current layout is just whatever comes out of the box with create-react-native-app. In my experience it's better to keep these in separate files (along with separate eslint/tslint, tsconfig, etc) as package.json can get quite unwieldy.

INSTALL.md
```diff
@@ -0,0 +1,44 @@
...
+
+
+## Internationalization (i18n)
+
+jsLingui provides some commands to help with i18n:
+
+    // Add a new locale(s) (generate empty messages catalog(s) for each locale)
+    yarn lingui add-locale [locales...]
```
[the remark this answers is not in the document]
tnightingale: I don't know, this is copy+pasted from the jsLingui docs. However it's not a command that will be run very frequently - at most, once per supported locale to initially set it up.

INSTALL.md
```diff
@@ -0,0 +1,44 @@
...
+
+    // Extract all strings from codebase into catalogs (per-locale messages.json files)
+    yarn lingui extract [files...]
+
+    // Compile messages catalogs into optimized & importable javascript modules
+    yarn lingui compile --strict
```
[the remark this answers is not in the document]
tnightingale: Can certainly add the compile step to Travis' build steps. Whether or not to commit the compiled `messages.js` is up for debate. There's benefit to having them committed as it is one less step required to produce a build. Once our application strings stabilize, these compiled files will not change much.

App.js
```diff
@@ -1,3 +1,11 @@
+// Unfortunately we need to load the Intl polyfill for Android. iOS has Intl
+// available natively; we should see it is possibly conditionally load the
+// polyfill only when needed.
+import 'intl';
+// Include Unicode CLDR locale data for each of our locales
+import 'intl/locale-data/jsonp/en.js';
+import 'intl/locale-data/jsonp/ar.js';
```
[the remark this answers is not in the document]
tnightingale: Yep! Fwiw `INSTALL.md` only exists because `README.md` was full of CRNA boilerplate. We likely want to merge the two (I think Rasmus' pull request already did this).

src/components/navigation_bar/connected_navigation_bar.ts
```diff
@@ -4,8 +4,7 @@ import { Props, Actions, NavigationBar } from './navigation_bar';
 import { Store } from '../../application/store';
 import * as store from '../../stores/navigation_bar';
 
-const mapStateToProps = (): Props => ({
-});
+const mapStateToProps = (): Props | void => {};
```
rasmus-storjohann-PG: Shouldn't this be `const mapStateToProps = (): Props? => {};`

src/components/navigation_bar/navigation_bar.tsx
```diff
@@ -13,11 +13,12 @@ const TabButton: React.StatelessComponent<TabButtonProp> = ({ onPress, title }:
     <Button onPress={onPress} title={title} />
 );
 
-export interface I18nProps {
+interface I18nProps {
     i18n: I18n;
 }
 
 export interface Props {
+    children: React.ReactChildren;
```
rasmus-storjohann-PG: This should be 
```
children?: any
```
see https://github.com/Microsoft/TypeScript/issues/6471

package.json
```diff
@@ -58,11 +58,10 @@
   },
   "dependencies": {
     "@lingui/react": "2.0.1",
-    "@types/react-native": "^0.52.20",
     "expo": "^27.0.0",
     "history": "^4.7.2",
     "intl": "^1.2.5",
-    "react": "16.3.2",
+    "react": "16.3.1",
```
rasmus-storjohann-PG: Thanks, I'll start worrying more about the peer dependency warnings.

INSTALL.md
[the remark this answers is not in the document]
tnightingale: Thanks for the `--clean` tip! I hadn't seen that option.

There are a couple of alternative workflows for creating and managing `obsolete` messages  (the obsolete flag currently gets set when a string changes and a new/replacement message is added). 

Currently our messages are keyed/identified by their base language (English) value. This means when we modify the message in the base language it creates a _new_ message and marks the old one (and any existing translations) as obsolete. I think keeping obsolete translations around might be beneficial until we get a chance to update the translation on the new version.

An alternative approach is to key/identify messages with message ids (not the message value itself). This means any modifications in any language will not produce new messages which, in turn, require translation.

locale/ar/messages.json
```diff
@@ -0,0 +1,155 @@
+{
```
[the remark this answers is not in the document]
tnightingale: That's strange, I'm curious as to what the differences are. My suspicion is that it's just the `origin` property that is changing. I commented earlier about removing that metadata from our locale messages.json files.

INSTALL.md
tnightingale: I suspect `INSTALL.md` can now be decommissioned in favour of `README.md`.

INSTALL.md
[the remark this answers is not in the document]
rasmus-storjohann-PG: Generally I prefer to not check in generated files. In this case, I think it will make our build process more reliable if it always does everything and we don't have to jump through special hoops when strings change.

locale/ar/messages.json
rasmus-storjohann-PG: Depends on whether translators will be working directly with the JSON or whether there is some tool that makes their experience better --- if the latter, leaving the origin in should not affect them. We don't need to decide this now.

package.json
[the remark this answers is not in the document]
rasmus-storjohann-PG: The compile step could just be incorporated into a general `build` step that does `tsc` as well.

rasmus-storjohann-PG:
@trisapeace Thanks so much for taking the time to look at the PR!

tnightingale:
Ok, I've moved the locale state into Redux and made the language switcher a dedicated component. Note the addition of the application/locales.ts file for setting up all i18n bits in one place. 

All but one TypeScript error/warning has been fixed. The last one is a variation on an earlier one and it has me stumped:
```
src/application/index.tsx(16,36): error TS2339: Property 'catalogs' does not exist on type 'IntrinsicAttributes & IntrinsicClassAttributes<Component<Pick<Props & Actions, never>, ComponentS...'.
```
I'm passing an extra prop to a Redux-connected component from the outside. Afaik this is a totally valid thing to do but I don't know how to communicate this to TypeScript; the Redux/connect TS declarations are externally implemented/maintained.

bensinjin:
This is looking good to me. I have reviewed the discussions and don't really have anything to add at this point.

package.json
```diff
@@ -3,13 +3,15 @@
   "version": "0.1.0",
   "private": true,
   "devDependencies": {
+    "@lingui/babel-preset-react": "2.0.1",
+    "@lingui/cli": "2.0.1",
     "@types/enzyme": "^3.1.9",
     "@types/enzyme-adapter-react-16": "^1.0.2",
     "@types/history": "^4.6.2",
     "@types/jest": "^22.2.2",
     "@types/react": "^16.0.40",
```
trisapeace: Yay! Finally got it to work locally for me.
I had to re-install Expo on my phone.
But I also updated the `react` and `react-native` versions here to match the ones in `dependencies`. Do you think that that is necessary too?

tnightingale: > But I also updated the react and react-native versions here to match the ones in dependencies. Do you think that that is necessary too?
I'm not sure I follow, you had to update the `@types/react` and `@types/react-native` to match the `react` and `react-native` versions to get this to work?

tnightingale:
After a digging through the Redux & React TS definitions and their respective issue trackers I've found many examples of people experiencing issues with the type declarations for redux's `connect()`. Unfortunately I think the dynamic nature of the function means that it's very hard/impossible to provide a declaration that covers every use-case.

My solution to the error mentioned above is  to pass the catalogs prop to the component via `mapStateToProps` instead of passing it from the outside (eg: `<ConnectedI18nProvider catalogs={catalogs} /> // Causes TS error`). This change can be seen here: 9df57af.

@rasmus-storjohann-PG @bensinjin @trisapeace I'd like to see this PR merged, could you guys give it another spin?

src/application/locales.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export const locales = [
+    { code: 'en', label: 'English', catalog: enMessages },
+    { code: 'ar', label: 'Arabic', catalog: arMessages },
+];
+
+export const catalogs = locales.reduce((all: {[index: string]: object}, locale: Locale): object => {
+    all[locale.code] = locale.catalog;
+    return all;
+}, {});
```
rasmus-storjohann-PG: So this may be because I'm new to JS/TS, but this code makes my head hurt. I really like single responsibility to also apply to lines of code, to have code that does one thing at a time. If that means pulling out a couple of helper functions, that's fine, or just using more local variables with clear names to explain what's going on. Tom, I know you just want to get this PR in, but since this is your first major PR I think it's important to use it to align our styles and establish what our code base should look like.

src/application/test_helpers/random_test_values.ts
```diff
@@ -1,7 +1,11 @@
...
     return 1000 * Math.random();
 };
 
-export const anInteger = () => {
+export const anInteger = (): number => {
     return Math.floor(aNumber());
 };
+
+export const aLocaleCode = (): string => {
+    return 'ab';
```
rasmus-storjohann-PG: The idea of these functions is that they return random strings to improve test coverage. I will be introducing `uuid` as a library, so if you like you can use that and just grap the two first characters of the random string. Following the pattern of the server side, I will also implement being able to run the tests with a seed value so we can repeat tests that are failing intermittently based on the random values.

src/components/language_switcher/language_switcher.tsx
```diff
@@ -0,0 +1,27 @@
...
+    locales: Array<Locale>;
+}
+
+export interface Actions {
+    setLocale(langCode: string): void;
+}
+
+export const LanguageSwitcher = ({setLocale, currentLocale, locales}: Props & Actions): JSX.Element => {
+    return (
+        <View style={{ alignItems: 'center' }}>
```
rasmus-storjohann-PG: Really picky, subject to discussion in the team, so no need to change now, but in my code I am preferring the
```ts
... JSX.Element => (
        <View style={{ alignItems: 'center' }}>....
)
```
syntax, but always with the line break after the `(` to make it look like a function.

src/components/navigation_bar/navigation_bar.tsx
```diff
@@ -1,33 +1,34 @@
...
-    goToExplore: () => void,
-};
+    goToQuestionnaire: () => void;
+    goToPlan: () => void;
+    goToExplore: () => void;
+}
 
-export const NavigationBar: React.StatelessComponent<Props & Actions> = (props) => {
-    const { goToQuestionnaire, goToPlan, goToExplore } = props;
+export const NavigationBar: React.StatelessComponent<I18nProps & Props & Actions> = (props: I18nProps & Props & Actions): JSX.Element => {
```
rasmus-storjohann-PG: Generally no fan of inheritance, I am thinking that maybe making `I18nProps` the base class of all our `Props` interfaces (which I think TS can do) could make sense.

tnightingale: Hmm a base class would also work, I guess it depends on how prevalent the `withI18n()` HOC will be used.

If we're only using it ~25% of the time then we probably don't want all of our props to require an i18n prop. I guess we could use a nullable type in the base class, but then we have to litter code with `if (props.i18n !== null) ...` (it would break the convenient `const { i18n, propOne, propTwo }: MyCustomPropsType = props;` syntax).

Sounds like the noise/verbosity is the problem here, might be better solved by aliasing the Union/Intersection type, eg: 
```typescript
type CustomComponentType = I18nProps & Props & Actions
export const NavigationBar: React.StatelessComponent<CustomComponentType> = (props: CustomComponentType): JSX.Element => {
```

src/application/locales.ts
tnightingale: Not sure what you're specifically objecting to here but I'm guessing the glaring thing is TypeScript's [indexable types syntax](http://www.typescriptlang.org/docs/handbook/interfaces.html#indexable-types).

In this case it is necessary otherwise tsc/tslint complains about not being able to infer the index's type on line 24. It could be rewritten as the following:

```typescript
type Catalogs = {[index: string]: object};
export const catalogs = locales.reduce((all: Catalogs, locale: Locale): object => {
```
Unfortunately, regardless of the syntax & aliasing/formatting, this type declaration isn't much better than using `any` as the individual catalog objects (which, coincidentally, are also of type `{[index: string]: object}`) are a jsLingui construct and we can't do any better without writing/maintaining our own declarations.

src/components/navigation_bar/navigation_bar.tsx
rasmus-storjohann-PG: Yeah, the alias could work. I'm guessing that `withI18n()` will be used very widely, all non-trivial and most trivial components will contain translated text, which is why I like the inheritance idea. We should avoid null whenever possible, so I don't think we'd want to go with the nullable option.

rasmus-storjohann-PG: You don't need to make this change now, though...

src/components/language_switcher/language_switcher.tsx
rasmus-storjohann-PG: Oh, that's a good parallel to the `{}` in `if/else/for`. I always use `{}`.  Hmmm, that makes me want to prefer the form with the return now...

src/application/test_helpers/random_test_values.ts
rasmus-storjohann-PG: Leave the function there, it's totally the right approach, just as long as it does the right thing.

src/application/locales.ts
rasmus-storjohann-PG: For me, this code is complex, because I have to parse a non-trivial function implementation in order to understand what this module exports. That makes my mind straddle two completely different subject areas: What does this function do, and what does this module do. I don't want to have to think about both of those things at once. My suggestion is to go with something like this:
```
type Catalogs = {[index: string]: object};

export const catalogs = buildCatalog(locales);

const buildCatalog(locale: Locale): Catalogs => {
   // reduce call with some local variables to make the various pieces easier to follow
};
```

src/components/navigation_bar/navigation_bar.tsx
rasmus-storjohann-PG: And one more thing :-) I'm happy with how you're passing in the i18n through a prop rather than though `connect()`, it's pretty clear what's going on.

tnightingale: > I'm guessing that withI18n() will be used very widely, all non-trivial and most trivial components will contain translated text, which is why I like the inheritance idea.

I think you're misunderstanding; most of the time the [React components](https://lingui.github.io/js-lingui/ref/lingui-react.html#components) `<Trans></Trans>`, `<Plural></Plural>`, `<NumberFormat />` and `<DateFormat />` will be used. These implicitly retrieve the `I18n` object from React's context, behind the scenes.

The issue arises when there is a runtime-check requiring a `string` and not a instance of `React.Component` (one example is passing a prop to a React Native component like `<Button title={mustBeAString} />`). This should not be too common. In these cases you can access jsLingui's "lower-level" API via the `I18n` context object itself. JsLingui provides the helper `withI18n()`, a HOC to pull this out of React's context in a safe way and pass it to the child Component as a prop.

The problem with a base class for all Prop objects is that it would make *every* component require a I18n object to be passed via props (assuming we don't make it nullable, which would be silly as it defeats the purpose).

src/application/test_helpers/random_test_values.ts
```diff
@@ -4,8 +4,4 @@ export const aNumber = () => {
 
 export const anInteger = () => {
     return Math.floor(aNumber());
-};
-
-export const aLocaleCode = (): string => {
-    return 'ab';
-};
+};
```
rasmus-storjohann-PG: Sorry, I wasn't clear on this, I would prefer if you kept the function, with the current implementation as a little bit of tech debt, we should fix it to return a random string, but that can wait. The approach to testing using little helper functions like this is very important for making readable tests.

Also, wrong issue number is commit message.

package.json
trisapeace: I'm not sure. I made the mistake of doing two things and now I'm not sure which one worked. I updated `@types/react` and `@types/react-native` at the same time as I decide to uninstall and reinstall Expo on my phone. So I'm not sure which one helped.

What are these `@types/...` dependencies used for? Shouldn't the version numbers be in sync with the other dependencies further down?

package.json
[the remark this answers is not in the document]
trisapeace: The general `build` step sounds like a great idea.

src/stores/__tests__/locale.test.ts
```diff
@@ -0,0 +1,51 @@
...
+describe('the reducer', () => {
+    it('should default to build a store with default locale code `en`', () => {
+        const theStore = locale.reducer();
+        expect(theStore.code).toBe('en');
+    });
+
+    it('when called with SET_LOCALE should return store with locale code from action', () => {
+        const theStore = buildStore();
+        const theAction = {
+            type: constants.SET_LOCALE as typeof constants.SET_LOCALE,
```
trisapeace: This seems rather bulky. Will we need to use `as typeof` every time we use one of these constant values?

src/stores/__tests__/locale.test.ts
```diff
@@ -0,0 +1,51 @@
...
+import * as helpers from '../../application/helpers/redux-helpers';
+
+let buildStore = (): locale.Store => locale.reducer(undefined, undefined);
+
+let buildStoreWithLocale = (code: string): locale.Store => {
+    const action = helpers.makeAction(constants.SET_LOCALE, { code });
+    return locale.reducer(undefined, action);
+};
+
+describe('the action for', () => {
```
trisapeace: I like the way you structure your tests. Very clear.

src/stores/__tests__/locale.test.ts
rasmus-storjohann-PG: It is bulky, but it helps TS do something a little magical, to use the value of the type field to infer the type of the payload. I wrote up something about it here https://github.com/rasmus-storjohann/ReduxTypescript/blob/master/dry-redux-with-typescript.md#discriminated-unions-for-types.

package.json
rasmus-storjohann-PG: @types are type definitions for JS libraries, they allow TS to work with JS libraries and still understand their types. Some libraries include their own @types, others need the @types to be installed as their own packages. I don't know if there is a convention around keeping their version numbers in sync, you'd think that would make sense, but were in JS land, so....

rasmus-storjohann-PG:
@tnightingale still one thing outstanding: Can you fix your commit messages so they refer to the correct issue? Some of them have #55 (which is the PR number, not the issue number) and the earliest ones in the PR don't have an issue number at all. Having these in place is very valuable in the future to be able to trace the history of changes in the code base.

I find github code review UX to be very frustrating. This is such a bummer, it was totally fine a year ago, they've totally messed it up. We need to talk about how to use github for reviews.

PR #58, opened 2018-05-03
Feature/issue 41 questionnaire component

rasmus-storjohann-PG:


rasmus-storjohann-PG:
Hey @bensinjin @tnightingale @trisapeace @ApolloMuses please have a look. This PR gets as far as being able to toggle each answer in the questionnaire on or off. Still quite a bit of business logic to implement, coming in the next PR.

src/application/constants.ts
```diff
@@ -4,3 +4,5 @@ export const RESET_COUNTER = 'COUNTER:RESET_COUNTER';
 export const SET_MESSAGE_TEXT = 'MESSAGE:SET_MESSAGE';
 
 export const SET_MAIN_TAB = 'ROUTER:SET_MAIN_TAB';
+
+export const SELECT_ANSWER = 'QUESTIONNAIRE:SELECT_ANSWER';
```
bensinjin: What is the convention for values to these constants?

bensinjin:
On another note my review for this PR was a bit nutty but what I really needed was the ability to comment on each particular commit in its entirety. Moving forward I may just use comments on the PR as I think it lends itself better to discussion when I'm not pointing to a specific line of code.

src/application/constants.ts
rasmus-storjohann-PG: We don't have a convention yet. Looking forward to figuring this out as we go forward. The main constraints are that they need to be unique (would be nice to have a system for that) and they need to make sense in a debugger.

src/components/questionnaire/select_all_questions.ts
```diff
@@ -0,0 +1,38 @@
...
+    });
+};
+
+const selectAnswersForQuestion = (questionId: model.Id, answers: model.AnswersMap): ReadonlyArray<viewModel.Answer> => {
+    const keys = answerKeysForGivenQuestion(questionId, answers);
+
+    return buildViewModelForAnswers(keys, answers);
+};
+
+const answerKeysForGivenQuestion = (questionId: model.Id, answers: model.AnswersMap): Array<string> => {
```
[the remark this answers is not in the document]
rasmus-storjohann-PG: Yup, I've definitely been thinking about memoization here, I'll put that in the second PR for this story.

trisapeace:
@rasmus-storjohann-PG I haven't had a chance to look at this one or try it out yet. But it looks like a lot of files have been removed or moved around. Would it be possible to add a description to the PR to explain at a high-level the sorts of changes that you made so that it's easier for me to tell which changes are important? Thanks!

bensinjin:
@rasmus-storjohann-PG cheers for the responses and thanks for the commenting tip.

Regarding structure and reducers. I like the screens approach. In the past I've used the notion of screens as a way to group components and it feels pretty intuitive in my opinion.  In terms of reducers I see the name of the reducer function as a representation of the state that reducer manages in the store and tend to agree with something similar to what @ApolloMuses outlines. Along these lines I'd expect a task list reducer function to be called "taskList" and to take a "taskListIinitialState" variable as the default value for its state parameter.  I'd personally either suffix these files with "reducer" or put them in a "reducers" directory to add additional clarity. I think it's probably ok to use the term "index" as the place for the combineReducers call but to me just calling it store.ts is more clear.

Regarding styling I've yet to put a new PR up (I've held off for the time being while we iron out some more details around the store) but I do think there is merit to keeping styling out of components as it will hopefully lend itself to style reuse.

The private and view model stuff seems sensible to me and I'm sure there will be more discussions around selectors and dealing with normalized data from the store.  Nice work.

src/stores/page_switcher.ts
```diff
@@ -0,0 +1,53 @@
...
+    Questionnaire, MyPlan, ExploreAll,
+}
+
+export type Store = Readonly<ReturnType<typeof buildDefaultStore>>;
+
+export type SetMainPageAction = Readonly<ReturnType<typeof setMainPage>>;
+
+// tslint:disable-next-line:typedef
+export const setMainPage = (mainPage: Page | string) => (
+    helpers.makeAction(constants.SET_MAIN_TAB, { mainPage })
```
rasmus-storjohann-PG: Move validation from reducer to action creator

trisapeace:
Yay, it runs on my iPhone. 

I get a warning in Expo both in the command line and in the app that:
```
21:23:10: Warning: Each child in an array or iterator should have a unique "key"
 prop.

Check the render method of `Question`. See https://fb.me/react-warning-keys for
more information.
    in Answer (at question.js:14)
    in Question (at questionnaire.js:13)
    in RCTView (at View.js:60)
    in View (at View.js:9)
    in ViewNB (at connectStyle.js:384)
    in Styled(ViewNB) (at questionnaire.js:7)
    in RCTScrollContentView (at ScrollView.js:791)
    in RCTScrollView (at ScrollView.js:887)
    in ScrollView (at KeyboardAwareHOC.js:397)
    in _class (at Content.js:14)
    in Content (at connectStyle.js:384)
    in Styled(Content) (at questionnaire.js:5)
    in Component (created by Connect(Component))
    in Connect(Component) (at main_page_switcher.js:10)
    in Component (at main.js:8)
    in RCTView (at View.js:60)
    in View (at Container.js:15)
    in Container (at connectStyle.js:384)
    in Styled(Container) (at main.js:6)
    in Component (at with_font_loading.js:21)
    in _class (created by withI18n)
    in withI18n (created by Connect(withI18n))
    in Connect(withI18n) (at index.js:10)
    in I18nProvider (created by Connect(I18nProvider))
    in Connect(I18nProvider) (at index.js:9)
    in Provider (at index.js:8)
    in ErrorBoundary (at index.js:7)
    in Application (at registerRootComponent.js:35)
    in RootErrorBoundary (at registerRootComponent.js:34)
    in ExpoRootComponent (at renderApplication.js:33)
    in RCTView (at View.js:60)
    in View (at AppContainer.js:102)
    in RCTView (at View.js:60)
    in View (at AppContainer.js:122)
    in AppContainer (at renderApplication.js:32)
- node_modules\react-native\Libraries\ReactNative\YellowBox.js:69:16 in error
```
Is this something that we can fix?

I'm not a big fan of the UI of the checkmark. It doesn't really look like a checkmark, it re-aligns the text in the button when I check/uncheck it in order to centre it, and my buttons are too narrow to show both the text and the check for some of the longer ones causing it to run off the side of the button at the beginning and end.

I can see the jsLingui test at the top, which is cool. Could we add fake translations for the strings in the questionnaire? Just adding the [AR] to the beginning seems to work.

It looks like you've made some changes to page switching in the code, but I don't see any way to switch pages in the app anymore. Should I be able to?

I don't see the footer. Should I?

I can see the header, but it doesn't do anything. 
Should the Back arrow even be there when there aren't any pages to go back to?
Is the name "Pathways" supposed to be sort of  right-aligned? It feels like it should be centred.

I'll add a screenshot to the Slack channel. See you tomorrow!

rasmus-storjohann-PG:
> Warning: Each child in an array or iterator should have a unique "key"
 prop.

Warning fixed, thanks @trisapeace

rasmus-storjohann-PG:
# About the source code restructuring in this PR:
* All the 'hello world' tutorial stuff is gone
* Using native base widget set
*  Introducing consistent naming of components as `Component` and `ConnectedComponent`, and using namespace syntax to disambiguate.
* Created new questionnaire components
* Experimenting with `view_model` as concept, in my next PR this goes away and that code ends up together with the selectors.
* Created fixture for off-line first data source, to be replaced by server API calls once we're online
* Renamed navigation bar to page switcher.
* Reorganized how helper code is organized, it's getting close.

PR #60, opened 2018-05-11
Feature/issue 41 questionnaire component part 2

rasmus-storjohann-PG:


rasmus-storjohann-PG:
Hi @bensinjin @tnightingale @trisapeace please have a look at one more PR for the questionnaire.
* Supports answers that accept only one or more than one answers
* Implements functionality to check and uncheck answers
* Refactored where test helpers go
* Fixed a lot of lint warnings related to readonly
* Combined view model stuff into the selector
* Created test helpers for question and aswer entities, using builder pattern for this, I'd like to see this pattern used as widely as possible *in test code* for building data structures, they are not as useful for production code.

bensinjin:
@rasmus-storjohann-PG This is looking good. Nice progress on the tests and selectors. I'm happy with this.

tslint.json
```diff
@@ -1,4 +1,5 @@
 {
+    "rulesDirectory": "node_modules/tslint-immutable/rules/",
```
tnightingale: Very minor; the `node_modules/` directory prefix is unnecessary, everything in `node_modules` is already included in the rules lookup path, so this could be `"rulesDirectory": "tslint-immutable/rules/"`.

PR #62, opened 2018-05-11
Feature/issue 61 clean up readme

rasmus-storjohann-PG:


rasmus-storjohann-PG:
@tnightingale can you have a look at this very simple PR. I've changed up the package.json scripts a bit to use generic terms rather than terms from the particular library we're using.

README.md
```diff
@@ -1,34 +1,78 @@
+# Pathways backend
```
trisapeace: @rasmus-storjohann-PG 
Should this be `# Pathways client`?

Issue #44, opened 2018-04-17
Task detail component


rasmus-storjohann-PG:
This is the one-task-fills-the-whole-screen view of a task. 

AC: 
* When clicking on the ">" button of a given task in the task list, open the detail page of that task.
* Style the detail page as follows:

![image](https://user-images.githubusercontent.com/32307781/40735542-48b1fe5a-63f0-11e8-8740-31a04948f081.png)

Out of scope to keep the PR small: The Find Services sub-screen. The Learn More section. Any new functionality, including marking a task as completed, share and "..." button. So all tasks will show the COMPLETED TASK label for now.

For context, this is the whole screen:

![screenshot from 2018-05-30 10-01-09](https://user-images.githubusercontent.com/32307781/40735828-103ec246-63f1-11e8-9277-6fa150b1574f.png)

rasmus-storjohann-PG:
Suggested shape for the Task entity:
```
interface Task {
    id: string;
    title: string; // localized
    description: string; // localized
    category: string; // i.e. taxonomy: education, health, etc.
    importance: number; // on enum, values 1-5 or something
    starred: boolean; // this is likely to be tweaked
    completed: boolean;
    isFree: boolean; // this is likely to be tweaked
    lastUpdated: date; // don't know what we'll use to represent dates
}
```
This would be the form of the data for the consumption in the component, a normalized structure should be used in the store, and a selector function should be tasked with converting from one to the other. At least that is my current thinking on normalization. @tnightingale does that sound right to you?

rasmus-storjohann-PG:
@bensinjin we may have had a miscommunication, I took this story to mean the one-task-fills-the-whole-screen component. Did you take it to mean the one-task-in-a-list-of-tasks component? If so, it makes sense that you wanted to do this story at the same time as #42, but as I see it, I think your PR only covers #42.

bensinjin:
Yeah looks that way. The issue number referenced in the commits on the PR have been updated.

Issue #42, opened 2018-04-17
Task list component


rasmus-storjohann-PG:
List of tasks, with a star icon on each as per revised mocks. Top of task list contains task list intro popup which can be dismissed.

PR #65, opened 2018-05-17
Feature/issue #42 tasks list

bensinjin:
@tnightingale @trisapeace @rasmus-storjohann-PG  So this is the first crack at tasks and task list(s). I suspect very much this will evolve. Following I've outlined what I've implemented and assumptions I've made:

- A "task" currently represents an actionable item a user can interact with ie. add/remove/star/share
- A task holds a reference to its "definition" as defined by pathways.  A task definition includes for example a title, description, tags, category etc... The separation here allows us to update task definition data without touching task data.
- A task list is a container for tasks. I'd intend this to be used for "saved tasks", "suggested tasks", "sub tasks" etc...
- The "My Plan" page currently displays tasks taken from what I'd envision to be the user's "saved tasks".  I've rendered it twice just to demonstrate what some of the newer designs suggest .. naturally UI state is not wired up (see below).

All the necessary functionality for tasks to "work" is currently in place (run the tests).  I've not wired up the UI to demonstrate how the actions work as that would fall out of the scope of issues #42 & #44.  I'd like to see the next tasks related issue be around demonstration how to we can add tasks from the suggested list into the saved list or something along those lines. Anyways let me know what you think and cheers for the feedback.


src/components/tasks/task_list.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+export interface Props {
+    readonly tasks: ReadonlyArray<TaskViewModel>;
+}
+
+export interface Actions {
+}
+
+// tslint:disable-next-line:no-any
+const renderTaskDetail = ({item}: any): JSX.Element => {
```
rasmus-storjohann-PG: I think there should be a way to make this `any` go away if we can specify the element type to the `FlatList` below...

src/fixtures/tasks.ts
```diff
@@ -0,0 +1,100 @@
...
+export interface TaskDefinition {
+    readonly id: Id;
+    readonly title: string;
+    readonly description: string;
+    readonly tags: ReadonlyArray<string>; // i.e important, do soon, free etc.
+    readonly category: string; // i.e. education, health, transportation etc.
+    readonly importance: number;
+}
+
+export interface TaskDefinitions {
```
rasmus-storjohann-PG: I'm not a fan of the pattern where we have types that differ in just one letter, as in `TaskDefinition` and `TaskDefinitions`. If we did it just once, that would not be so bad, but to establish the pattern of doing this everywhere, we're making for less readable code. Can we instead go with `TaskDefinitionMap` or something similar?

src/selectors/__tests__/tasks.test.ts
```diff
@@ -0,0 +1,25 @@
...
+        expect(firstTask).toHaveProperty('id');
+        expect(firstTask).toHaveProperty('title');
+        expect(firstTask).toHaveProperty('description');
+        expect(firstTask).toHaveProperty('category');
+        expect(firstTask).toHaveProperty('importance');
+        expect(firstTask).toHaveProperty('starred');
+        expect(firstTask).toHaveProperty('completed');
+        expect(firstTask).toHaveProperty('suggested');
+        expect(firstTask).toHaveProperty('tags');
+    });
```
rasmus-storjohann-PG: Check out `/src/selectors/__tests__/questionnaire.tests.ts` for a pattern to use when having a lot of asserts with the same preconditions. It's one way to get to one assertion per test without duplicating a lot of code.

Also, the task selector is doing denormalization of data, this needs to be tested. Not just because everything should be tested (which it should) but because it's error prone and bugs are not that easy to detect through manual testing. There will be future refactoring in this area that will introduce bugs, let's squash those now.

src/components/tasks/task.tsx
```diff
@@ -14,12 +15,12 @@ export const Task: React.StatelessComponent<TaskViewModel & Actions> = (props: T
...
         <View style={styles.sideColumn}>
             <Button rounded success>
-                <Text>Add</Text>
+                <Text><Trans>Add</Trans></Text>
             </Button>
         </View>
         <View style={styles.centerColumn}>
             <View style={styles.stackedItems}>
-                <Text>{props.title}</Text>
+                <Text><Trans>{props.title}</Trans></Text>
```
tnightingale: `<Trans />` only works for static strings. In this case `props.title` is dynamic content.

src/stores/__tests__/tasks.test.ts
```diff
@@ -0,0 +1,44 @@
...
+import * as models from '../tasks';
+
+describe('tasks reducer', () => {
+    const fakeTask: models.Task = {
+        'id': 'fake1',
+        'taskDefinitionId': 'fake1Definition',
+        'starred': false,
+        'completed': false,
+        'suggested': false,
+    };
```
rasmus-storjohann-PG: This pattern does not scale, please make a builder as in `/src/stores/__tests__/helpers/questionnaire_helpers.ts`. If you need data in particular formats, add functions for that to `/src/application/__tests__/helpers/random_test_values.ts`. Same as on the server side, having these helpers in place allows us to write simple, expressive tests in the future.

src/stores/__tests__/tasks.test.ts
```diff
@@ -0,0 +1,44 @@
...
+    const initialStore: models.Store = models.reducer(
+        {tasks: {}, taskDefinitions: {}},
+        models.addToTaskList(fakeTask),
```
rasmus-storjohann-PG: It's a bit of a code smell to use production code as helper functions in test code. Production code should only be called in order to test the code you're calling.

src/stores/__tests__/tasks.test.ts
```diff
@@ -0,0 +1,44 @@
...
+    it('can add tasks to tasks list', () => {
+        expect(initialStore.tasks.fake1).toEqual(fakeTask);
+        expect(Object.keys(initialStore.tasks)).toHaveLength(1);
+    });
+
+    it('can remove tasks from tasks list', () => {
+        const finalStore = models.reducer(initialStore, models.removeFromTaskList(fakeTask.id));
+        expect(Object.keys(initialStore.tasks)).toHaveLength(1);
+        expect(Object.keys(finalStore.tasks)).toHaveLength(0);
+    });
```
rasmus-storjohann-PG: These two tests "can add tasks to tasks list" and "can remove tasks from tasks list" both have the same setup of `initialStore`, but only one of them is using it. "can add tasks to tasks list" does not actually do the addition, it's happening in the setup. Please group your tests together by common setup in `beforeEach()` calls. When I see a test that contains only assertions, I expect to see one or more additional, very similar tests that make other assertions about the same state that was created in a `beforeEach()` call. Also, I think a builder for tasks will make it a lot easier to build a store with a task that is starred or not starred, complete or not complete, etc. Sorry if my comments kinda run on here, we can talk about this if you like.

src/stores/tasks.ts
```diff
@@ -0,0 +1,103 @@
...
+        case constants.ADD_TO_TASK_LIST:
+            return {
+                ...store,
+                tasks: {
+                    ...store.tasks,
+                    [action.payload.task.id]: action.payload.task,
+                },
+            };
+        case constants.REMOVE_FROM_TASK_LIST:
+            const tasks = { ...store.tasks };
```
rasmus-storjohann-PG: I was gonna say that you could just say `const tasks = store.tasks;` but that would be a shallow copy, right? So your syntax creates a deep copy? Heads up that I'm going to suggest we start using [ramda](https://ramdajs.com/) which contains tons of functions for doing this kind of thing.

src/stores/tasks.ts
```diff
@@ -0,0 +1,103 @@
...
+            // tslint:disable-next-line:no-expression-statement
+            delete(tasks[action.payload.taskId]);
+            return { ...store, tasks };
+        case constants.TOGGLE_TASK_COMPLETED: {
+            const task = store.tasks[action.payload.taskId];
+            return setTaskValue(store, task, 'completed', !task.completed);
```
rasmus-storjohann-PG: or `return toggleTaskValue(store, task, 'completed');` ? That change would simplify the helper function signature, I think two functions accepting boolean and string respectively is more readable that one function accepting either.

src/stores/tasks.ts
tnightingale: `const tasks = store.tasks;` creates a new reference to the _same_ object.

`const tasks = { ...store.tasks };` creates a _new object_ with properties that reference the same objects/values as `store.tasks` (shallow copy).

In this case, `tasks` is created as local variable for the purpose of performing a `delete` mutation. 

`tasks` is then merged into a shallow copy of the store and the resulting object is returned: 
```
delete(tasks[action.payload.taskId]);
return { ...store, tasks };
```

Unfortunately javascript doesn't have a clear/easy syntax for removing/deleting properties from immutable objects. Something like Ramda (or any of the many other immutable/functional toolkit libs) would certainly help tidy up these awkward bits.

src/components/tasks/task.tsx
bensinjin: Oh of course. Fixed.

src/components/tasks/index.ts
```diff
@@ -0,0 +1,21 @@
+import { Dispatch } from 'redux';
+import { connect } from 'react-redux';
+import { Component, Props } from './task_list';
+import { Actions } from './task';
+import { Store } from '../../application/store';
+import { selectTasks } from '../../selectors/tasks';
+import * as models from '../../stores/tasks';
```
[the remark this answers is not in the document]
bensinjin: Yeah ok. Fixed.

src/application/constants.ts
```diff
@@ -3,3 +3,10 @@ export const SET_MAIN_TAB = 'ROUTER:SET_MAIN_TAB';
 export const SET_LOCALE = 'I18N:SET_LOCALE';
 
 export const SELECT_ANSWER = 'QUESTIONNAIRE:SELECT_ANSWER';
+
+export const ADD_TO_TASK_LIST = 'TASKS:ADD_TO_TASK_LIST';
+export const REMOVE_FROM_TASK_LIST = 'TASKS:REMOVE_FROM_TASK_LIST';
+export const TOGGLE_TASK_COMPLETED = 'TASKS:TOGGLE_TASK_COMPLETED';
+export const TOGGLE_TASK_STARRED = 'TASKS:TOGGLE_TASK_STARRED';
+export const TOGGLE_TASK_SUGGESTED = 'TASKS:TOGGLE_TASK_SUGGESTED';
+export const SHARE_TASK = 'TASKS:SHARE_TASK';
```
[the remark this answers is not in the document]
bensinjin: I wasn't clear on this myself. I was also not entirely clear how we're breaking up our subdirectories under components which is what I was sorta following for the prefix to these values.

src/components/tasks/task.tsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import { View } from 'react-native';
+import { Text, Button, Icon } from 'native-base';
+import { taskDetail as styles } from './styles';
+import { TaskViewModel } from '../../selectors/tasks';
```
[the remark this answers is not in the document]
bensinjin: Fixed.

src/stores/tasks.ts
rasmus-storjohann-PG: https://ramdajs.com/docs/#omit

src/components/tasks/task_list.tsx
bensinjin: Fixed.

src/fixtures/tasks.ts
bensinjin: Fixed.

src/components/tasks/actions.ts
```diff
@@ -0,0 +1,9 @@
+import * as stores from '../../stores/tasks';
```
trisapeace: In some places we use curly braces around these imports and not in others:

`import * as stores from '../../stores/tasks';`
or
`import { * as stores } from '../../stores/tasks'`

For consistency, I suggest that we always use the curly braces and then it gives us the flexibility to add more things to export from the same file without needing to change it. What do you think?

src/application/constants.ts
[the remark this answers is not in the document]
rasmus-storjohann-PG: How about something like this:
```
export namespace Task {
    export const ADD_TO_LIST = 'TASKS:ADD_TO_LIST';
    export const REMOVE_FROM_LIST = 'TASKS:REMOVE_FROM_LIST';
    export const TOGGLE_COMPLETED = 'TASKS:TOGGLE_COMPLETED';
    export const TOGGLE_STARRED = 'TASKS:TOGGLE_STARRED';
    export const TOGGLE_SUGGESTED = 'TASKS:TOGGLE_SUGGESTED';
    export const SHARE = 'TASKS:SHARE';
}
```

src/components/tasks/task.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+export const Task: React.StatelessComponent<selector.Task & Actions> = (props: selector.Task & Actions): JSX.Element => {
+    return (
+        <View style={styles.wrapper}>
+            <View style={styles.sideColumn}>
+                {!props.addToTaskList ? undefined : (
+                    <Button
+                        rounded
+                        success
+                        onPress={(): stores.AddToTaskListAction => props.addToTaskList(selector.normalizeTask(props))}>
```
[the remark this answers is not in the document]
bensinjin: Right of course. I was probably a bit eager to get something actually visibly "working". I'll be adopting what's outlined here.

src/selectors/__tests__/tasks.test.ts
bensinjin: Fixed. Got the tests down to one assertion per test, broke the denormalization piece out of the selector, and added tests for denormalization in isolation.

src/stores/__tests__/tasks.test.ts
bensinjin: Fixed, implemented builders + tests for builders.

src/stores/__tests__/tasks.test.ts
bensinjin: This is now using a builder. It is still, however, passing the store from the builder to the reducer and checking the state after actions are applied ... I'm not sure I follow why that would be a code smell?

src/stores/__tests__/tasks.test.ts
bensinjin: Fixed.  These tests have all been substantially cleaned up and organized and are now using various builders. Let me know how it's looking.

src/stores/tasks.ts
bensinjin: Sure, fixed.

src/application/constants.ts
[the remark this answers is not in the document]
bensinjin: Namespace works for me. I'm still not clear on the convention for the namespace or is that not something we're not being picky about?

src/components/tasks/index.ts
```diff
@@ -0,0 +1,32 @@
+import { Dispatch } from 'redux';
+import { connect } from 'react-redux';
+import { Component, Props, renderTask, renderSuggestedTask } from './task_list';
+import { Actions } from './task';
```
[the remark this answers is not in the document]
bensinjin: You're right. You'll notice this is also the case with the questionnaire piece. We're in the process of solidifying conventions around this. I'm sure @rasmus-storjohann-PG has thoughts.

src/components/tasks/actions.ts
bensinjin: I'd just try and follow this https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import and opt to use curly braces when accessing named exports or creating new aliases to multiple named imports.

bensinjin:
Hold off on reviews for this. I'm currently in the process of adjusting the data model.

src/stores/__tests__/helpers/tasks_helpers.ts
```diff
@@ -0,0 +1,124 @@
...
+        return this;
+    }
+
+    withImportance(importance: number): TaskDefinitionBuilder {
+        this.importance = importance;
+        return this;
+    }
+
+    build(): TaskDefinitionBuilder {
+        return this;
```
rasmus-storjohann-PG: I'm not reviewing, just having a quick peek. This function should not return `TaskDefinitionBuilder`, it should return `TaskDefinition`. You use the builder to gradually build up the state for the thing to build, and when you're done, call build to get the thing you've been building. `TaskBuilder#build()` should similarly return a `Task`.

src/stores/__tests__/tasks_helpers.test.ts
```diff
@@ -0,0 +1,162 @@
...
+                .withDescription(description)
+                .withTags(tags)
+                .withCategory(category)
+                .withImportance(importance)
+                .build();
+        });
+
+        test('id property', () => {
+            expect(taskDefinition.id).toBe(id);
+        });
```
rasmus-storjohann-PG: The test will be more self-contained if you do
```
test('id property', () => {
    const id = aString();
    const taskDefinition = new TaskDefinitionBuilder().withId(id).build();
    expect(taskDefinition.id).toBe(id);
});
```

src/stores/__tests__/helpers/tasks_helpers.ts
bensinjin: Right. Yeah I'm in the process of fixing this. Thanks for the clarification on the builders I had missed this important detail.

src/stores/__tests__/tasks_helpers.test.ts
bensinjin: Fixed.

bensinjin:
The store for tasks now looks as follows:

```
{
    taskMap // Task objects with fields: title, description, tags etc...
    taskUserSettingsMap //  Task user settings objects with fields: taskId, starred, completed etc...
    savedTasksList // A list of task ids representing "saved" tasks
    suggestedTasksList // A list of task ids representing "suggested" tasks
}
```

@rasmus-storjohann-PG I had already done the refactoring work for this structure prior to discussions with you about potentially keeping everything in one "table". I don't think there is a strong enough case against this structure and have opted to leave it as is until the need arrises to change it.

src/stores/__tests__/helpers/tasks_helpers.ts
bensinjin: Fixed

bensinjin:
I've stubbed this in https://github.com/pg-irc/pathways-frontend/issues/78 which would rely on this work.

src/application/constants.ts
```diff
@@ -4,9 +4,12 @@ export const SET_LOCALE = 'I18N:SET_LOCALE';
...
 
-export const ADD_TO_TASK_LIST = 'TASKS:ADD_TO_TASK_LIST';
-export const REMOVE_FROM_TASK_LIST = 'TASKS:REMOVE_FROM_TASK_LIST';
-export const TOGGLE_TASK_COMPLETED = 'TASKS:TOGGLE_TASK_COMPLETED';
-export const TOGGLE_TASK_STARRED = 'TASKS:TOGGLE_TASK_STARRED';
-export const TOGGLE_TASK_SUGGESTED = 'TASKS:TOGGLE_TASK_SUGGESTED';
-export const SHARE_TASK = 'TASKS:SHARE_TASK';
\ No newline at end of file
+export namespace Task {
+  export const ADD_TO_SAVED_LIST = 'ADD_TO_SAVED_LIST';
```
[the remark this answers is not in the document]
bensinjin: Fixed

src/components/tasks/index.ts
```diff
@@ -1,32 +1,32 @@
...
 const mapSuggestedTasksDispatchToProps = (dispatch: Dispatch<Store>): Actions => ({
-    addToTaskList: (task: stores.Task): stores.AddToTaskListAction => dispatch(stores.addToTaskList(task)),
-    shareTask: (taskId: stores.Id): stores.ShareTaskAction => dispatch(stores.shareTask(taskId)),
+    addToSavedList: (taskId: stores.Id): stores.AddToSavedListAction => dispatch(stores.addToSavedList(taskId)),
+    shareTask: (): stores.ShareAction => dispatch(stores.share()),
 });
 
-export const ConnectedTasks = connect(mapTasksStateToProps, mapTasksDispatchToProps)(Component);
+export const ConnectedSavedTasks = connect(mapSavedTasksStateToProps, mapSavedTasksDispatchToProps)(Component);
 export const ConnectedSuggestedTasks = connect(mapSuggestedTasksStateToProps, mapSuggestedTasksDispatchToProps)(Component);
```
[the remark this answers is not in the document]
bensinjin: Fixed.  I've opted to remove the index.ts file as I don't feel it's necessary to keep it around just to export what these two files will already be exporting. Let me know if that's an issue.

src/stores/tasks.ts
```diff
@@ -56,47 +63,62 @@ export const reducer = (store: Store = buildDefaultStore(), action?: TaskAction)
...
-            // TODO
-            return store;
+        case constants.ADD_TO_SAVED_LIST:
+            return addToTaskList(store, 'savedTasksList', store.savedTasksList, action.payload.taskId);
+        case constants.REMOVE_FROM_SAVED_LIST:
+            return removeFromTaskList(store, 'savedTasksList', store.savedTasksList, action.payload.taskId);
+        case constants.ADD_TO_SUGGESTED_LIST:
+            return addToTaskList(store, 'suggestedTasksList', store.suggestedTasksList, action.payload.taskId);
+        case constants.REMOVE_FROM_SUGGESTED_LIST:
+            return removeFromTaskList(store, 'suggestedTasksList', store.suggestedTasksList, action.payload.taskId);
```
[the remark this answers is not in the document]
bensinjin: No matter how you look at it you're still going to have to be updating some value in the store - that's how you trigger the re render in the UI. Adding and removing elements from an array is no more clunky than updating any value in the store (in my opinion).

Issue #66, opened 2018-05-18
Create a generic loading HOC to be used as application splash-screen


tnightingale:
**Proposal:**
Generalize the `withFontLoading()` HOC to be a `withLoader()` HOC that takes a `loading: boolean` prop and depending on its value, either renders either a loader/splash screen or the wrapped component.

**Explanation:**
As our application grows there will be async tasks that will need be completed on bootstrap before we can perform our first meaningful (interactive) render.

Currently there is a [`withFontLoading()`](pg-irc/pathways-frontend/blob/master/src/components/helpers/with_font_loading.tsx) HOC that delays rendering its child components (the entire application UI) until it ensures the correct fonts have loaded. It is implemented using React's component lifecycle hooks & component state to track the status of the font loading.

A new bootstrap task is to [fetch the user's current locale/RTL layout preference](#56); we want to prevent the UI from flashing the default locale and instead wait until the user's locale preference has been loaded before rendering any screens.

Initially its input can be simple; just a `loading: boolean` property but functionality could be expanded:
- the 'loading' render output can be configurable; pass a component to be used as splash-screen/loading message
- handle individual failure/error cases
- display individual task loading progress/completion

ApolloMuses:
This is a good idea, using HOC is a great way to generalize loading for repeated screens as well as something like authentication flow. Here's a example of using HOC for authentication written for a React app.

`
```

import React, { Component } from 'react';
import { connect } from 'react-redux';


export default function(ComposedComponent) {
  class Authentication extends Component {
    static contextTypes = {
      router: React.PropTypes.object
    } //to use react-router's context, must declare a context types for each

    componentWillMount() {
      if (!this.props.auth) {
        this.context.router.push('/'); //pushes the user back to main page
      }
    }

    componentWillUpdate(nextProps) {
      //handles when users clicks sign out
      if (!nextProps.auth) {
          this.context.router.push('/');
      }
    }

    render() {
      return <ComposedComponent {...this.props} />
    }
  }

  const mapStateToProps = (state) => {
    return { auth: state.auth }
  }
  return connect(mapStateToProps)(Authentication);
}

/*
in some other file, we want to use this HOC

..in this other file

import Authentication //HOC
import Resources //The component I want to wrap (prevent users from gaining access)

const ComposedComponent = Authentication(Resources);

...in some render method

<ComposedComponent />
 */

```
`

tnightingale:
[...]

Issue #56, opened 2018-04-28
Investigate support for alt layouts for RTL languages


tnightingale:
Continuation of #31 (#55); some locales (mostly middle eastern) require text to be rendered right-to-left instead of left-to-right. Our i18n solution needs to handle this.

The OS (iOS and Android) appear to provide some helpers here if user sets their system language to one that is RTL but we probably need to do this on our own, at the application level (of course while taking into consideration what the user has set globally).

React Native & Expo components claim to mostly support RTL layouts: https://facebook.github.io/react-native/blog/2016/08/19/right-to-left-support-for-react-native-apps.html. We need to verify this and determine if there are any caveats/gotchas.

AC: Switching language to Arabic causes horizontal layouts to switch for RTL rendering.

tnightingale:
Good news and bad news:

React Native's `I18nManager`'s methods, `allowRTL()` & `forceRTL()`, and its `isRTL` property all seem to work as advertised; they reverse all layout properties that take values like start/end, left/right, row/row-reverse and others (I haven't found a definitive list yet).

_Unfortunately for the layout switch to take effect, the application needs to be restarted._ This impacts the 'flow' of the language switcher in a way that may not be acceptable.

~~~On top of this, there's no method to restart the application automatically without [a plugin](https://github.com/avishayil/react-native-restart); the user must kill & restart the application. (`Expo.Util.reload()` exists but [it only works in the Expo app](https://github.com/expo/expo/issues/10#issuecomment-302247064)).~~~

Update: seems like `Expo.Util.reload()` works fine in stand-alone builds. I have demo builds you can try for yourself here:

**iOS**
https://exp-shell-app-assets.s3-us-west-1.amazonaws.com/ios%2F%40tnightingale%2Fpathways-cca86dc3-a9df-41ee-88c6-d9aff21a779a-simulator.tar.gz
To run on simulator:
1) Download
2) untar, eg: `tar -xzvf /path/to/tarfile`
3) Then drag/drop .app file onto Simulator.

***Android***
https://exp-shell-app-assets.s3-us-west-1.amazonaws.com/android%2F%40tnightingale%2Fpathways-c4f10c51-53e7-11e8-80f2-0a580a782087-signed.apk
To install on emulator:
1) Download
2) use `adb`, eg: `adb install /path/to/apk-file`

The alternative is to implement RTL overrides in userspace 👎

tnightingale:
[...]

Issue #36, opened 2018-04-17
Local storage infrastructure


rasmus-storjohann-PG:
ACs: Describe persisted entities and their properties, remaining agnostic to SQL or key/value storage
* Questions + optional responses
* Tasks, all tasks in my task list are available off line.
* Tasks completed.
* User ordering of tasks.
* Services for a task.
* Starring of tasks.
* Deleted tasks: user removed from task list.
* Task annotated with taxonomy term
* Task importance / urgency.
* User language settings.
* User access token.
* "Don't show this again" state for various popups
* Migration, validation to ensure that migration worked correctly.

library choice: redux-persist or expo-sqlite. 
UX: consider spotify UX for downloading content for offline access

tnightingale:
I have briefly looked into commonly used mobile/React Native storage options. Unfortunately there isn't much. Redux-persist seems to be the most common choice among React Native devs (clearly they're all also using Redux).

The other frequently mentioned options are:
- [Expo SQLite]
- [react-native-sqlite-storage]
- [Realm]
- [Firebase]
- [PouchDB]/[Couchbase Mobile]

### [Expo SQLite](https://docs.expo.io/versions/latest/sdk/sqlite)
**Pros:** 
Full access to sqlite database, referential integrity, transactions
**Cons:** 
No support APIs, eg: migrations, querybuilder, etc
**Notes:**
[Migrations are potentially not too hard to implement ourselves](https://github.com/litehelpers/Cordova-sqlite-storage#database-schema-versions) (example comes from the Cordova version of react-native-sqlite-storage).

I also looked at http://knexjs.org/ as a way of providing some support. There are a couple examples of people using it in React Native with both Expo SQLite & react-native-sqlite-storage. Unfortunately it's fundamentally built for a nodejs/server/non-mobile environment and requires a slew of nodejs polyfills. The migration & seeding functionality assumes CLI & filesystem access so likely isn't very useful to us.

### [react-native-sqlite-storage]
**Pros:**
Same as Expo SQLite
**Cons:**
Same as Expo SQLite but also requires us to eject from CRNA and install manually with some non-trivial, platform-dependant steps ([iOS](https://github.com/andpor/react-native-sqlite-storage#how-to-use-ios) & [android](https://github.com/andpor/react-native-sqlite-storage#how-to-use-android))
**Notes:**
Same as Expo SQLite

### [Realm]
**Pros:**
[...]

PR #67, opened 2018-05-18
Issue #56: Add RTL layout and integrate with locale switcher

tnightingale:
_Turned into a larger task than expected as it exposed the need for async actions and persistent storage._

There are details in #56 but a quick summary of the 'surprises' encountered:
- Setting RTL layout in RN (via `I18nManager.forceRTL(true)`) requires application process to be restarted
- Restarting application process after user selects desired locale means that we need to **immediately** persist the user's preference (this would have been required eventually anyway)
- Cannot restart application process until we know that the user's preference was successfully saved (no redux-persist-style lazy-write)
- Adjust application bootstrap process to load user's locale preference and then set current locale accordingly

As a result, this pull request makes some decisions regarding async actions and persistent storage (#36):
- Async actions: Set/load current locale side-effects have been implemented as _Sagas_ with redux-saga as a dependency.
- Persistence: Storing user's preferred locale has been implemented with React-Native's [AsyncStorage] API. AsyncStorage is a good fit in this case as it provides a lightweight key-value API which is ideal for user preferences. We most-certainly will want to continue to plan for a more robust persistence solution for our other data needs but AsyncStorage is a useful tool to have in for simple things.

This issue brought up the need for a loading screen while application bootstraps itself. I have created #66 to track this task and have a solution in a forthcoming pull request.


src/components/language_switcher/language_switcher.tsx
```diff
@@ -1,24 +1,29 @@
 import React from 'react';
```
rasmus-storjohann-PG: I suggest that we keep file names and component names consistent, so `Switcher` should be renamed to `LanguageSwitcher`.

src/components/language_switcher/language_switcher.tsx
```diff
@@ -1,24 +1,29 @@
...
 import { View, Text, Button } from 'react-native';
-import { Trans } from '@lingui/react';
+import { withI18n, Trans } from '@lingui/react';
+import { Locale, LanguageSwitcherProps } from './view_model';
 
-export interface Props {
-    readonly currentLocale: string;
-    readonly locales: ReadonlyArray<Locale>;
-}
```
rasmus-storjohann-PG: The idea of the `Props` and `Actions` is that they are exported along with the component itself to facilitate writing calls to `connect()`. So that when you build the `ConnectedLanguageSwitcher`, the `Actions` and `Props` interfaces comes from this file along with the component itself. I think we should try to keep using this pattern consistently, since it decouples (through being opinionated) the code that connects components up from the component code itself. (You'll find that I define `Action` interfaces even when they are empty for constancy.)

src/components/language_switcher/language_switcher.tsx
```diff
@@ -1,24 +1,29 @@
...
-            <Button key={locale.code}
-                title={locale.label}
-                onPress={(): void => setLocale(locale.code)}
-                disabled={locale.code === currentLocale} />
-        ))}
-        <Text><Trans>jsLingui test</Trans></Text>
-    </View>
-);
\ No newline at end of file
+export const LanguageSwitcher = I18nLanguageSwitcher;
```
rasmus-storjohann-PG: I'd be fine with
```
export const LanguageSwitcher = withI18n()(Switcher);
```

src/components/language_switcher/view_model.ts
```diff
@@ -0,0 +1,21 @@
...
+export interface Props {
+    readonly currentLocale: Locale;
+    readonly locales: ReadonlyArray<Locale>;
+}
+
+export interface Actions {
+    setLocale(locale: Locale): void;
+}
+
+export type LanguageSwitcherProps = I18nProps & Props & Actions;
```
rasmus-storjohann-PG: As suggested above, I would remove `LanguageSwitcherProps` and move the `Props` and `Actions` to the component file. I would handle the `I18nProps` by having `Props` inherit from it.

src/stores/__tests__/fonts.test.ts
```diff
@@ -0,0 +1,57 @@
...
+        it('should create action with payload containing the locale', () => {
+            const theLoadFontsAction = loadFontsActions.success(someFonts);
+            expect(theLoadFontsAction.payload.fonts).toBe(someFonts);
+        });
+
+    });
+
+    describe('failure', () => {
+
+        const errorMessage = '[test] Error occurred during loadFonts';
```
rasmus-storjohann-PG: Is the content of this string relevant to the test? If not, it's simpler to just do
```
const errorMessage = aString();
```

src/stores/__tests__/locale.test.ts
```diff
@@ -1,53 +1,163 @@
 // tslint:disable:no-expression-statement
 
+import { Locale } from '../../components/language_switcher/view_model';
 import * as locale from '../locale';
 import * as constants from '../../application/constants';
-import * as helpers from '../helpers/make_action';
 
-const buildStore = (): locale.Store => locale.reducer(undefined, undefined);
+const aLocale = { code: 'ar', label: 'Arabic', catalog: {}, isRTL: true };
```
rasmus-storjohann-PG: You will do us all a favour by creating a test helper for building a locale, `AnswerBuilder` might be the best pattern to follow, but there are others. This pattern works well for writing tests, and in particular if the builder is already in place, it makes it easier to write tests in the future, so it's a really good investment to put in early. You should be able to make use of a similar approach to do the `buildStoreWithLocale` and `buildStoreLoadingLocale`

src/stores/fonts.ts
```diff
@@ -0,0 +1,28 @@
...
+}
+
+interface Fonts {
+    readonly [name: string]: any; // tslint:disable-line:no-any
+}
+
+// tslint:disable-next-line:typedef
+export const request = (fonts: Fonts) => {
+    return helpers.makeAction(constants.LOAD_FONTS_REQUEST, { fonts });
+};
```
rasmus-storjohann-PG: Picky: prefer the syntax without the `return`

src/stores/fonts.ts
```diff
@@ -0,0 +1,28 @@
+import * as constants from '../application/constants';
+import * as helpers from './helpers/make_action';
+
+export namespace LoadFonts {
+    export type Request = Readonly<ReturnType<typeof request>>;
+    export type Success = Readonly<ReturnType<typeof success>>;
+    export type Failure = Readonly<ReturnType<typeof failure>>;
+    export type Result = Success | Failure;
+}
```
rasmus-storjohann-PG: This is inconsistent with how we've been doing it elsewhere, there should be no `export namespace`, and it should be `RequestAction` etc. The import can be changed from `import { LoadFonts } from '../stores/fonts';` to  `import * as LoadFonts from '../stores/fonts';`, so the namespace is used at the call site.

src/stores/locale/load_current_locale.ts
```diff
@@ -0,0 +1,25 @@
...
+    export type Request = Readonly<ReturnType<typeof request>>;
+    export type Success = Readonly<ReturnType<typeof success>>;
+    export type Failure = Readonly<ReturnType<typeof failure>>;
+    export type Result = Success | Failure;
+}
+
+// tslint:disable-next-line:typedef
+export const request = () => {
+    return helpers.makeAction(constants.LOAD_CURRENT_LOCALE_REQUEST);
+};
```
rasmus-storjohann-PG: Where are these functions called from? (I still have to spend some time wrapping my head around the saga stuff)

src/application/locales.ts
```diff
@@ -21,3 +37,43 @@ function buildCatalogsMap(withLocales: ReadonlyArray<Locale>): Catalogs {
...
+
+/**
+ * Persistence helpers:
+ * TODO: Consider putting these side-effect-causing functions somewhere else.
+ */
+
+/**
+ * I18nManager.forceRTL() requires application to be restarted for changes to
+ * take effect. To ensure that we load correct locale upon application restart,
+ * this needs to be called *after* desired locale has been persisted to storage.
```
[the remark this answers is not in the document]
tnightingale: This is an ugly block of code that reads from a global variable (essentially, an OS system call) and conditionally triggers two side-effects. As stated in the comment, I'm open to suggestions for alternative ways and/or places to put this code.

The conditional logic which reads the global (`I18nManager.isRTL`) and the possible side-effects that result (`I18nManager.forceRTL()` and `Expo.Updates.ReloadFromCache()`) are deliberately wrapped in a function so that it can be declared/called in a Saga as a side effect (see /src/sagas/locale.ts:15). The idea is that it keeps the messy side-effects separate from our application logic, keeps our application logic testable, etc...

src/application/router.ts
```diff
@@ -1,16 +1,21 @@
...
 import * as constants from '../application/constants';
-import { Store as StoreForApplicationState } from '../stores';
 
 const routesMap = {
     [constants.SET_MAIN_TAB]: '/user/:mainPage',
 };
 
-export const createStoreWithRouter = (reducerForApplicationState: Reducer<StoreForApplicationState>) => {
+interface ApplicationRouter {
+    readonly reducer: Reducer<LocationState<string, any>>; // tslint:disable-line:no-any
```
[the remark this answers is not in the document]
tnightingale: I overlooked this; there were a lot of tslint errors in this file when I realized I needed to extract the store creation logic from the router creation logic. I added the tslint directives to reduce some of that noise.

However this particular case feels like were providing types for 3rd party code again...

src/application/store.ts
```diff
@@ -1,8 +1,21 @@
...
-export const store = createStoreWithRouter(reducer);
+const router = buildRouter();
+const saga = buildSaga();
+
+const reducer = combineReducers({ location: router.reducer, applicationState: reducerForApplicationState });
+const middleware = applyMiddleware(router.middleware, saga.middleware);
+export const store = createStore(reducer, compose(router.enhancer, middleware));
+
+runSaga(saga.middleware); // tslint:disable-line:no-expression-statement
+store.dispatch(loadCurrentLocaleActions.request()); // tslint:disable-line:no-expression-statement
```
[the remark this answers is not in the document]
tnightingale: Yeah this is all one-off application composition stuff, it always ends up being a bit messy, fortunately we haven't had any chicken&egg situations yet...

I needed to separate the store creation bits in order to add other middleware (Saga, and later redux-devtools). I tried my best to provide a consistent interface for the creation & composition of our current extensions and future ones too.

src/stores/locale/load_current_locale.ts
tnightingale: `request()`, `success()` and `failure()` are currently _entirely_ convention-based.

- `request()` is called/dispatched from wherever is appropriate; in this case during the bootstrapping of the application in /src/application/store.ts but usually from a component after a user action. 
- `success()` and `failure()` are called as effects in the Saga (/src/sagas/locale.ts) depending on the results of the call to `request()`.

Definitely happy to improve on this convention as it's pretty arbitrary as it stands.

src/application/locales.ts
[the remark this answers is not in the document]
rasmus-storjohann-PG: Right, I could have been more helpful in my comment, let me have another go: One on the principles from clean code is that each function should only operate at one level of abstraction, so high level functions contain only high level code, and low level functions only low level code. I think the problem here is that you have a low level function, `setRTLAndReloadIfNeeded()`, that is doing a high level thing, which is to decide whether or not to switch the RTL state. I think that's something that `applyLocaleChange()` should be doing. I would break up `setRTLAndReloadIfNeeded()` into a `isReloadNeeded()` and a `reloadRTL()` or something like that. Then you'd have an if-statement in `applyLocaleChange()` which would make it clear there what the logic is. My rule of thumb of where to put the if-statement is that if the if-statement is an implementational detail that is irrelevant to understanding the big picture, stick it in a low-level function, but if the if-statement is an essential part of the logic big picture, separate out the predicate function and function with side effect, and put the if-statement in the high level function.

bensinjin:
This looks good to me. Looking forward to get a little more familiar with Redux Saga.

tnightingale:
@rasmus-storjohann-PG rebased

PR #68, opened 2018-05-18
Issue #66 Add loading HOC

tnightingale:
Details in #66.

_NOTE: This pull request's branch is based on the [feature/56-rtl_layout](67) branch. #67 should be merged before this branch._

Generalizes withFontLoading() into a generic loader HOC which adds a `loading:boolean` prop to any component.


rasmus-storjohann-PG:
I'll review this PR once the preceding one is merged, I think that will make it easier for both of us.

rasmus-storjohann-PG:
@tnightingale the PR needs rebasing as expected

rasmus-storjohann-PG:
@tnightingale this PR also needs a rebase, to make most of the commits disappear.

tnightingale:
@rasmus-storjohann-PG I'm just working on this right now, having an issue that I think stems from a bad conflict resolution. Will let you know when I've rebased

tnightingale:
@rasmus-storjohann-PG rebased and fixed the issue.

Had a bit of a distraction as the call to `Expo.Updates.reloadFromCache()` seems to occasionally crash the Expo app in the Android simulator (no issue on iOS); causing Expo to hang on the loading spinner screen indefinitely.

I have confirmed that this does not occur in stand-alone builds of the application, you can see for yourself by downloading this build https://expo.io/builds/738db8d2-60de-4926-a5ae-90fa9ed94e03 and installing with `adb install /path/to/downloaded/apk`.

rasmus-storjohann-PG:
Great to see, I will review this tomorrow, my brain is mush right now.

src/application/index.tsx
```diff
@@ -8,10 +8,12 @@ import { ConnectedI18nProvider } from '../components/i18n_provider';
 import { LocaleManager } from './locale';
 import enMessages from '../../locale/en/messages';
 import arMessages from '../../locale/ar/messages';
+import zhMessages from '../../locale/zh/messages';
 // tslint:disable-next-line:no-expression-statement
 LocaleManager.registerLocales([
     { code: 'en', label: 'English', catalog: enMessages, isRTL: false },
     { code: 'ar', label: 'Arabic', catalog: arMessages, isRTL: true },
+    { code: 'zh', label: 'Chinese', catalog: zhMessages, isRTL: false },
 ]);
```
rasmus-storjohann-PG: Looks like we'll eventually want index.tsx to just contain a single import for all the locale stuff.

src/components/main/loader.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import { Text } from 'native-base';
```
rasmus-storjohann-PG: I was able to simplify the loader

src/selectors/application_loading.ts
```diff
@@ -0,0 +1,7 @@
+import {} from '../application/store';
+import * as app from '../application/store';
+
+export const isApplicationLoading = (appStore: app.Store): boolean => {
```
rasmus-storjohann-PG: I find it's really helpful if we consistently name files by the symbol they exports, to I would prefer this file to be named `is_application_loading.ts`.

src/stores/__tests__/fonts.test.ts
```diff
@@ -1,23 +1,27 @@
...
-import * as loadFontsActions from '../fonts';
+import * as fonts from '../fonts';
 import * as constants from '../../application/constants';
 
 const someFonts = {
     MyFont: {},
 };
 
+const buildStoreLoading = (): fonts.Store => {
+    return { ...fonts.buildDefaultStore(), loading: true };
```
rasmus-storjohann-PG: Using production code as helper functions in tests is discouraged. If you ever want to refactor the production function, it nice to not have to change a lot of test code that depends on it for convenience only. In this case, not using `buildDefaultStore()` is actually both simpler and clearer, you can just do `return { loading: true }`. I'd argue that you don't need this function at all, and can just do that inline in the tests below.

In the case of future changes to the store definition, unit tests where the store is defined locally will fail the TypeScript type checking when the stores are passed to the reducer, so we don't need to use `buildDefaultStore()` to protect against that.

src/application/index.tsx
tnightingale: Yes, this is still floating around looking for a place to live; I moved it out of `src/application/locale/index.ts` as the locale management code has evolved to be more data-oriented and less a hard-coded module (though it's still a singleton).

It's not inconceivable to have these locale definitions (& their message catalogs) downloaded/updated at runtime. Though we would need to find a solution to dynamically load the I18n polyfill's runtime data:
```
// src/application/locale/index.ts
import 'intl/locale-data/jsonp/en.js';
import 'intl/locale-data/jsonp/ar.js';
import 'intl/locale-data/jsonp/zh.js';
```

src/components/main/loader.tsx
tnightingale: Looks good!

tnightingale:
Looks good to me!

Issue #69, opened 2018-05-22
Add Redux Devtools


tnightingale:
[Redux DevTools] provides useful insight into an application's state and state-transformations. Particularly when test-driving other people's code during code reviews; it is helpful to see the actions/action creators & reducers "in-action". 

Adding/enabling [Redux DevTools] for a React-Native app is a bit daunting due to the variety of options and plugins available.

I have had success with using [Remote Redux DevTools] to expose our application's Redux store and [remote-rn-debugger] to inject the [remotedev-app] monitor into React Native's [remote debugger].

Pull-request forthcoming...

[Redux DevTools]: https://github.com/reduxjs/redux-devtools
[Remote Redux DevTools]: https://github.com/zalmoxisus/remote-redux-devtools
[remote-rn-debugger]: https://github.com/jhen0409/remotedev-rn-debugger
[remotedev-app]: https://github.com/zalmoxisus/remotedev-app
[remote debugger]: https://github.com/jhen0409/remotedev-rn-debugger

PR #70, opened 2018-05-22
Issue #69 Add + configure redux devtools and dependencies

tnightingale:
Adds Redux DevTools + dependencies and configures according to default instructions. Result is Redux DevTools available in React Native's remote debugger window.

Despite the required changes to how we create the Redux store, Redux DevTools is only enabled when application is running/built in development mode.

See #69 for more details.


src/application/router.ts
```diff
@@ -12,5 +14,13 @@ export const createStoreWithRouter = (reducerForApplicationState: Reducer<StoreF
...
-    return createStore(rootReducer, compose(router.enhancer, applyMiddleware(router.middleware)));
+    const composeEnhancers = composeWithDevTools({
+      name: Platform.OS,
+      hostname: 'localhost',
+      port: 5678,
+    });
+    return createStore(rootReducer, composeEnhancers(
+        router.enhancer,
+        applyMiddleware(router.middleware),
+    ));
```
rasmus-storjohann-PG: This function is getting really complicated, and it will grow further, so I think it's important to keep it tidy. I played with it for a while and I got to this:
```ts
import { Platform } from 'react-native';
import { Reducer, Store, GenericStoreEnhancer, createStore, applyMiddleware, combineReducers } from 'redux';
import { composeWithDevTools } from 'remote-redux-devtools';
import * as reduxFirstRouter from 'redux-first-router';
import createMemoryHistory from 'history/createMemoryHistory';
import { Store as StoreForApplicationState } from '../stores';

const routesMap = {
    [constants.SET_MAIN_TAB]: '/user/:mainPage',
};

type RouterReducer = Reducer<reduxFirstRouter.LocationState<string>>;
type AppReducer = Reducer<StoreForApplicationState>;

const createRootReducer = (location: RouterReducer, applicationState: AppReducer): Reducer<any> => (
    combineReducers({ location, applicationState })
);

const createEnhancer = (routerEnhancer: GenericStoreEnhancer, middleware: GenericStoreEnhancer): GenericStoreEnhancer => {
    const devToolsOptions = {
        name: Platform.OS,
        hostname: 'localhost',
        port: 5678,
    };
    return composeWithDevTools(devToolsOptions)(routerEnhancer, middleware);
};

export const createStoreWithRouter = (appReducer: AppReducer): Store<any> => {
    const history = createMemoryHistory();
    const router = reduxFirstRouter.connectRoutes(history, routesMap);
    const rootReducer = createRootReducer(router.reducer, appReducer);
    const middleware = applyMiddleware(router.middleware);
    const enhancer = createEnhancer(router.enhancer, middleware);
    const store = createStore(rootReducer, enhancer);
    return store;
};
```
My thinking is to keep the high level function to the regular pattern of call a function with simple arguments and store in a new local ... and then pull out the stuff that doesn't fit in that pattern into helper functions. Note that this code compiles, I don't know whether it runs ;-)

tnightingale: Ultimately Redux store creation needs to be separated from the router creation as there are a bunch of things that need to augment the store/redux in the same way as the router (and there will be more as we add more features/dependencies). 

#56 contains an example of doing so.

.gitignore
```diff
@@ -21,3 +21,6 @@ locale/_build
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
+
+# Auto-generated by macOS's Finder
```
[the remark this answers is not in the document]
tnightingale: I find this arbitrary (not that I particularly care about whether the comment exists or not); I only added the comment to be consistent with the rest of the file.

Are you aware of the other comments in the file?

src/application/router.ts
tnightingale: Redux store creation follows a consistent pattern:
1. create a list of middlewares `const listOfMiddleware = [middleware1, middleaware2, ...];`
2. turn middlewares into an enhancer with `const middlewareEnhancer = applyMiddleWare(...listOfMiddleware);`
3. compose middleware enhancer with other 3rd party or custom enhancers: `const enhancer = compose(...[middlewareEnhancer, 3rdPartyEnhancer, customEnhancer]);`
4. compose reducer with: `const reducer = combineReducers(...[reducer1, reducer2, reducer3, ...]);`
5. create a store with composite reducer, composite enhancer & optionally, initial state: `const store = createStore(reducer, enhancer /*, initialState */);`

Now, in reality it's never that simple, as redux-first-router has demonstrated; often, for better or worse, middlewares, enhancers a reducers are inter-connected.

In my experience sticking to the above pattern helps avoid surprises when needing to extend Redux.

rasmus-storjohann-PG: OK, you have more experience with this that I do so I will defer to you on this. Sounds like this could be a case for using a buider (one of my favourite patters) so we'd be able to build the different pieces in any order and pass them to the builder, and then call build() which would be the only code that knows how all the different pieces are assembled. Just an idea, something to try once we have most of the pieces in place, so not for a while yet.

tnightingale: Ya, a store builder would probably be appropriate wrap all this up.

bensinjin:
Should be handy. Looks good to me.

Issue #74, opened 2018-05-23
Back button does not go back to the very first page


rasmus-storjohann-PG:
AC: Start the app, navigate using the buttons in the app footer, hit the back button repeatedly, verify that the app eventually goes back to the very first page.

PR #76, opened 2018-05-27
Issue #74: routing initial page

trisapeace:
@rasmus-storjohann-PG

I tried to make the changes based on your feedback in a couple of different ways and nothing worked out quite as cleanly as I would have liked.

Essentially the problem came down to how the redux-first-router implemented its Back button. I was expecting it keep a history of the route events that happened and all the payload that was included in those route changes - in which case I was going to have both the `mainPage` Page object and a `mainPageRoute` string as values in the SET_MAIN_PAGE event payload. However, based on what I was observing and a peek into the code (https://github.com/faceyspacey/redux-first-router/blob/b98a645a7eb29e8c419bb8ea29582022d657be91/src/connectRoutes.js#L465) it looks like it is re-creating the payload based on parsing the route into its parts. So if I had a routesMap like this:
```
const routesMap = {
    [constants.SET_MAIN_PAGE]: '/page/:mainPageRoute',
};
```
then I would end up in my reducer with the SET_MAIN_PAGE event payload containing only the `mainPageRoute` string and not the `mainPage` Page object.

Another thing that looked promising was the ability to add a `fromPath` and `toPath` to the routesMap (https://github.com/faceyspacey/redux-first-router/blob/master/docs/connectRoutes.md#routesmap) in order to do a transformation on the event payload before applying it to the route and then to the route string before making it back into an event payload. I think this is the way that we should be doing it, but it always expected a string coming in and a string coming out and couldn't seem to handle the `Page` enum. 

So this is what I came up with as a compromise. It's essentially the same as I had before, except
- I changed the numeric Page enum into a string enum
- Instead of doing a `.replace()` on the routesMap string, I just re-wrote the `/page/` part of the route in the `initalEntries` array. I'm not sure whether this is any better because if we decide to change `/page/` then we'll need to change it in more than one place.

Also, I changed the name of my branch and PR and that confused GitHub, so this is a new PR instead of just being a change to the previous one (https://github.com/pg-irc/pathways-frontend/pull/64).

What do you think?


rasmus-storjohann-PG:
Hi @trisapeace , thanks for really digging into this peace, there's a lot to work with here.

As I understand it, the router is all about translating URLs to application states. Eventually, we want to have deep routing, so that I can text you a URL something like "pathways:/tasks/get_drivers_lices#related_tasks" and the app will know to open up the relevant task and scroll to the relevant section. So I hope it makes sense that the router and router history will work just with URLs. I guess it could work with raw strings like the whole "pathways:/tasks/get_drivers_lices#related_tasks", but as you found, it actually works with parsed URLs. I guess the parsing happens according to the string '/page/:mainPage', where the parser (because of that ':') expects '/page/foo', where 'page' is a constant and whatever comes after it is assigned to the 'mainpage' property of the route action. 

So I think this means that we need to go with the flow and give the router URLs to work with. The thing is that we want to separate concerns, so that nothing in the rest of the app should know anything about URLs. So we need a layer that translates between the whatever representation we want to use for the rest of the app (an enum for now), and the URLs for the router. That's what `getPageFromString()` is for. So I think the signature on that function needs to be opposite from what you have, since it will translate from the app-wide representation of the enum, to the string '/page/foo'. Rename the function as appropriate.

Later when we start building deep linking, it will no longer work to have an enum to indicate what page we're on. We will be able to make that change without affecting the router, since `getPageFromString()` is sitting in between and translating from whatever representation we use then, to URL strings as before. Separation of concerns accomplished.

My hunch is that `fromPath` and `toPath` are probably useful for more complex cases than what we're working on now, so I would probably stay away from those for now,

src/application/router.ts
```diff
@@ -3,13 +3,16 @@ import * as reduxFirstRouter from 'redux-first-router';
...
 
 const routesMap = {
-    [constants.SET_MAIN_TAB]: '/user/:mainPage',
+    [constants.SET_MAIN_PAGE]: '/page/:mainPage',
 };
 
 export const createStoreWithRouter = (reducerForApplicationState: Reducer<StoreForApplicationState>) => {
-    const history = createMemoryHistory();
+    const history = createMemoryHistory({
+        initialEntries: ['/page/' + initialPage],
```
rasmus-storjohann-PG: So with my suggested function for translating from the enum to the URL string, this should then become
```
initialEntries: [getPageFromString(Page.Questionnaire)],
```

rasmus-storjohann-PG:
That's all my comments for now, I think. Please let me know if you have any questions. I really appreciate your help on this.

trisapeace:
@rasmus-storjohann-PG  @tnightingale @bensinjin 
I think this PR is ready for review.

- Each page type (Questionnaire, MyPlan, ExploreAll) now has its own Action constant to represent switching to that main page
- This prevents issues with stringifying the Page enum in the path when using something like `:mainPage` in the `routesMap`
- The router now now has a function that will map a Page enum into a string route. I don't see any need for any file outside the router to use this function, so I didn't export it. Any other function should be interacting with the route by triggering one of these three Actions
I found having the `setMainPage` function in `page_switcher.ts` sort of confusing. In an ideal situation we could be using the `<Link>` tags within the `.tsx` files (the reason that we aren't here is that we are using `<Button>` which cannot be wrapped with `<a>` tags, which is what `<Link>` does under the hood) and not even have to define these dispatches. So I moved them into the `index.ts` where they can be closer to where they are used
- Removed some other unused code
- Updated tests so they pass. However, I needed to change the jest regex before it would run any of  the tests. I'll make a different PR for that change so you can look at it and let me know what you think.
- I'm getting some linting errors about not having defined the type for my arrow functions in `index.js` but I'm not sure what I need to put. Does anyone have any suggestions?

src/application/router.ts
```diff
@@ -3,9 +3,25 @@ import  { LocationState } from 'redux-first-router';
...
+            return getRouteFromPage(initialPage);
+    }
+};
 
 const routesMap = {
-    [constants.SET_MAIN_TAB]: '/user/:mainPage',
+    [constants.SET_QUESTIONNAIRE_PAGE]: getRouteFromPage(Page.Questionnaire),
+    [constants.SET_PLAN_PAGE]: getRouteFromPage(Page.MyPlan),
+    [constants.SET_EXPLORE_PAGE]: getRouteFromPage(Page.ExploreAll),
 };
```
tnightingale: This is looking good; it's much better to have the base routes enumerated explicitly than to use dynamic parameters.

How do you feel about the relationship between `routesMap`/`getRouteFromPage()` and the `reducer` in `src/stores/page_switcher.ts`? I'm concerned there's some redundancy there. From what I understand it boils down to bi-directional mapping between URL paths and action types (`constants`). Because they're inherently dependant on each other I wonder if they should be put together? Maybe by moving `getRouteFromPage()` into the page_switcher store?

src/stores/page_switcher.ts
```diff
@@ -1,53 +1,32 @@
...
 
 export type Store = Readonly<ReturnType<typeof buildDefaultStore>>;
 
-export type SetMainPageAction = Readonly<ReturnType<typeof setMainPage>>;
-
-// tslint:disable-next-line:typedef
-export const setMainPage = (mainPage: Page | string) => (
-    helpers.makeAction(constants.SET_MAIN_TAB, { mainPage })
-);
+export const initialPage = Page.Questionnaire;
```
tnightingale: I wonder if this is going to stay hard-coded or whether we're going to want to add a layer to persist user's last page visited so that it survives a process restart (similar to the user's locale preference lookup we do during application bootstrap in `src/application/store.ts`)?

src/stores/page_switcher.ts
```diff
@@ -1,53 +1,32 @@
...
 
 // tslint:disable-next-line:typedef
 const buildDefaultStore = () => (
-    { mainPage: Page.Questionnaire }
+    { mainPage: initialPage }
```
bensinjin: Wonder if we can call this "currentPage"?

src/stores/page_switcher.ts
```diff
@@ -1,53 +1,32 @@
...
 );
 
-export const reducer = (store: Store = buildDefaultStore(), action?: SetMainPageAction): Store => {
+export const reducer = (store: Store = buildDefaultStore(), action?: any): Store => {
```
bensinjin: You can avoid "any" here by creating a type per action and then using the "|" union operator. See for example https://github.com/pg-irc/pathways-frontend/blob/master/src/stores/tasks.ts#L14.

bensinjin:
This work touches on some of concerns I had about the router currently so I've opted to close my review PR and talk here. I think this is a good direction and is similar to what I was thinking in terms of firing different actions for each page. My task detail work hopes to address dynamic path parameters ie. task/:taskId and will build nicely on this. Good work!

src/application/router.ts
trisapeace: I'm happy either way. That method is in `router.ts` because it was pointed out by @rasmus-storjohann-PG that it's nice to have all the strings that are actually going to be use in the route in one file so that there is just one place to look if you want to change something. By the time we're in the Page Switcher's reducer, we are working with the `Page` enum and not the route strings themselves. Does that work for you @tnightingale ?

src/stores/page_switcher.ts
trisapeace: Very good question. I didn't have the locale code yet when I started working on this branch. I'll take a look and see what was done there and see if we can do something similar here.

src/stores/page_switcher.ts
trisapeace: Sure, not a problem.

src/stores/page_switcher.ts
trisapeace: Cool. I'll check that out.

PR #77, opened 2018-05-27
Tweak coding conventions for discussion

rasmus-storjohann-PG:
See also https://hackernoon.com/react-components-naming-convention-%EF%B8%8F-b50303551505


bensinjin:
I feel this is clearer and personally prefer it to what we have. Have you also thought about our directory structure?

rasmus-storjohann-PG:
I'm going to keep this PR open, please ignore it until after the NVP is done, then we can discuss.

rasmus-storjohann-PG:
Changes along the lines of what is proposed here will be made after the MVP is done.

bensinjin:
Another thing I've started to do is to type our lists of things in selectors. We have a lot of ReadonlyArray<some-type-here> I think with types like "TaskList" or "TaskIdList" we have to think less about what these types should be in other places (primarily component props) and just import the list type from the selector.

bensinjin:
Some other things I've personally struggled to land on conventions for and would like to discuss:
- When rendering sub components of larger components, in order to keep components small and easier to follow, should we be using for example: `{renderHeading()}` or `<Heading />`. I tend to lean towards the latter.
- Do we add newlines between types and interfaces? do we try and group exports and private data?

PR #81, opened 2018-05-30
Issue #74 Ideal routing with an initial page DO NOT MERGE

trisapeace:
# DO NOT MERGE

@rasmus-storjohann-PG 
Based on your feedback and also what I've learned about the redux-first-router, this is what I think would be the most elegant solution. The theory being that whenever `SET_MAIN_PAGE` is called (whether by the links in the footer or by the Back button) then it will set the `mainPage` to a value from the `Page` enum. If that were so, then we'd have the mapping between the `string` used in the route to the `Page` used in the Action defined by the `toPath` and `fromPath` functions in `router.ts`. 

However, when I try to compile, I get this error message:
```
src/application/router.ts(39,60): error TS2345: Argument of type '{ [constants.SET_MAIN_PAGE]: { path: string; toPath: (page: Page) => string; fromPath: (pageRoute...' is not assignable to parameter of type 'RoutesMap<{ path
: string; toPath: (page: Page) => string; fromPath: (pageRoute: string) => Page; ...'.
  Property '[constants.SET_MAIN_PAGE]' is incompatible with index signature.
    Type '{ path: string; toPath: (page: Page) => string; fromPath: (pageRoute: string) => Page; }' is not assignable to type 'Route<{ path: string; toPath: (page: Page) => string; fromPath: (pageRoute: string) => Page; }, a
...'.
      Type '{ path: string; toPath: (page: Page) => string; fromPath: (pageRoute: string) => Page; }' is not assignable to type 'RouteObject<{ path: string; toPath: (page: Page) => string; fromPath: (pageRoute: string) => Pa
ge...'.
        Type '{ path: string; toPath: (page: Page) => string; fromPath: (pageRoute: string) => Page; }' is not assignable to type '{ capitalizedWords?: boolean; navKey?: string; path: string; thunk?: RouteThunk<any>; fromPat
h?(p...'.
          Types of property 'fromPath' are incompatible.
            Type '(pageRoute: string) => Page' is not assignable to type '(path: string, key?: string) => string'.
              Type 'Page' is not assignable to type 'string'.
```

I believe that it's telling me that the type signature of `toPath` and `fromPath` must be `(string) => string` whereas mine are `(string) => Page` and `(Page) => string`. Is there any way around this in Typescript?

If I don't use the `toPath` and `fromPath` methods, then the router just calls `toString()` on the enum values as they come into the path and then supplies them as `string` values when putting it back into the `mainPage` payload of the `SET_MAIN_PAGE` action when you use the Back button.

If you have time after sprint planning tomorrow, I'd love to talk this over.

(Note, since it's not compiling, I didn't bother to update the tests again to reflect these changes.)

### DIFF
--- a/src/application/constants.ts
+++ b/src/application/constants.ts
@@ -1,4 +1,4 @@
-export const SET_MAIN_TAB = 'ROUTER:SET_MAIN_TAB';
+export const SET_MAIN_PAGE = 'ROUTER:SET_MAIN_PAGE';
 
 export const SET_LOCALE = 'I18N:SET_LOCALE';
 
--- a/src/application/router.ts
+++ b/src/application/router.ts
@@ -3,13 +3,16 @@ import * as reduxFirstRouter from 'redux-first-router';
 import createMemoryHistory from 'history/createMemoryHistory';
 import * as constants from '../application/constants';
 import { Store as StoreForApplicationState } from '../stores';
+import { initialPage } from '../stores/page_switcher';
 
 const routesMap = {
-    [constants.SET_MAIN_TAB]: '/user/:mainPage',
+    [constants.SET_MAIN_PAGE]: '/page/:mainPage',
 };
 
 export const createStoreWithRouter = (reducerForApplicationState: Reducer<StoreForApplicationState>) => {
-    const history = createMemoryHistory();
+    const history = createMemoryHistory({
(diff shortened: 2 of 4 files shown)


trisapeace:
We weren't happy with this version either because it relies on the string version of the Page enum

PR #87, opened 2018-05-31
Issue #69 Fix devtools server config

tnightingale:
Updated:
- Gives Remote Redux DevTools access to our Redux store state when running in debugging mode.
- Removes `remotedev-rn-debugger` as it was causing build step to hang and doesn't work consistently anyway.

There are via two methods to access Redux DevTools remotely:
- Visit http://remotedev.io/local/ : (should "just work" on iOS and Android simulators via https/port 443).
- The Redux DevTools Chrome Extension (only works with iOS simulator, likely because of port forwarding issues on Android's Simulator):
  1. Enable remote debugging in Expo.
  2. Open Chrome's dev inspector (on macOS: option+cmd+i).
  3. Open the Redux tab press.
  4. Start a remote session (ctrl+cmd+downarrow on macOS).

~Add devserver config back to store creation (lost in a rebase conflict resolution).~

~I just noticed that redux devtools only works for me on iOS; I get socket connection errors in the console on Android. Seems like a common problem: https://github.com/zalmoxisus/remote-redux-devtools/search?q=socket+hung+up+android&type=Issues but the recommended fixes did not work for me.~

### DIFF
--- a/bin/postinstall.sh
+++ b/bin/postinstall.sh
@@ -10,4 +10,3 @@
 # https://github.com/aws/aws-sdk-js/issues/1926
 
 sed -i -e "s/\(^declare var require: NodeRequire;\)/\/\/\1/g" node_modules/\@types/node/index.d.ts
-yarn remotedev-debugger --hostname localhost --port 5678 --injectserver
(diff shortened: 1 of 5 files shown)


bensinjin:
This looks good. The current set up is actually breaking my local.

PR #88, opened 2018-06-01
Issue #44 Implement navigation to full task display (WIP)

bensinjin:
Hey @rasmus-storjohann-PG @tnightingale I wanted your opinions on this initial commit (f855c1a). It includes how I'm currently handling routing to the task detail page and should probably reflect how we expect to handle what I'm calling _page parameters_.  At this time I'm just following what was in place in terms of setting the mainPage value, have added a new entry for "TaskDetail", and am utilizing the page parameters to handle loading the task. Anyways lemme know your thoughts I would personally like to tweak the current implementation as outlined below.

### DIFF
--- a/src/components/main/index.ts
+++ b/src/components/main/index.ts
@@ -9,6 +9,8 @@ import { withI18n } from '@lingui/react';
 
 const mapStateToProps = (store: Store): main.Props => ({
     mainPageInProps: store.applicationState.mainPageInStore.mainPage,
+    pageParameters: store.applicationState.mainPageInStore.pageParameters,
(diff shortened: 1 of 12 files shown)


bensinjin:
I think what we should work towards is having our existing pages be at /questionnaire, /my-plan, /explore etc ... and have tasks be at tasks/:taskId. Currently it doesn't feel quite right having these "static" pages have the "dynamic" format : '/user/:mainPage'.

bensinjin:
It looks like much of this discussion is happening over here: https://github.com/pg-irc/pathways-frontend/pull/76. I'm going close this a talk there.

Issue #90, opened 2018-06-03
`yarn test` should run all the tests


trisapeace:
Right now, when I run `yarn test` or `yarn run jest`, it says that there are 0 matching file names for the `testRegex` string:
`testRegex: \\__tests__\\[\\w]+\.(test|spec)\.tsx?$ - 0 matches`

rasmus-storjohann-PG:
@trisapeace can you confirm that this issue has been fixed so we can close this story?

trisapeace:
Yes, this works. Thanks!

PR #91, opened 2018-06-03
Issue #90 Fix jest regex

trisapeace:
After some playing around, it seems to me that it doesn't understand the `\w` character set. When I replace that with a `.` then it manages to find and run all the tests. However, there is still a locale error that I don't understand. Here is the output:
```
$ yarn test
yarn run v1.7.0
$ yarn run build && yarn run jest
$ tsc
$ C:\Users\Theresa\PeaceGeeks\pathways-frontend\node_modules\.bin\jest
 PASS  src\sagas\__tests__\fonts.test.ts (5.803s)
 PASS  src\sagas\__tests__\locale.test.ts
  ● Console

    console.error src\sagas\locale.ts:42
      Failed to load current locale (43e96a8d-9bf8-4318-ae3b-7b4df3743940)

 PASS  src\stores\__tests__\questionnaire.test.ts
 PASS  src\stores\__tests__\tasks.test.ts
 PASS  src\stores\__tests__\tasks_helpers.test.ts
 PASS  src\selectors\__tests__\tasks.test.ts
 PASS  src\stores\__tests__\page_switcher.test.ts
 PASS  src\stores\__tests__\locale.test.ts
 PASS  src\selectors\__tests__\questionnaire.test.ts
 PASS  src\application\locale\__tests__\locale.test.ts
 PASS  src\stores\__tests__\questionnaire_helpers.test.ts
 PASS  src\stores\__tests__\fonts.test.ts

Test Suites: 12 passed, 12 total
Tests:       113 passed, 113 total
Snapshots:   0 total
Time:        11.243s, estimated 25s
Ran all test suites.
Done in 21.94s.
```

@tnightingale I can see that you were changing things with the `jest` command. Was it working for you with the `\w`?

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
     "transformIgnorePatterns": [
       "node_modules/(?!(react-native|expo|@expo/.*|native-base)/)"
     ],
-    "testRegex": "/__tests__/[\\w]+\\.(test|spec)\\.tsx?$",
+    "testRegex": "/__tests__/.+\\.(test|spec)\\.tsx?$",
     "moduleFileExtensions": [
       "ts",
       "tsx",
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
     "transformIgnorePatterns": [
       "node_modules/(?!(react-native|expo|@expo/.*|native-base)/)"
     ],
-    "testRegex": "/__tests__/.+\\.(test|spec)\\.tsx?$",
+    "testRegex": "/__tests__/[-a-zA-Z0-9_]+\\.(test|spec)\\.tsx?$",
     "moduleFileExtensions": [
       "ts",
       "tsx",


tnightingale:
Yeah that's weird, `\w` works for me both via `yarn test` and via vscode's task runner (configuration in `.vscode/launch.json`). It is a pretty standard regex character. I'm surprised that `.` works for you but `\w` doesn't.

What OS are you running `yarn install` on?

The locale error is just noting an exception that is caught and handled (correctly) during one of the tests.

trisapeace:
Thanks for taking a look @tnightingale . I'm on Windows 7, running yarn 1.7.0 and jest 22.4.3.

I had tried 
`[\\w]+`
`[\\w]*`
`(\\w)+`
`\\w+`
and none of them seemed to work. But the normal character classes did. So I've changed it now to be what `\w` means, and added the hyphen to the list because I could see a time when we use hyphens in test file names and we don't notice that the tests aren't being run. And that seemed to work. What do you think?

rasmus-storjohann-PG:
I can't reproduce this problem, all the tests run with `yarn test`. I do see the warning about a locale test, though
```
● Console

    console.error src/sagas/locale.ts:42
      Failed to load current locale (6f53175e-582d-470b-98ff-27d13b921b8a)
```

rasmus-storjohann-PG:
I'll merge this PR since it solves a problem for @trisapeace and doesn't seem to cause one for the rest of the team.

PR #92, opened 2018-06-05
Issue #61 Fix yarn clean by adding single quotes to glob arguments

rasmus-storjohann-PG:


rasmus-storjohann-PG:
@tnightingale @bensinjin @trisapeace any objections? It seems to make rimraf more reliable.

Issue #95, opened 2018-06-05
Separate "set locale" and "load current locale" sagas


tnightingale:
The existing "load current locale" saga implementation dispatches a "set locale" success action. This allowed it to use the existing reducer implementation for SET_LOCALE_SUCCESS. In hindsight this implementation is messy and a bit surprising.

We should treat these as two separate async actions (though the associated reducer can still update the same state).

tnightingale:
PR in #96

Issue #59, opened 2018-05-09
Translation of dynamic content


rasmus-storjohann-PG:
AC: When switching language, dynamic content appears in appropriate language.

Fixture needs to be populated with stubbed multilingual content

tnightingale:
I'm proposing we add a new type for translated content in our data model:
```typescript
interface LocalizedText {
    [localeCode: string]: string
}
```
I will use Task as an example:
```typescript
interface Task {
    readonly const id: Id;
    readonly const name: LocalizedText;
    readonly const content: LocalizedText;
    readonly const isRecommendedToAllUsers: boolean;
    readonly const isStarred: boolean;
    readonly const isComplete: boolean;
}
```
A representation would now look like this:
```json
{
    "id": "12345",
    "name": {
        "en": "the task name",
        "ar": "اسم المهمة",
        "zh": "任務名稱"
    },
    "content": {
        "en": "the task content",
        "ar": "محتوى المهمة",
        "zh": "任務內容"
    },
    "isRecommendedToAllUsers": false,
    "isInMyPlan": false,
    "isStarred": false,
    "isComplete": false
}
```
With this model we need to perform a transformation on all LocalizedText fields, something like:
```js
function selectLocalizedText (locale: Locale, localizedText: LocalizedText): string {
    return localizedText[locale.code]; // TODO: handle fallback locale (eg: via something like locale.default)
}
```

There are two options for where this transformation could occur:
1. During component render (the component would also require locale info eg: current locale code and, if supported, fallback locale).
2. During data fetching from store, for eg: in a selector (it would first get locale info from elsewhere in the store).
 
Option 1 requires getting the current locale from somewhere, either by fetching it from our store (using a selector and a connected component somewhere in the component tree) or fetching it from jsLingui's `I18n` context object (we would need to wrap relevant components with jsLingui's `withI18n` HOC).

Option 2 implies a second type for all of our localizable entities:
```typescript
interface LocalizedTask {
    readonly const id: Id;
    readonly const name: string;
    readonly const content: string;
    readonly const isRecommendedToAllUsers: boolean;
    readonly const isStarred: boolean;
    readonly const isComplete: boolean;
}
```

_I am currently leaning towards option 2 as it keeps the localization logic out of our components as much as possible._

[...]

PR #96, opened 2018-06-05
Issue #95 Seperate "set locale" and "load current locale" sagas

tnightingale:
This tidies up the locale sagas implementation and fixes an issue on master where we get stuck on the loading screen (because the reducer currently handles the wrong action on initial startup (before a users locale preference has been persisted to database).

### DIFF
--- a/src/application/locale/index.ts
+++ b/src/application/locale/index.ts
@@ -57,6 +57,10 @@ export class LocaleManager {
         return this.instance.locale(code);
     }
 
+    static getFallbackLocale(): Locale {
+        return this.instance.getFallbackLocale();
+    }
+
     static get locales(): ReadonlyArray<Locale> {
         return this.instance.locales;
     }
@@ -80,6 +84,8 @@ export class LocaleManager {
         return this.localeManagerInstance;
     }
 
+    private fallbackLocaleCode: string;
+
     private locales: ReadonlyArray<Locale>;
 
     private catalogsMap: CatalogsMap;
@@ -95,6 +101,15 @@ export class LocaleManager {
             throw new Error(`Unknown locale code: ${code}`);
         }
         return locale;
+    }
+
+    private getFallbackLocale(): Locale {
+        if (this.fallbackLocaleCode) {
+            return this.locale(this.fallbackLocaleCode);
+        } else {
+            const [ fallbackLocale ]: ReadonlyArray<Locale> = this.locales;
+            return fallbackLocale;
+        }
     }
 
 }
--- a/src/sagas/locale.ts
+++ b/src/sagas/locale.ts
@@ -26,17 +26,13 @@ export function* watchLoadLocale(): IterableIterator<ForkEffect> {
     yield takeLatest(constants.LOAD_CURRENT_LOCALE_REQUEST, loadCurrentLocale);
 }
 
-type LoadCurrentLocaleActions = LoadCurrentLocale.Request | LoadCurrentLocale.Result | SetLocale.Success;
+export type LoadCurrentLocaleActions = LoadCurrentLocale.Request | LoadCurrentLocale.Result | SetLocale.Success;
 
 export function* loadCurrentLocale(): IterableIterator<CallEffect | PutEffect<LoadCurrentLocaleActions>> {
     try {
-        const currentLocaleCode = yield call(loadCurrentLocaleCode);
-        if (currentLocaleCode !== null) {
-            const currentLocale = LocaleManager.get(currentLocaleCode);
-            yield put(loadCurrentLocaleActions.success(currentLocale));
-        } else {
-            yield put(loadCurrentLocaleActions.failure('No locale preference set'));
-        }
+        const code = yield call(loadCurrentLocaleCode);
+        const locale = code !== null ? LocaleManager.get(code) : LocaleManager.getFallbackLocale();
+        yield put(loadCurrentLocaleActions.success(locale));
     } catch (e) {
         console.error(`Failed to load current locale (${e.message})`);
         yield put(loadCurrentLocaleActions.failure(e.message));
--- a/src/sagas/locale.ts
+++ b/src/sagas/locale.ts
@@ -34,7 +34,6 @@ export function* loadCurrentLocale(): IterableIterator<CallEffect | PutEffect<Lo
         if (currentLocaleCode !== null) {
             const currentLocale = LocaleManager.get(currentLocaleCode);
             yield put(loadCurrentLocaleActions.success(currentLocale));
-            yield put(setLocaleActions.success(currentLocale));
         } else {
             yield put(loadCurrentLocaleActions.failure('No locale preference set'));
         }
--- a/src/stores/__tests__/locale.test.ts
+++ b/src/stores/__tests__/locale.test.ts
@@ -2,6 +2,7 @@
 import { Locale } from '../../application/locale';
 import * as locale from '../locale';
 import * as constants from '../../application/constants';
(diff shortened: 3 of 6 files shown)


tnightingale:
@bensinjin @rasmus-storjohann-PG would be great to get this merged quickly as it fixes a bug in master. The changes are pretty minor.

bensinjin:
Looks fine to me.

src/stores/__tests__/locale.test.ts
```diff
@@ -154,10 +161,26 @@ describe('the reducer', () => {
...
+
+    it('when called with LOAD_CURERNT_LOCALE_FAILURE should return store with loading flag set false', () => {
+        const errorMessage = aString();
+        const theStore = buildStoreLoadingLocale();
+        const theAction = {
+            type: constants.LOAD_CURRENT_LOCALE_FAILURE as typeof constants.LOAD_CURRENT_LOCALE_FAILURE,
+            payload: { message: errorMessage, locale: aLocale },
+        };
+        const theNewStore = locale.reducer(theStore, theAction);
+        expect(theNewStore.loading).toBe(false);
```
rasmus-storjohann-PG: What happens with the error message, is it part of the store after?

src/stores/locale/index.ts
```diff
@@ -25,14 +25,21 @@ export const reducer = (store: Store = buildDefaultStore(), action?: ReducerActi
         return store;
     }
     switch (action.type) {
+
         case constants.LOAD_CURRENT_LOCALE_REQUEST:
             return { ...store, loading: true };
+        case constants.LOAD_CURRENT_LOCALE_SUCCESS:
+            return { ...store, code: action.payload.locale.code, loading: false };
+        case constants.LOAD_CURRENT_LOCALE_FAILURE:
+            return { ...store, loading: false };
```
rasmus-storjohann-PG: This action leaves a stale code in the state, it should probably be cleared? Same below.

tnightingale: In this case, it's not stale; the existing (or default locale is used).

However, as mentioned above, the error message on the action (`action.payload.message`) should be recorded and, short of anything better, should be surfaced to the user somehow.

src/stores/__tests__/locale.test.ts
tnightingale: It is not currently being put in the store (see locale reducer). I avoided tackling this until we have a clear path for handling this sort of thing. Maybe this should be outlined/discussed in a followup issue?

As noted below, unless we can think of something better, the least we can do is display a message to the user indicating that something went wrong.

src/sagas/locale.ts
```diff
@@ -34,7 +34,6 @@ export function* loadCurrentLocale(): IterableIterator<CallEffect | PutEffect<Lo
         if (currentLocaleCode !== null) {
             const currentLocale = LocaleManager.get(currentLocaleCode);
             yield put(loadCurrentLocaleActions.success(currentLocale));
-            yield put(setLocaleActions.success(currentLocale));
         } else {
             yield put(loadCurrentLocaleActions.failure('No locale preference set'));
```
tnightingale: It could be that 'failure' is too strong here. Instead we maybe we should just succeed with a the default (or `null`/`undefined`) locale code and reserve failure for an actual exception during loadCurrentLocaleCode (eg: a I/O related issue).

tnightingale:
> I've also noticed in the app that changing the locale from e.g. English to Chinese (i.e. without changing the RTL state) also is asynchronous, even though the app reboot is not needed. Would it be possible or easy to make that change run faster?

This is asynchronous as the user's preference needs to be persisted to storage (an asynchronous action).

But you are right, in cases where layout change (reload) isn't required, we don't need to block the (re)render while the preference is persisted. It currently does block for consistency/simplicity but it shouldn't be difficult to make the saga / reducer logic a bit smarter.

src/stores/__tests__/locale.test.ts
rasmus-storjohann-PG: We should think about error handling. Until then, TDD should encourage us to only write code that has observable (i.e. testable) effects. So, if we were using TDD, which I encourage, then the error message would not be there in the test or production code. This helps us focus our attention on always writing code that has value to our users. I'm not asking for any changes now, just using this as an opportunity to plug TDD. I don't tend to pass up on those much ;-)

src/sagas/locale.ts
rasmus-storjohann-PG: Yeah, isn't `currentLocaleCode === null` actually completely valid, e.g. when the app is first installed?

tnightingale: Yep!

src/stores/__tests__/locale.test.ts
tnightingale: That's a fair point. At the time of writing, my thinking was that not handling exceptions (or worse; silently suppressing them) is best avoided because they can easily be forgotten (hence the `console.error()` popping up in tests right now). In the future I'm wondering whether it would be better to just not include the `try { ... } catch(e) { ... }` so that any exceptions cause application to crash? (that way we're less likely to ignore/forget about it and we can add the error handling when we have a plan for how to implement it).

In saying all that, the error handling in this piece of functionality (`loadCurrentLocale()` and `setLocale()`) is incomplete and could definitely be improved.

tnightingale:
Ok, I have partially addressed error handling by making sure we store any potential error messages in the reducer. While not functional, this at least provides some justification for the existence of the message property in the async failure actions (and associated tests). We can build upon this in: #98 when the time is right.

I have created #101 with an explanation of how we can improve the language switching experience if toggling between languages that are have the same layout/text direction. I have also pushed a rough implementation to feature/101-improve_switch_locale.

Finally, I have added the initial implementation of a 'fallback locale' to `LocaleManager` this lets our saga handle the (non-error) case where `loadCurrentLocale()` returns `null` without treating it like an error has occurred. It will also allow me to address feedback in #59 around leaking locale concerns into unrelated test builders and such.

So, can we please merge this so I can get on to finalizing #59? :)

src/application/locale/index.ts
```diff
@@ -97,6 +103,15 @@ export class LocaleManager {
         return locale;
     }
 
+    private getFallbackLocale(): Locale {
+        if (this.fallbackLocaleCode) {
+            return this.locale(this.fallbackLocaleCode);
+        } else {
+            const [ fallbackLocale ]: ReadonlyArray<Locale> = this.locales;
```
rasmus-storjohann-PG: what does this do?

src/stores/locale/index.ts
```diff
@@ -25,14 +26,25 @@ export const reducer = (store: Store = buildDefaultStore(), action?: ReducerActi
         return store;
     }
     switch (action.type) {
+
         case constants.LOAD_CURRENT_LOCALE_REQUEST:
             return { ...store, loading: true };
+        case constants.LOAD_CURRENT_LOCALE_SUCCESS:
+            return { ...buildDefaultStore(), code: action.payload.locale.code };
```
rasmus-storjohann-PG: Interesting use of `buildDefaultStore()`, let's see if this is a pattern we want to use more widely. My hunch is probably not, but I could be wrong.

src/application/locale/index.ts
tnightingale: It takes the first value from the array.

Definitely happy to discuss how (or if) we want to enforce a fallback locale going forward but I figured grabbing the first registered locale would be sufficient until we have a better plan.

src/stores/locale/index.ts
tnightingale: Yeah fair enough, it keeps the possible states obvious for the time being.

PR #97, opened 2018-06-06
#59 Localized content

tnightingale:
_Note: If you have a fresh install of the app (Expo) on your test device (or simulator) you will need the changes from #96 otherwise your app might not get past the loading screen._ 

Contains conversion of Task (and related selectors) as described in #59 (option 2).


tnightingale:
@bensinjin would be particularly good to get your feedback on this (described in #59) sooner than later.

tnightingale:
@rasmus-storjohann-PG I've also added localization to Questionnaire types, please take a look when you have time.

_Note that if you have a fresh install of the app (Expo) on your test device (or simulator) you will need the changes from #96 otherwise your app might not get past the loading screen._

src/selectors/questionnaire.ts
```diff
@@ -1,41 +1,43 @@
...
     const { questions, answers }: model.Store = modelStore;
 
     return Object.keys(questions).map((key: string) => {
         const { id, text }: model.Question = questions[key];
         return {
             id,
-            text,
-            answers: selectAnswersForQuestion(id, answers),
+            text: selectLocalizedText(locale, text),
+            answers: selectAnswersForQuestion(locale, id, answers),
```
rasmus-storjohann-PG: It's tempting to curry `selectLocalizedText()`, so you can do this
```ts
    text: selectLocalizedText(text),
    answers: selectAnswersForQuestion(selectLocalizedText, id, answers),
```

src/stores/__tests__/helpers/locale_helpers.ts
```diff
@@ -45,3 +42,25 @@ class LocaleBuilder {
...
     }
 }
+
+export class LocalizedTextBuilder {
+
+    localizations: LocalizedText;
+
+    constructor(defaultLocaleCode: string, defaultLocaleText: string) {
+        this.addLocale(defaultLocaleCode, defaultLocaleText);
+    }
```
rasmus-storjohann-PG: Builder constructors should only take arguments if they cannot generate plausible random values themselves. One example is database foreign keys, there are a few examples of that in the backend repo. But here, I think you can just do random and make the builder a lot easier to use.

src/fixtures/questionnaire.ts
```diff
@@ -1,8 +1,10 @@
+import { LocalizedText } from '../application/locale';
```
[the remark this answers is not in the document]
tnightingale: Noted. As this is evolving, I think I would describe the `src/application/locale` module as "low-level". So it probable needs to be moved elsewhere. Maybe to a top-level module (ie: under `src/`)?

src/selectors/questionnaire.ts
```diff
@@ -1,41 +1,43 @@
...
-export interface Question {
+export interface LocalizedQuestion {
     readonly id: model.Id;
     readonly text: string;
-    readonly answers: ReadonlyArray<Answer>;
+    readonly answers: ReadonlyArray<LocalizedAnswer>;
 }
 
-export interface Answer {
+export interface LocalizedAnswer {
```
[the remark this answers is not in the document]
tnightingale: Sure, I figured you'd probably suggest this. I named it such to make it obvious what was being changed (and to keep track of things myself). Happy to invert the naming now though.

src/selectors/questionnaire.ts
```diff
@@ -1,41 +1,43 @@
...
     readonly id: model.Id;
     readonly text: string;
     readonly isSelected: boolean;
 }
 
-export const selectQuestionnaire = (appStore: app.Store): Questionnaire => (
-    denormalizeQuestions(appStore.applicationState.questionnaireInStore)
+export const selectQuestionnaire = (locale: Locale, appStore: app.Store): Questionnaire => (
+    denormalizeQuestions(locale, appStore.applicationState.questionnaireInStore)
```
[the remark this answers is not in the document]
tnightingale: I can do this but I'd like to point out that what is in the store is not a `Locale` object, it is a `localCode`. 

The `mapStateToProps()` function selects the `localeCode` (using a selector) and then gets the `Locale` object via `LocaleManager.get(localeCode)` (making it an 'impure' function). I can move the `LocaleManager.get()` call into the selector but it will make our selectors no-longer 'pure' functions.

No matter what, this feels a bit messy and bends conventional Redux 'rules'; I was just trying to contain the mess a little bit.

src/selectors/questionnaire.ts
tnightingale: I agree, but we would still need to pass the curried (stateful) `setLocalizedText` function in, eg:
```typescript
function denormalizeQuestions(selectLocalizedText: (text: LocalizedText): string, modelStore: model.Store);
```
I'm not sure if this is a meaningful improvement...

src/stores/__tests__/helpers/locale_helpers.ts
tnightingale: As mentioned elsewhere, in cases where we're testing against the `LocalizedText` values I need to dictate the localeCode in order to know what the resulting `LocalizedText` value will be. I agree, it forces unrelated things to have to consider locale. There should be a happy middle-ground here (maybe with default parameters?) I will take another look and see if I can clean it up.

src/selectors/questionnaire.ts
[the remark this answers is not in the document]
rasmus-storjohann-PG: You wanna just store the whole current Locale in the store? I'd be fine with that I think.

I think the reason this feels messy is that the `LocaleManager` owns the locale table. If it lived in the store instead, this would feel much more natural. Perhaps the `LocaleManager` could just be an app boot time thing that build the data structures but then they end up in the store.

src/selectors/questionnaire.ts
rasmus-storjohann-PG: I agree, it was just an idea...

src/stores/__tests__/helpers/locale_helpers.ts
rasmus-storjohann-PG: When things are hard to test, it indicates that there is a architecture problem. Do these tests hit the LocaleManager singleton in this case? If so, that would be pretty smelly. Singletons can be difficult in that way. I think there's a refactor waiting to happen here, but it can be pushed out for now.

tnightingale: No, they currently do not because we don't test `mapStateToProps()` (or, connected components).

tnightingale:
Ok I'm happy with where this is at. There are a *lot* of lines changed here, but the scope is not as large as it looks.

_FYI I have now actually rebased this branch on  #95 (PR #96) so that really needs to get in first._

I have moved the locale module (`src/application/locale`) into `src/` to signify that it is a low-level module that several other modules depend on.

I have reduced our dependance on the LocaleManager singleton (now called `LocaleDefinitionManager`). It is now only used in the following places:
- On application startup to register locales and establish a default/fallback.
- In the `ConnectedLocaleSwitcher` component to get extra locale metadata like label, isRTL, etc.
- In the `applyLocaleChanges()` saga to determine if a layout update/application restart is required before changing the locale.

Locale-aware selectors now grab the locale info from the store themselves. They do not need to consult the `LocaleDefinitionManager`. I have also included `localeInStore.fallbackCode` in the Redux store  to give our selectors an opportunity to fallback to a "base locale" (eg: english) if no translation exists for the string they're looking for. This value is intended to be set at application startup and not change. In the case where there is also no translation in the "base locale" we throw an error but we could just return an empty string, I'm not sure what is better (the chances of this happening _should be_ slim).

Finally, I have updated names of a bunch of the symbols (types, methods, variables, etc). Apologies as this produces most of the noise in the pull request but I feel the new names communicate their purpose better.

tnightingale:
Ok so I decided I might as well go all the way and rejig the locale definition stuff so that the Catalog info isn't included in the base locale info. This makes it possible to put info about all available locales into the store. The idea is that, like the `fallbackCode` value, `availableLocales` is set at application startup and never changed.

Now `ConnectedLocaleSwitcher` no longer needs to access the `LocaleInfoManager` (formally `LocaleDefinitionManager`, formally `LocaleManager`), it just uses the new `selectAvailableLocales()` selector to fetch them from the store.

I have slightly modified the `applyCurrentLocale()` saga and its side-effects. The side-effect signatures have changed slightly allowing me to push the RTL layout lookup into the new `setTextDirection()` function which feels like a better place for it.

Finally, there is some more cleanup of names and I have removed unused types, imports, test helpers  and methods.

I'm not going to look at or think about this any more. It is done :)

src/stores/locale/index.ts
```diff
@@ -25,14 +34,25 @@ export const reducer = (store: Store = buildDefaultStore(), action?: ReducerActi
...
     }
     switch (action.type) {
+
         case constants.LOAD_CURRENT_LOCALE_REQUEST:
             return { ...store, loading: true };
+        case constants.LOAD_CURRENT_LOCALE_SUCCESS:
+            return { ...buildDefaultStore(), code: action.payload.localeCode };
+        case constants.LOAD_CURRENT_LOCALE_FAILURE: {
+            const payload = action.payload;
+            return { ...buildDefaultStore(), errorMessage: payload.message };
```
rasmus-storjohann-PG: I am confused about the call to `buildDefaultStore()` here, it means you are throwing out all the state. Don't you want to hang on to things like `availableLocales` and such?

tnightingale: Yeah good catch, that definitely is a bug. I'll fix now.

tnightingale: (and update the tests accordingly)

tnightingale:
Ok @rasmus-storjohann-PG those last few commits should resolve most of the things you mentioned. I think the remaining bits are aesthetic and/or require more discussion and are non-blocking.

tnightingale:
Rebased on master

PR #115, opened 2018-06-11
Add tests for locale selectors

tnightingale:
Includes a ApplicationStoreBuilder helper.

### DIFF
--- a/src/selectors/__tests__/locale.test.ts
+++ b/src/selectors/__tests__/locale.test.ts
@@ -1,0 +1,72 @@
+// tslint:disable:no-let no-expression-statement
+import { LocaleStoreBuilder, LocalizedTextBuilder, LocaleBuilder } from '../../stores/__tests__/helpers/locale_helpers';
+import { selectLocale, selectAvailableLocales, selectLocalizedText } from '../locale';
+import { ApplicationStoreBuilder } from '../../stores/__tests__/helpers/store_helpers';
+import { Store } from '../../application/store';
+import * as locale from '../../stores/locale';
+import { aString } from '../../application/__tests__/helpers/random_test_values';
+import { Locale } from '../../locale';
+
+describe('locale selectors fetch', () => {
+
+    let theStore: Store;
+    let aLocaleStore: locale.Store;
+
+    beforeEach(() => {
+        aLocaleStore = new LocaleStoreBuilder().build();
+        theStore = {
+            applicationState: new ApplicationStoreBuilder().withLocaleStore(aLocaleStore).build(),
+        };
+    });
+
+    it('the locale code', () => {
+        const theLocale = selectLocale(theStore);
+        expect(theLocale.code).toBe(aLocaleStore.code);
+    });
+
+    it('the fallback locale', () => {
+        const theLocale = selectLocale(theStore);
+        expect(theLocale.fallback).toBe(aLocaleStore.fallback);
+    });
+
+    it('the available locales', () => {
+        const availableLocales = selectAvailableLocales(theStore);
+        expect(availableLocales).toBe(aLocaleStore.availableLocales);
+    });
+
+});
+
+describe('localized text selector', () => {
+
+    let theLocale: Locale;
+    let theText: string;
+    let theFallbackLocaleCode: string;
+    let theFallbackText: string;
+
+    beforeEach(() => {
+        theLocale = new LocaleBuilder().build();
+        theText = aString();
+        theFallbackLocaleCode = aString();
+        theFallbackText = aString();
+    });
+
+    it('selects the correct localized version of the text', () => {
+        const theLocalizedText = new LocalizedTextBuilder(theLocale.code, theText).build();
+        expect(selectLocalizedText(theLocale, theLocalizedText)).toBe(theText);
+    });
+
+    it('selects text from the fallback locale if the current locale is not available', () => {
+        const theLocalizedText = new LocalizedTextBuilder(theLocale.code, theText)
+                .addLocale(theFallbackLocaleCode, theFallbackText)
+                .build();
+        const aLocale = new LocaleBuilder().withFallback(theFallbackLocaleCode).build();
+        expect(selectLocalizedText(aLocale, theLocalizedText)).toBe(theFallbackText);
+    });
+
+    it('throws an error if the text is not available in the fallback locale', () => {
+        const theLocalizedText = new LocalizedTextBuilder(theLocale.code, theText).build();
+        const aLocale = new LocaleBuilder().build();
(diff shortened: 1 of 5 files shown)


rasmus-storjohann-PG:
We don't need tests for all builders, only those that are complex enough that bugs can hide in them.

src/selectors/__tests__/locale.test.ts
```diff
@@ -0,0 +1,72 @@
...
+    let theText: string;
+    let theFallbackLocaleCode: string;
+    let theFallbackText: string;
+
+    beforeEach(() => {
+        theLocale = new LocaleBuilder().build();
+        theText = aString();
+        theFallbackLocaleCode = aString();
+        theFallbackText = aString();
+    });
```
rasmus-storjohann-PG: I was curious, it turns out that this can be simplified to
```
    let theLocale: Locale = new LocaleBuilder().build();
    let theText: string = aString();
    let theFallbackLocaleCode: string = aString();
    let theFallbackText: string = aString();
```
This does mean that the same values are used for all the tests in the suite, but I don't think that is a big problem.

src/selectors/__tests__/locale.test.ts
```diff
@@ -0,0 +1,72 @@
...
+                .addLocale(theFallbackLocaleCode, theFallbackText)
+                .build();
+        const aLocale = new LocaleBuilder().withFallback(theFallbackLocaleCode).build();
+        expect(selectLocalizedText(aLocale, theLocalizedText)).toBe(theFallbackText);
+    });
+
+    it('throws an error if the text is not available in the fallback locale', () => {
+        const theLocalizedText = new LocalizedTextBuilder(theLocale.code, theText).build();
+        const aLocale = new LocaleBuilder().build();
+        expect(() => selectLocalizedText(aLocale, theLocalizedText)).toThrow();
```
rasmus-storjohann-PG: I find this test hard to follow. So you have a locale with code `foo` and fallback 'bar', and then you pass in localized text where there is no mapping from `foo` or `bar`. How about this:
```ts
    it('throws an error if the text is not available in the fallback locale', () => {
        const mainCode = aString();
        const fallbackCode = aString();
        const aLocale = new LocaleBuilder().withCode(mainCode).withFallback(fallbackCode).build();

        const someOtherCode = aString();
        const theLocalizedText = new LocalizedTextBuilder(someOtherCode, aString()).build();

        expect(() => selectLocalizedText(aLocale, theLocalizedText)).toThrow(
            `Missing text for locale: ${mainCode} or fallback ${fallbackCode}`,
        );
    });
```

tnightingale:
I meant to note that I wasn't sure how useful the `ApplicationStoreBuilder` would be (or whether there might be a better way to go about doing it). So it currently only has a method for setting the locale store. I figure we can expand as we need.

We also still have some weirdness where applicationStore is nested one level in the main Redux store. The selectors expect the whole store so the tests currently wrap `ApplicationStoreBuilder`'s output with the extra layer.

src/stores/__tests__/helpers/store_helpers.ts
```diff
@@ -0,0 +1,30 @@
...
+    build(): Store {
+        return {
+            fontsInStore: this.fontsInStore,
+            localeInStore: this.localeInStore,
+            currentPageInStore: this.currentPageInStore,
+            questionnaireInStore: this.questionnaireInStore,
+            tasksInStore: this.tasksInStore,
+        };
+    }
+}
```
rasmus-storjohann-PG: I understand why you created this builder to test the locale code. However, I think that the locale code is a bit unique in that it involves multiple top-level store components. I am hoping that will not run into that situation often, and so my hope is that this builder would see very limited use. So if I was doing this, I'd just create the store using an object literal in the place where you need it and not bother with the builder.

rasmus-storjohann-PG:
Also rename the commit to refer to issue #59

src/stores/__tests__/helpers/store_helpers.ts
tnightingale: Currently the locale selectors take the whole store object. But I thought the intention was for all of our selectors to do so. If we make that change (or build new selectors accordingly then we're going to get a fair amount of use out of this builder (or something similar).

I guess the bigger question is, how much work do we want to do in our `mapStateToProps()` implementations (which don't currently get tested)?

src/selectors/__tests__/locale.test.ts
tnightingale: Good suggestion, I'll update (but I probably won't assert the error message).

tnightingale:
@rasmus-storjohann-PG see last commit for improvements on LocalizedText selector tests. I think they should be clearer (and simpler).

src/stores/__tests__/helpers/store_helpers.ts
rasmus-storjohann-PG: The plan is that all selectors take the whole store as argument. Still, it would be nice to be able to test selectors without building up a whole store each time. One solution is to follow the pattern of `selectQuestionnaire()`, where the exported selector function is trivial and needs no tests, and it calls a helper function (in that case `denormalizeQuestions()`) that is fully tested. 

By the single responsibility principle, `mapStateToProps()` should never do real work, since it's job is just to define a mapping. The selector pattern is a very good way to make `mapStateToProps()` implementations readable.

src/selectors/__tests__/locale.test.ts
rasmus-storjohann-PG: It's your call, but I sometimes feel a little uncertain with a test that just asserts that "something is thrown" because there could be all sorts of things that could throw. Having some check that the expected thing is thrown makes me feel better, even if it's just a substring on the error message. To be clear, the test is not that the exception has exactly this error message, it is that an exception was thrown for the expected reason.

src/stores/__tests__/helpers/locale_helpers.ts
```diff
@@ -63,7 +63,7 @@ export class LocalizedTextBuilder {
 
     localizations: LocalizedText;
 
-    constructor(defaultLocaleCode: string, defaultLocaleText: string) {
+    constructor(defaultLocaleCode: string = aString(), defaultLocaleText: string = aString()) {
```
rasmus-storjohann-PG: I won't block the PR, but I have two problems with this: I think constructor arguments to builders should be avoided when possible, and I think default arguments to functions really hurts readability. But it's a small thing, let's move on.

rasmus-storjohann-PG:
@tnightingale I'm fine with merging this PR as is, but I'll leave it to you to decide whether you want to make any additional adjustments or just merge.

Issue #82, opened 2018-05-30
Style list of tasks in my plan


rasmus-storjohann-PG:
AC: Style My Plan header and tasks in My Plan as follows:

![image](https://user-images.githubusercontent.com/32307781/40735198-6c854022-63ef-11e8-85fa-cd93ea56904b.png)

Out of scope: Do not build any new functionality, such as event handling, as part of this story.

Heads up for future changes, all of this is out of scope for this task and for your information only: The TASKS I PLAN TO DO string will be replaced with a different string for tasks in different states. The star icon (which is not a button) represents tasks that are recommended, it will not appear on tasks that are not recommended. The eight dots will be used as a handle for dragging tasks to reorder them.

From page 13 of our current mocks, for context this is the whole screen:

![image](https://user-images.githubusercontent.com/32307781/40735240-86457932-63ef-11e8-971e-3d99e4698dbb.png)

bensinjin:
Ticket #83 is now included in this work.

rasmus-storjohann-PG:
BUG
* In the task detail page, in RTL, the Information and Find Services tabs are shown in the same order as in LTR.

bensinjin:
https://github.com/GeekyAnts/NativeBase/issues/518 discusses this

rasmus-storjohann-PG:
Can we make this manually, i.e. in RTL create the Services tab first so it ends up on the left, and open the Task tab explicitly?

bensinjin:
Unfortunately it's not that simple it's actually the RTL setting that breaks how the tabs work.  The ability to explicitly open a tab doesn't work properly in RTL and we get other funny issues like tabs rendering completely empty.

rasmus-storjohann-PG:
Closing this story, tracking bug in #135

PR #116, opened 2018-06-11
Feature/issue 44 task detail

bensinjin:
Styling for task detail page. This PR also defines some application wide styles that will be used throughout other styling Issues.

Things to note: 
- Currently you need to click the ">" button from the task list item to navigate to the task detail display. I've improved this UX a bit (the whole row is clickable) in Issue_82_83_style_task_lists which is based off this branch.
- I've not bothered to style the tab colors as they by default use the OS's color scheme which I ruled was preferable as we've not actually landed on a design yet
- Some icons will render different on different operating systems and the designs don't necessarily take this into account. Currently the icons should render their OS specific version of the icon.


src/components/tasks/connected_saved_tasks.ts
```diff
@@ -16,6 +17,7 @@ const mapDispatchToProps = (dispatch: Dispatch<Store>): Actions => ({
     toggleCompleted: (taskUserSettingsId: stores.Id): stores.ToggleCompletedAction => dispatch(stores.toggleCompleted(taskUserSettingsId)),
     toggleStarred: (taskUserSettingsId: stores.Id): stores.ToggleStarredAction => dispatch(stores.toggleStarred(taskUserSettingsId)),
     shareTask: (): stores.ShareAction => dispatch(stores.share()),
+    goToTaskDetail: (taskId: stores.Id): SetTaskDetailPageAction => dispatch(setTaskDetailPage(taskId)),
```
tnightingale: I'm surprised about the return types here. Does `dispatch()` return the action dispatched?

bensinjin: Yep https://redux.js.org/api-reference/store#dispatch

rasmus-storjohann-PG: I was surprised too

src/components/main/main_page_switcher.tsx
```diff
@@ -23,6 +30,15 @@ export const Component: React.StatelessComponent<Props & Actions> = (props: Prop
         case store.Page.ExploreAll:
             return <exploreAll.Component />;
 
+        case store.Page.TaskDetail:
+            return <TaskDetail
+                        task={selectTaskById(
+                                props.locale,
+                                props.tasksStore,
+                                props.currentPageParameters.taskId)
+                            }
```
rasmus-storjohann-PG: The call to `selectTaskById()` should be in a `mapStateToProps()`. I guess that would mean that we would have a `ConnectedTaskDetail` here. This is part of a bigger discussion, outside the scope of this PR (and of the MVP.) Just if you have ideas in the next while on how we would do this, make sure to hang on to those ideas, we'll need them.

src/components/tasks/task_list.tsx
```diff
@@ -65,6 +67,7 @@ export const renderSuggestedTask = (item: selector.Task, actions: TaskActions):
             tags={item.tags}
             addToSavedList={actions.addToSavedList}
             shareTask={actions.shareTask}
+            goToTaskDetail={actions.goToTaskDetail}
```
rasmus-storjohann-PG: This could also be
```ts
goToTaskDetail={() => actions.goToTaskDetail(item.id)}
```
which would make the call site much nicer

src/components/tasks/actions.ts
```diff
@@ -8,4 +9,5 @@ export interface TaskActions {
     readonly toggleCompleted?: (taskUserSettingsId: stores.Id) => stores.ToggleCompletedAction;
     readonly toggleStarred?: (taskUserSettingsId: stores.Id) => stores.ToggleStarredAction;
     readonly shareTask?: () => stores.ShareAction;
+    readonly goToTaskDetail?: (taskId: stores.Id) => SetTaskDetailPageAction;
```
rasmus-storjohann-PG: Why the `?`, when will this ever not be defined?

src/components/tasks/task.tsx
```diff
@@ -33,6 +34,7 @@ export const Task: React.StatelessComponent<selector.Task & Actions> = (props: s
             <View style={styles.sideColumn}>
                 <View style={styles.stackedItems}>
                     <Button
+                        onPress={(): SetTaskDetailPageAction => props.goToTaskDetail(props.id)}
```
rasmus-storjohann-PG: By the type, `props.goToTaskDetail` can be undefined, except that I don't think it actually can. Either add a guard here or remove the `?` elsewhere.

rasmus-storjohann-PG: Also, the `arrow-forward` icon should switch directions in Arabic. I thought maybe the OS and Expo were smart enough to know that `arrow-forward` is a left pointing arrow in RTL, but apparently not. We should have a helper function that takes the current locale and returns the appropriate icon for forward and backward arrows.

src/components/tasks/task_detail.tsx
```diff
@@ -0,0 +1,78 @@
...
+                        <Row>
+                            <Text style={applicationStyles.bold}><Trans>LEARN MORE</Trans></Text>
+                        </Row>
+                    </Grid>
+                </Tab>
+                <Tab heading={<TabHeading><Text><Trans>FIND SERVICES</Trans></Text></TabHeading>}>
+                </Tab>
+            </Tabs>
+        </Content>
+    </Container>
```
rasmus-storjohann-PG: I would prefer smaller components just like smaller functions.

src/components/tasks/task_detail.tsx
```diff
@@ -0,0 +1,78 @@
...
+import { Button, Content, Text, Icon, Container, Tab, Tabs, TabHeading } from 'native-base';
+import * as selector from '../../selectors/tasks';
+import { Col, Row, Grid } from 'react-native-easy-grid';
+import { applicationStyles } from '../../application/styles';
+import { taskDetailStyles } from './styles';
+import { Trans } from '@lingui/react';
+
+export interface Props {
+    readonly task: selector.Task;
+}
```
rasmus-storjohann-PG: Nice!

src/components/tasks/task_detail.tsx
```diff
@@ -0,0 +1,78 @@
...
+                    </Col>
+                    <Col size={15}>
+                        <Button dark transparent>
+                            <Icon name='more' />
+                        </Button>
+                    </Col>
+                </Row>
+            </Grid>
+            <Tabs style={taskDetailStyles.tabs}>
+                <Tab heading={<TabHeading><Text><Trans>INFORMATION</Trans></Text></TabHeading>}>
```
rasmus-storjohann-PG: Would you be able to demonstrate what localization of the `INFORMATION` string would look like? There is quite a bit of UX text in this story, it'd be really nice to have that translated (just with google translate for now).

src/application/router.ts
```diff
@@ -13,6 +13,10 @@ const getRouteFromPage = (page: Page): string => {
             return '/plan';
         case Page.ExploreAll:
             return '/explore';
+        // This is how our route "should" be.
+        // Currently in discussions on why this doesn't work.
+        //case Page.TaskDetail:
+        //    return '/task-detail/:taskId';
```
rasmus-storjohann-PG: I'm going to see if I can make this work.

src/components/tasks/styles.ts
```diff
@@ -27,3 +27,27 @@ export const taskDetail = StyleSheet.create({
         flexWrap: 'wrap',
     },
 });
+
+export const taskDetailStyles = StyleSheet.create({
```
rasmus-storjohann-PG: Inconsistent naming with `task` above, I like `taskStyles` and `taskDetailStyles`

src/components/tasks/task.tsx
```diff
@@ -1,10 +1,11 @@
 import React from 'react';
 import { View } from 'react-native';
 import { Text, Button, Icon } from 'native-base';
-import { taskDetail as styles } from './styles';
+import { task as styles } from './styles';
```
rasmus-storjohann-PG: When you're renaming a component in the import not to disambiguate but to make a better name, it's a good hint as to what the thing you're importing ought to be called.

src/stores/__tests__/page_switcher.test.ts
```diff
@@ -34,8 +35,25 @@ describe('setting the explore page', () => {
...
+        const theAction = pageSwitcher.setTaskDetailPage(aString());
+        expect(theAction.type).toBe(constants.SET_TASK_DETAIL_PAGE);
+    });
+});
+
 describe('the reducer', () => {
-    it('should default to build a store with Page.Questionnaire', () => {
+    it('should default to build a store with a currentPage property', () => {
+        const theStore = pageSwitcher.reducer();
+        expect(theStore).toHaveProperty('currentPage');
```
rasmus-storjohann-PG: And what is the value of that property, seems very relevant here. Same with the next test. When that is tested elsewhere, this test is redundant, I think

src/components/main/main_page_switcher.tsx
bensinjin: Yeah for sure. I definitely did think about having the `ConnectedTaskDetail`. I'll be sure to compile my thoughts around this.

src/components/tasks/task_list.tsx
bensinjin: This isn't actually the call site. The call site is on the tasks's button and looks like: `onPress={(): SetTaskDetailPageAction => props.goToTaskDetail(props.id)}` (task.tsx). Or am I misunderstanding what you're getting at?

src/components/tasks/actions.ts
bensinjin: Currently never. Fixed.

src/components/tasks/task.tsx
bensinjin: Yeah that sounds good. I'm wondering if that more falls under the realm of styling the task lists. Is it reasonable I do this work in my PR for #82?

src/components/tasks/task_detail.tsx
bensinjin: Understandable although I think much of size comes from using Row and Col. I personally like Row and Col as it helps keeps our styles files smaller and adds clarity to flexbox. Either way I'm all for breaking these components up further. Did you want me to try another approach?

tnightingale: I wondered if there was a way to define the layout in its own component? That way we could have a selection of reusable layouts? Might be a bit of a dream though.

src/components/tasks/task_detail.tsx
bensinjin: Fixed.

src/application/router.ts
bensinjin: That would be great :)

src/components/tasks/styles.ts
bensinjin: Yep I agree. I didn't think renaming the task styles was pertinent to this issue (probably nit picky) and have done so in my task list pr (#117).

src/components/tasks/task.tsx
rasmus-storjohann-PG: Sure, #82 if you prefer, I don't have a preference.

src/components/tasks/task.tsx
bensinjin: I hear ya. This was actually a remnant of the confusion caused around what exactly the taskDetail was (task list item vs. full display) and was renamed to indicate the style was for the list item. In #117 convention is clearer around names for style sheets.

src/stores/__tests__/page_switcher.test.ts
bensinjin: This test describes the store structure. The test two down tests the value of that property. Whether or not these properties exist and what their we'd expect their values to be are separate tests no? Would you like them together?

src/components/tasks/task_detail.tsx
bensinjin: That would be nice.. I think it will become more clear on how we actually accomplish something like that when we have the need for the same layout in two places. I'd envision a collection of layout HOCs will potentially begin to emerge.

tnightingale:
Yeah it's a bit tricky because components can't impose structure on their
children unless you pass them as props (which isn't necessarily a bad
thing).

src/components/tasks/task_list.tsx
bensinjin: Oh .. I believe I was. Wouldn't it have to be `goToTaskDetail={(item) => actions.goToTaskDetail(item.id)}` and then at the call site be `onPress={goToTaskDetail(props)}` or something?

rasmus-storjohann-PG: If you made the change I suggest here, then the call site would be nicer, in stead of `onPress={(): SetTaskDetailPageAction => props.goToTaskDetail(props.id)}` you would get `onPress={props.goToTaskDetail}`

src/components/tasks/task.tsx
bensinjin: Fixed in #82

bensinjin:
@rasmus-storjohann-PG I think I've addressed everything here - how close do you think this is to merging?

src/components/tasks/task_list.tsx
rasmus-storjohann-PG: No I didn't try it ... doesn't look as simple as I thought ... keep it in mind for next time, I think there must be a way and it could potentially be used very widely.

src/components/tasks/actions.ts
rasmus-storjohann-PG: Oh, and you didn't fix all the other ones while you were at it?

src/components/tasks/task_detail.tsx
rasmus-storjohann-PG: I find that when it comes to pulling out shared code, it's best to have two or three full implementations in place first. With just one example, it can be hard to know exactly what will actually be shared.

src/components/tasks/styles.ts
rasmus-storjohann-PG: Getting a good design in place for styling is important, and it's an area where I'm on the learning curve, so I'm following your lead. Naming is the most important part of design, and it is often something that takes a bit of time to get right, so it's normal to tweak and experiment with naming as we go.

src/components/tasks/actions.ts
bensinjin: The others are intentionally optional and are used to determine what actions are available in different contexts. For example we only render the "add" button when we have an "addToSavedList" property.  This could be done in other ways.

PR #117, opened 2018-06-12
Feature/issue 82 style task lists

bensinjin:
This issue was created off the #44 branch and will require PR #116 to be merged prior. 

These issues have been done together as they share the same component and it's easier to work on them / review them together.

Note (why Travis fails): NativeBase was updated in this branch to allow us access to some newer features. Unfortunately the newest release (which we want) has a typo in it and requires this PR: https://github.com/GeekyAnts/NativeBase/pull/1990 to be merged to allow us to run tests. 

I've opted to fork native-base for the time being see comment below.


rasmus-storjohann-PG:
Tag all your commits with just one story, and close out the other story with a comment indicating that the implementation done in story #xyz

rasmus-storjohann-PG:
I'll review this PR once the previous one is merged, otherwise it gets complicated

bensinjin:
Another note: NativeBase has an Accordion component which would probably be nicer to work with than the library I brought in for the collapsible functionality. The major difference between an accordion and collapsible is an accordion does not allow multiple sections open at once which may not work for us ...

src/components/my_plan/my_plan.tsx
```diff
@@ -1,15 +1,120 @@
...
+    toggleSuggestedTasksCollapsed(): void {
+        this.setState({ suggestedTasksCollapsed: !this.state.suggestedTasksCollapsed});
+    }
+
+    getIconNameForCollapsible(collapsed: boolean): string {
+        if (I18nManager.isRTL && collapsed) {
+            return 'arrow-dropleft';
+        }
+        return collapsed ? 'arrow-dropright' : 'arrow-dropdown';
+    }
```
rasmus-storjohann-PG: Your functions should tell a story about how the app is supposed to work. As written, this function tells me that if isRTL and collapsed, return dropleft, otherwise...   I don't think that's how you'd explain this to someone. I think this tells a better story:

```ts
   getIconNameForCollapsible(collapsed: boolean): string {
       if (collapsed) {
           return I18nManager.isRTL ? 'arrow-dropleft' : 'arrow-dropright';
       }
       return 'arrow-dropdown';
   }
```

src/components/my_plan/my_plan.tsx
```diff
@@ -1,15 +1,120 @@
...
+                            </Row>
+                        </Grid>
+                    </ListItem>
+                    <Collapsible collapsed={this.state.suggestedTasksCollapsed}>
+                        <ConnectedSuggestedTasks />
+                    </Collapsible>
+                </Content>
+            </Container>
+        );
+    }
```
rasmus-storjohann-PG: What would it look like if this was broken up into smaller components? I'm really going to want to find good ways to avoid having more than 20 lines of JSX in one chunk. Especially since this is a class based component, it should be fairly straightforward.

bensinjin:
I've opened https://github.com/pg-irc/pathways-frontend/issues/119 around this PR.

src/components/tasks/task.tsx
```diff
@@ -14,34 +15,44 @@ export type Actions = TaskActions;
...
+                    </Col>
+                    <Col size={20}>
+                        <Row style={taskStyles.rightColumn}>
+                            <Icon style={taskStyles.icon} name='star-circle' type='MaterialCommunityIcons'/>
+                            <Icon style={taskStyles.icon} name={I18nManager.isRTL ? 'arrow-back' : 'arrow-forward'} />
+                        </Row>
+                    </Col>
+                </Row>
+            </Grid>
+        </ListItem>
```
rasmus-storjohann-PG: Same here, break this up, especially when you have a `?:` in there, the `?:` should fit on one line

src/components/tasks/task_list.tsx
```diff
@@ -17,24 +17,16 @@ export interface Props {
...
+    </List>
+);
 
 export const renderSavedTask = (item: selector.Task, actions: TaskActions): JSX.Element => {
     return (
         <Task
+            key={item.id}
             id={item.id}
             taskUserSettingsId={item.taskUserSettingsId}
             title={item.title}
```
rasmus-storjohann-PG: Is't there a way to use `...` to do all of this in one line of code to copy all properties from item and actions into the task props?

package.json
```diff
@@ -79,7 +79,7 @@
     "expo": "^27.0.0",
     "history": "^4.7.2",
     "intl": "^1.2.5",
-    "native-base": "^2.5.1",
+    "native-base": "affinitybridge/NativeBase",
```
rasmus-storjohann-PG: Once you no longer need this, I hope you'll squish this commit. The goal is that the project builds and tests pass at every commit.

src/components/my_plan/my_plan.tsx
bensinjin: Yeah I like that better too. Fixed.

src/components/my_plan/my_plan.tsx
bensinjin: Fixed.

src/components/tasks/task.tsx
bensinjin: Fixed.

src/components/tasks/task_list.tsx
bensinjin: I didn't find a terrific solution to this.

bensinjin:
I think I've addressed outstanding issues. Please take a look.

src/components/collapser/collapser.tsx
```diff
@@ -0,0 +1,42 @@
...
+        super(props);
+        this.state = { collapsed: this.props.initiallyCollapsed };
+    }
+
+    render(): JSX.Element {
+        return (
+            <View>
+                <ListItem
+                    style={this.props.style ? this.props.style : undefined}
+                    noBorder button noIndent onPress={(): void => this.toggleCollapsed()}>
```
rasmus-storjohann-PG: `onPress={toggleCollapsed}`

src/components/my_plan/my_plan.tsx
```diff
@@ -1,15 +1,83 @@
...
+import { Container, Content, Text, View, Col, Row, Grid, Button, Icon } from 'native-base';
 import { ConnectedSavedTasks } from '../tasks/connected_saved_tasks';
 import { ConnectedSuggestedTasks } from '../tasks/connected_suggested_tasks';
+import { Trans } from '@lingui/react';
+import { applicationStyles } from '../../application/styles';
+import { myPlanStyles } from './styles';
+import { Collapser } from '../collapser/collapser';
 
-export const Component: React.StatelessComponent = (): JSX.Element => (
+export const MyPlan: React.StatelessComponent = (): JSX.Element => (
```
rasmus-storjohann-PG: I predict we'll find that `MyPlanComponent` and `MyPlanConnectedComponent` are the kinds of names we'll prefer

src/components/my_plan/my_plan.tsx
```diff
@@ -1,15 +1,83 @@
...
+                expandedHeader={getHeaderForSuggestedTasks(false)}
+                content={<ConnectedSuggestedTasks />}
+                initiallyCollapsed={true}
+                style={myPlanStyles.suggestedTasks}
+            />
         </Content>
     </Container>
 );
+
+const getHeaderForSavedTasks = (collapsed: boolean): JSX.Element => (
```
rasmus-storjohann-PG: bools as function arguments are a code smell, prefer enums.

src/components/tasks/task.tsx
```diff
@@ -1,47 +1,54 @@
...
-                        <Icon name='arrow-forward' />
-                    </Button>
-                </View>
-            </View>
-        </View>
+        <ListItem style={style} button noIndent onPress={goToTaskDetail}>
+            <Grid>
+                <Row>
+                    <Col size={10}>
+                        {!props.addToSavedList ? getDragButton() : getSaveButton(addToSavedList) }
```
rasmus-storjohann-PG: The `!` reduces readability, try to always test against the positive, both in `:?` and `if`, so
```ts
{props.addToSavedList ? getSaveButton(addToSavedList) : getDragButton() }
```

src/components/tasks/task_list.tsx
rasmus-storjohann-PG: Me neither, WIP

src/components/collapser/collapser.tsx
bensinjin: Because this is a class based component that won't actually work because 'this' will not have the correct context.

tnightingale: I don't think it is a big deal but the React people recommend binding in the constructor these days because it avoids the (apparently non-trivial) overhead of creating new function every time render() is called.

bensinjin:
Ah ok good to know. I can change my ways :)

Issue #106, opened 2018-06-08
Language switcher in header


rasmus-storjohann-PG:
Button in app header (visible on all screens):

![screenshot from 2018-06-08 14-03-24](https://user-images.githubusercontent.com/32307781/41180915-b90017ec-6b24-11e8-947a-54c447a446de.png)

List of languages, with scroll behaviour, note how Arabic slides behind the "English" label:

![screenshot from 2018-06-08 14-04-41](https://user-images.githubusercontent.com/32307781/41180961-e7bea238-6b24-11e8-89e3-2132f51e7ca1.png)

Note: The list of languages are defined in the fixtures, it will be a much smaller list, check with Michael.

rasmus-storjohann-PG:
Closed as a dupe of #118.

tnightingale:
I think this should remain open and #100 should be closed as the duplicate.

rasmus-storjohann-PG:
Reopened, thanks github for making PRs look just like issues.

tnightingale:
Putting this (pull req #118) back into "In Progress" to address styling issues.

Issue #104, opened 2018-06-08
MVP master story


rasmus-storjohann-PG:
In scope:
- [x] Tasks and questionnaire comes from local data in app 
- [ ] See #72 for data model (still subject to refinement)
- [ ] See [here for information on LTR](https://material.io/design/usability/bidirectionality.html#mirroring-layout)
- [x] #105 Welcome screen with language selection
- [x] #106 Language switcher in header 
- [x] #107 Home page (“How can we help you today?”) 
- [x] #108 Explore-all
- [x] #109 Explore Section - Related tasks and articles
- [x] #110 Article with related tasks and articles
- [x] #111 Related tasks and articles on Task detail page
- [x] #112 The simplest possible recommendation engine, maybe just the ones recommended to everybody
- [x] #80 Questionnaire, with styling (checkboxes and radio buttons) 
- [x] #82 My Plan task list, with styling
- [x] #86 Implement task state machine
- [x] #113 Notifications - Simple example of notification when completing a task 
- [x] #84 Service search (api call to server), services without styling
- [x] #44 Task Detail, some of this is out of scope for #44 but on scope for MVP, need another story: simple implementation includes: Mark as done / Done / Add to My Plan buttons, Information tab, Find Services tab, 
- [x] #114 Some NCG data from local data in app 

Out of scope:
* Welcome / onboarding video 
* “Need help?” button and Help page contents 
* “Search for topic or service near you” - needs more discussion and wireframing 
* My Plan (onboarding) tutorial overlays 
* Notifications - notifications architecture - needs discussion / research
* Questionnaire - advanced user elements like conditional logic, sliders, accordions, etc
* Recommendation algorithm
* Support for “Updated” tasks since original set was imported 
* User login / profile “saving” 
* Edit Profile link to overlay menu 
* Task Detail - share buttons, deep linking support, additional context help overlays (why was this recommended) 
* Saving user state to server
* Fancy language switcher
* User sign in/account creation
* User account management screens 
* Tasks and questionnaire comes from server
* Some NCG data imported into the server
* Glossary 
* Advanced swipe gestures 
* Server - all content management user interface

Issue #100, opened 2018-06-06
Build and style app header


rasmus-storjohann-PG:
AC: Has language button and "Need help" button as per wireframes

rasmus-storjohann-PG:
Closed as a dupe of #106

PR #118, opened 2018-06-12
Add locale switcher to header

tnightingale:
Adds a button to the header which also displays the current locale. When clicked triggers sidebar drawer to slide out with locale switcher inside.

Some possible outstanding tasks (need to decide whether these are within scope of MVP and, if so,  are within scope of this issue):

- [ ] Should `localeSwitcherOpen` reducer/store state & `openLocaleSwitcher()` actions be bundled in with rest of locale actions/state or be put alongside other general application UI state?
- [ ] Add "Need help" button to header (seems unrelated & not sure if it made it into MVP?)
- [ ] Keep drawer open through application restart (do we want this? I suspect it's more important to persist user's current page)
- [ ] Short-hand and/or localized names for locales (eg: 'EN' instead of 'English')
- [x] Make sure header button icons correctly mirror in RTL
- [ ] ... what else?


bensinjin:
This looks good. I think we probably want to group our UI related state together but that's just my opinion. I think we don't want to persist the drawer being open (page is important I agree) although for it to feel consistent would be nice. I would almost lean towards the LTR language switching to work like the RTL switch meaning the drawer closes when you select a language. I'm not sure how hard that is or if it goes too far against how people expect the drawer to work (I personally think closing is a sorta unintuitive).  Finally I'm not sure if this is meant to mirror the designs but it could def. use some love in Android.

tnightingale:
> I think we probably want to group our UI related state together but that's just my initial thought.

Yes, I think so too but I get stuck trying to define what constitutes UI state and what is not. I can see a scenario where we create a `userInterfaceStore` for all things UI that it just becomes a 
dumping ground for a bunch of unrelated state. :-\ In the meantime, I think our current state grouping/organization is sufficient.

> I would almost lean towards the LTR language switching to work like the RTL switch meaning the drawer closes when you select a language.

My concern was that it might be disorientating to have the drawer disappear. But I actually think that this is the best approach.

> Finally I'm not sure if this is meant to mirror the designs but it could def. use some love on Android.

I assume you're referring to the fact that it sits under the OS's status/notification bar. This is currently a problem application-wide (it's even worse with an iPhone X's 'notch'). It needs to be addressed with a top-level style.

rasmus-storjohann-PG:
Did you notice there is a story #106? All work to get to the MVP should be covered in stories under #104, unless there's something we've overlooked. Please check #104 carefully before writing new stories.

tnightingale:
No I did not, this is in response to #100 (which I did not create, it was assigned to me during the last sprint planning meeting).

bensinjin:
> I assume you're referring to the fact that it sits under the OS's status/notification bar. This is currently a problem application-wide (it's even worse with an iPhone X's 'notch'). It needs to be addressed with a top-level style.

I just tried it on Android the and the buttons are square and blue and it generally does not match designs provided.

rasmus-storjohann-PG:
"Need help" button is out of scope as per https://docs.google.com/document/d/1130BnFQS0T4SrE7lA12bZjjTZuY8Fr7xzObunFxNkJE/edit#

rasmus-storjohann-PG:
@tnightingale totally my bad, I thought I was looking at a story, not a PR.

tnightingale:
Good points @bensinjin I will update styling to match mocks (#118).

rasmus-storjohann-PG:
`localeSwitcherOpen` is not part of the locale store, it's either part of some localeSwitcher store or just stored in the component. I'm interested in exploring when component state should just live in the component. To be clear, I'm not a redux-store-for-everything-no-class-based-components zealot.

rasmus-storjohann-PG:
I would not expect the drawer to remain open after app restart ... I don't think. Do whatever is easier for now.

rasmus-storjohann-PG:
Localized names (English) for locales in the menu, definitely.

tnightingale:
> I would not expect the drawer to remain open after app restart ... I don't think. Do whatever is easier for now.

But should it remain open after a same-layout language is selected?

It seems like the consensus is trending towards "no" which means that we can move the `localeSwitcherOpen` state into component state.

rasmus-storjohann-PG:
Yeah, I think the drawer closes when you select a language, so we don't need an additional click to confirm the selection..

tnightingale:
> Localized names (English) for locales in the menu, definitely.

So, to clarify: locale names should be in English, eg:
- English
- Arabic
- Chinese

Not:
- English
- عربى
- 中文

rasmus-storjohann-PG:
For the store organization, really pay attention to the structure that makes the components easy to test, that will typically be the best design. I anticipate that this will likely move us more towards a `fooComponentStore` and `barComponentStore` as opposed to `allUIStore.foo` and `allUIStore.bar`.

rasmus-storjohann-PG:
Locale names should be in the current locale, so "English" or "Anglais" depending,

rasmus-storjohann-PG:
Hang on a second, I spoke too soon. We should not go against the mocks so easily, I'm guessing Cherry put thought into this. And this is what the mockis have looked like all along. I think we should proceed according to the mocks.

rasmus-storjohann-PG:
Yup, confirmed with Cherry, follow the mocks, my bad.

src/stores/__tests__/locale.test.ts
```diff
@@ -106,6 +114,16 @@ describe('the reducer', () => {
         expect(theNewStore).toBe(theOriginalStore);
     });
 
+    it('when called with OPEN_LOCALE_SWITCHER should return store with localeSwitcherDrawerOpen === action.payload.open', () => {
+        const theStore = locale.reducer();
+        const theAction = {
+            type: constants.OPEN_LOCALE_SWITCHER as typeof constants.OPEN_LOCALE_SWITCHER,
+            payload: { open: true },
```
rasmus-storjohann-PG: `payload: { open: aBoolean() },` then you test both paths in one test

rasmus-storjohann-PG:
Actually, I don't think you're done. Will not merge yet. Bug me if anything is unclear around this.

tnightingale:
Ok, I have:
 - moved the drawer logic into component state
 - tidied up styling using proper React Native and Native Base components (FYI Native Base's `<List />` is deprecated in favour of RN's `<FlatList />` and `<SectionList />` components)
 - updated the locale labels to be their localized (per-language) versions

I have implemented this with RN's `<SectionList />` as it's supposed to give us "sticky-headers" but it doesn't seem to work. I'm interested to hear if anyone knows what's up with that?
There's currently a bunch of dummy locales registered (A, B, C, D, ..., etc) to demonstrate the list scrolling (and supposed to show "sticky-headers").

Please review :)

rasmus-storjohann-PG:
Change the commit messages to reference #106 rather than #100, sorry for the confusion on that.

rasmus-storjohann-PG:
I like the UX of the drawer closing immediately on selecting a language, it's consistent whether the app switches directionality or not, which is good.

rasmus-storjohann-PG:
There's an arrow on each entry in the select language menu which is not there in the mocks, is that a side effect of the SecitonList?

src/application/store.ts
```diff
@@ -16,8 +16,26 @@ import zhMessages from '../../locale/zh/messages';
...
+    { code: 'i', label: 'I', catalog: zhMessages, isRTL: false },
+    { code: 'j', label: 'J', catalog: enMessages, isRTL: false },
+    { code: 'k', label: 'K', catalog: arMessages, isRTL: true },
+    { code: 'l', label: 'L', catalog: zhMessages, isRTL: false },
+    { code: 'm', label: 'M', catalog: enMessages, isRTL: false },
+    { code: 'n', label: 'N', catalog: arMessages, isRTL: true },
+    { code: 'o', label: 'O', catalog: zhMessages, isRTL: false },
+    { code: 'p', label: 'P', catalog: enMessages, isRTL: false },
+    { code: 'q', label: 'Q', catalog: arMessages, isRTL: true },
+    { code: 'r', label: 'R', catalog: zhMessages, isRTL: false },
```
rasmus-storjohann-PG: I'm not sure why these extra entries are needed. UX needs to look OK with a small set of supported languages, the MVP will have at most three.

src/components/language_switcher/connected_language_switcher.ts
```diff
@@ -4,12 +4,14 @@ import { LanguageSwitcher, Props, Actions } from './language_switcher';
...
 const mapStateToProps = (store: Store): Props => {
-    return {
-        currentLocale: selectLocale(store),
-        availableLocales: selectAvailableLocales(store),
-    };
+    const locales = selectAvailableLocales(store);
+    const locale = selectLocale(store);
+    const currentLocale = locales.find((aLocale: LocaleInfo) => locale.code === aLocale.code);
+    const availableLocales = locales.filter((aLocale: LocaleInfo) => locale.code !== aLocale.code);
+    return { currentLocale, availableLocales };
```
rasmus-storjohann-PG: This should be in a selector, there should be no logic in `mapStateToProps()`.

src/components/language_switcher/language_switcher.tsx
```diff
@@ -16,21 +17,56 @@ export type LanguageSwitcherProps = Props & Actions;
...
-            </View>
-            <View style={{alignItems: 'flex-start'}}>
-                <Text style={{textAlign: 'left'}}>
-                    <Trans>This sentence exists to demonstrate the translation functionality that exists in this application.</Trans>
-                </Text>
-            </View>
-        </View>
+        <ListItem key={item.code} onPress={(): void => item.setLocale(item.code)}>
+            <Left><Text>{item.label}</Text></Left>
+            <Right><Icon name={I18nManager.isRTL ? 'arrow-back' : 'arrow-forward'} /></Right>
```
rasmus-storjohann-PG: OK, I understand the arrows now ...

src/components/main/header.tsx
```diff
@@ -1,24 +1,37 @@
...
-                <Icon name='arrow-back' />
-            </Button>
-        </Left>
-        <Body>
-            <Title>Pathways</Title>
-        </Body>
-    </Header>;
+export interface UiActions {
+    readonly onLanguageSelect: () => void;
+}
```
rasmus-storjohann-PG: Seems to me that these two actions interfaces can be combined into one.

src/components/main/main.tsx
```diff
@@ -1,16 +1,51 @@
...
 import React from 'react';
-import { Container } from 'native-base';
+import { Container, Drawer } from 'native-base';
 import * as header from './header';
 import * as mainPageSwitcher from './main_page_switcher';
 import * as footer from './footer';
+import { ConnectedLanguageSwitcher } from '../language_switcher/connected_language_switcher';
+
+interface DrawerState {
+    readonly isDrawerOpen: boolean;
```
rasmus-storjohann-PG: So with this state in the component, you don't need the state in the store, right?

src/stores/__tests__/helpers/locale_helpers.ts
```diff
@@ -122,6 +123,7 @@ export class LocaleStoreBuilder {
             fallback: this.fallback,
             loading: this.loading,
             errorMessage: this.errorMessage,
+            localeSwitcherOpen: this.localeSwitcherOpen,
```
rasmus-storjohann-PG: No longer needed, I think

src/application/store.ts
tnightingale: To demonstrate the acceptance criteria in #106:

> List of languages, with scroll behaviour, note how Arabic slides behind the "English" label:

I mentioned this in my comment above:

> There's currently a bunch of dummy locales registered (A, B, C, D, ..., etc) to demonstrate the list scrolling (and supposed to show "sticky-headers").

The sticky-headers functionality, if it worked as advertised, would provide the "slides behind" effect described in AC.

> Also, we need to make sure that the UX works with a smaller number of supported locale.

It works fine when there's only 3 languages, the list just does not scroll. I will remove the extra languages.

src/components/language_switcher/connected_language_switcher.ts
tnightingale: I'm not sure I agree with this policy; it is arbitrary and therefore prevents situations where it might be appropriate to have component-specific logic (though, to be clear, it should definitely be tested).

src/components/main/main.tsx
tnightingale: Not sure what you're getting at, the state has been removed from the store.

src/stores/__tests__/helpers/locale_helpers.ts
tnightingale: Good catch, will remove.

src/components/main/header.tsx
tnightingale: These need to be separate as `Header#Actions` is exported and combined with the other Main component `Actions` & `Props` (from `footer` and `mainPageSwitcher` in `src/main/main.tsx`). This union type is used to create the connected component which has no knowledge of the internal state of its subcomponents.

Perhaps we need to improve our naming conventions to help clarify this?

src/components/main/main.tsx
tnightingale: Sorry, I see now that I hadn't completely removed traces from the store. Should be gone now.

src/components/language_switcher/language_switcher.tsx
```diff
@@ -16,21 +17,71 @@ export type LanguageSwitcherProps = Props & Actions;
...
+                    keyExtractor={(item: LocaleInfo): string => item.code}
+                    sections={[currentLocaleSection]}
+                    renderItem={renderLocaleListItem}
+                    renderSectionHeader={renderSectionHeader} />
+            </Content>
+        </Container>
+    );
+};
+
+type SectionHeaderInfo = {
```
tnightingale: I'm not particularly happy with these types but unfortunately React-Native's types ([@types/react-native](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-native/index.d.ts#L3855-L3991) don't fully match the API (and the API itself is quite convoluted). Happy to take suggestions if anyone has a better way of organizing this.

src/application/store.ts
rasmus-storjohann-PG: OK, now I understand, thanks

src/components/main/header.tsx
rasmus-storjohann-PG: OK, makes sense. Yeah, we'll need to look at naming in this context... post MPV

tnightingale:
Rebased and removed the extra locales

src/components/language_switcher/language_switcher.tsx
```diff
@@ -51,22 +48,40 @@ type SectionItemInfo = {
...
+            <Text>{item.label}</Text>
         </ListItem>
     );
+}
+
+function createLocaleListItemBuilder(onPress: (code: string) => void): LocaleListItemBuilder {
+    return (locale: LocaleInfo): LocaleListItem => {
+        return { ...locale, onPress: (): void => onPress(locale.code) };
+    };
+}
```
rasmus-storjohann-PG: Nice, I've been looking for a way to do this

tnightingale: Yeah you can actually avoid the extra overhead of creating per-item functions and just access it from the `section` object that's also passed to the itemRender method but sadly RN's types don't include that argument (but it's definitely there).

PR #120, opened 2018-06-13
Feature/issue 108 explore all component

rasmus-storjohann-PG:


src/components/explore/paginate.ts
```diff
@@ -0,0 +1,8 @@
+import * as R from 'ramda';
+
+export const paginate = <T>(pageSize: number, items: ReadonlyArray<T>): ReadonlyArray<ReadonlyArray<T>> => {
+    if (items.length === 0) {
+        return [];
+    }
+    return [R.take(pageSize, items), ...paginate(pageSize, R.drop(pageSize, items))];
```
rasmus-storjohann-PG: Write with `R.split` instead.

src/components/explore/explore_all.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+export interface ExploreAllActions {
+}
+
+type AllExploreProps = ExploreAllProps & ExploreAllActions;
+
+export const ExploreAllComponent: React.StatelessComponent<AllExploreProps> =
+    (props: AllExploreProps): JSX.Element => {
+        const sectionsGroupedIntoThrees = paginate(3, props.sections);
+        return <Content>
```
bensinjin:
```
return (
    <Content>
    ...
);
```
Looks a little nicer.

src/components/__tests__/paginate.test.ts
```diff
@@ -0,0 +1,40 @@
...
+    it('returns multiple pages when number of elements are greater than page size', () => {
+        input = [1, 2, 3, 4, 5];
+        expected = [[1, 2], [3, 4], [5]];
+        expect(paginate(pageSize, input)).toEqual(expected);
+    });
+    it('works with any type', () => {
+        const inputStrings: ReadonlyArray<string> = ['a'];
+        const expectedStrings: ReadonlyArray<ReadonlyArray<string>> = [['a']];
+        expect(paginate(pageSize, inputStrings)).toEqual(expectedStrings);
+    });
```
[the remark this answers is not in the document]
bensinjin: Agreed, nice work.

src/components/explore/explore_all.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+    // tslint:disable-next-line:typedef
+    return R.reduce((acc, name) => R.concat(acc, name), '', names);
+};
+
+interface ButtonRowProps {
+    readonly sections: ReadonlyArray<ExploreSection>;
+}
+
+const RowOfSectionButtons = (props: ButtonRowProps): JSX.Element => (
```
bensinjin: Wonder if this is reusable. Maybe not the most agile to break it out now.

src/components/explore/explore_all.tsx
```diff
@@ -0,0 +1,64 @@
...
+export interface ExploreAllActions {
+}
+
+type AllExploreProps = ExploreAllProps & ExploreAllActions;
+
+export const ExploreAllComponent: React.StatelessComponent<AllExploreProps> =
+    (props: AllExploreProps): JSX.Element => {
+        const sectionsGroupedIntoThrees = paginate(3, props.sections);
+        return <Content>
+            <Text>Learn about</Text>
```
bensinjin: After a rebase potentially do something like:
```
import { applicationStyles } from '../../application/styles'
...
<Text style={applicationStyle.pageTitle} /><Trans>Learn about</Trans></Text>
```

src/stores/explore.ts
```diff
@@ -0,0 +1,11 @@
+import { buildExploreFixture, Store } from '../fixtures/explore';
+export { Store, ExploreSection, ExploreSectionMap } from '../fixtures/explore';
+
+const buildDefaultStore = (): Store => (
+    buildExploreFixture()
+);
+
+// tslint:disable-next-line:no-any
+export const reducer = (store: Store = buildDefaultStore(), _?: any): Store => {
```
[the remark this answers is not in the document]
rasmus-storjohann-PG: This is an excellent point. It may get tricky because telling TypeScript that the reducer could be called with any action might give it conniptions. Punting this out past the MPV. Tracking in #123.

src/selectors/explore.ts
```diff
@@ -0,0 +1,25 @@
...
+    return denormalizeSections(locale, sections);
+};
+
+const denormalizeSections = (locale: Locale, store: model.ExploreSectionMap): ReadonlyArray<ExploreSection> => {
+    const buildSection = (id: string): ExploreSection => {
+        const name = selectLocalizedText(locale, store[id].name);
+        const icon = store[id].icon;
+        return { name, icon };
+    };
+    return R.map(buildSection, R.keys(store));
```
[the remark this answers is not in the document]
rasmus-storjohann-PG: I'd be OK either way with this, but I'm using R as much as possible right now just to get to know the library and explore how the different pieces of R work together (spoiler: Not always that great).

src/components/explore/explore_all.tsx
```diff
@@ -0,0 +1,64 @@
...
+    };
+
+const computeUniqueKeyForSections = (sections: ReadonlyArray<ExploreSection>): string => {
+    const names = R.pluck('name', sections);
+
+    // why doesn't this work??
+    // return R.reduce(R.concat, '', names);
+
+    // tslint:disable-next-line:typedef
+    return R.reduce((acc, name) => R.concat(acc, name), '', names);
```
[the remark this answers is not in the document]
rasmus-storjohann-PG: It should be unnecessary, but the compiler doesn't agree. I was able to make it work by hacking in a `toString()` call. Without the toString, I get the compiler error 
```
error TS2322: Type 'string | {}[]' is not assignable to type 'string'.
  Type '{}[]' is not assignable to type 'string'.
```
so it seems that reduce returns a string or an array of objects ... tricky, will be on this learning curve for a bit, but I think it's worth it.

src/components/__tests__/paginate.test.ts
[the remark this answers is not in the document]
rasmus-storjohann-PG: Thanks. Too bad they're gone again, since ramda has spitEvery

src/components/explore/explore_all.tsx
rasmus-storjohann-PG: Fixed ... I think .... don't quite know what I'm doing here

src/components/explore/explore_all.tsx
rasmus-storjohann-PG: Seeing something once, imagining that it could be reused, and then pulling it out into some central location tend to not give good results. We often get big "misc" folders of code that isn't actually shared and therefore is in the wrong place. Seeing something two, three or four times, recognizing that it can be reused, pulling it out into a location that makes sense with where it is used, and with an API that actually is flexible enough to be used in several context is the better way to go.

src/components/explore/explore_all.tsx
rasmus-storjohann-PG: Sweet, thanks

src/components/explore/styles.ts
```diff
@@ -0,0 +1,23 @@
+import { StyleSheet } from 'react-native';
+
+// use application wide settings once PR #117 is merged
+export const appColors = {
+    lightGrey: 'lightgrey',
+    lighterGrey: '#F5F5F5',
+    darkGrey: '#666666',
+};
```
rasmus-storjohann-PG: Hang on, I'll fix this...

src/components/explore/explore_all.tsx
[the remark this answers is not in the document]
tnightingale: That looks like the TypeScript type declaration authors were being lazy. Conventionally, `reduce` can return whatever type is passed in as the 3rd argument (actually, whatever type is returned from the 1st argument but you're really pushing it if you make them different).

I've encountered similar 'holes' in the official TypeScript definitions for React-Native.

tnightingale:
👍🏻

src/components/explore/explore_all.tsx
```diff
@@ -0,0 +1,55 @@
...
+export interface ExploreAllActions {
+}
+
+type AllExploreProps = ExploreAllProps & ExploreAllActions;
+
+export const ExploreAllComponent: React.StatelessComponent<AllExploreProps> =
+    (props: AllExploreProps): JSX.Element => {
+        const sectionsGroupedIntoThrees = R.splitEvery(3, props.sections);
+        return <Content padder>
+            <Text style={applicationStyles.pageTitle}>Learn about:</Text>
```
bensinjin: @rasmus-storjohann-PG last bit - I think the 'Learn about' needs to be wrapped in `<Trans>`

PR #125, opened 2018-06-21
Feature/issue 109 taxonomies

rasmus-storjohann-PG:


rasmus-storjohann-PG:
@tnightingale @bensinjin @trisapeace here's the second part of the work to get the explore section pages. Ben, please have a look at how the explore page hooks up to the explore section page, using a connected component for both, I think it might be little cleaner than the tasklist/task way of doing things.

src/fixtures/taxonomies.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export interface TaxonomyMap {
+    readonly [taxonomyId: string]: Taxonomy;
+}
+
+export interface Store {
+    readonly taxonomyMap: TaxonomyMap;
+}
+
+export interface TaxonomyTermReference {
```
bensinjin: What's the reasoning behind having the `TaxonomyTermReference`? Couldn't the taxonomyId, taxonomyTermId, icon. etc ... just exist on the TaxonomyTerm and references would be arrays of Ids?

bensinjin: Something like this is what I'd imagine:
```
TaxonomyTerm: {
    id: number,
    taxonomyId: number,
    name: string,
    icon: string,
}
Taxonomy: {
    id: string,
    name: string,
}
```

src/selectors/details/tasks.ts
```diff
@@ -0,0 +1,20 @@
+import { TaxonomyTermReference } from '../../stores/taxonomies';
+import { Task, TaskMap } from '../../stores/tasks';
+import { getExploreTaxonomyTerms } from '../taxonomies';
+import * as R from 'ramda';
+
+export const findTasksByExploreTaxonomyTerm =
```
bensinjin: Along the lines of what I previously mentioned this would probably be `findTasksByTaxonomy` and take a taxonomyId.

bensinjin:
This looks good but I did have a couple of questions (see comments). Regarding connected components for pages that works for me.  Finally I do sort of feel overuse of R is confusing in places but am I to understand you'd like to use it wherever possible?

src/selectors/details/tasks.ts
rasmus-storjohann-PG: Yeah it could make sense to make this more general, right now this function is hard-coded to only look at taxonomy terms where the taxonomy id is "Explore".

rasmus-storjohann-PG:
I'm looking to use R when possible for a few different reasons:
* Functional programming is good, and using a strong library of well designed functions is a very good basis to move towards a functional style.
* The various pieces of R are obviously made to work together, so by using R extensively, specifically preferring R functions over things like `Object#keys()` and `Array#map()` we're more likely to use the library correctly. 
* It also makes for better style (purely my opinion), compared to mixing up R and non-R functions and  it's clearly a lot less work that implementing our own versions of all the algorithms that R gives us.

src/selectors/details/tasks.ts
bensinjin: No I did actually mean `findTasksByTaxonomyId` meaning: give me all tasks that have terms that are members of a particular taxonomy. `findTasksByTaxonomyTermId()`would likely also exist at some point.

src/fixtures/taxonomies.ts
bensinjin: My point was the structure to connect tasks to a taxonomy term would be something as simple as a property on the task like: `taxonomyTerms: ['t1, 't2]` from the term you could then find the associate taxonomy.

bensinjin: This was just my initial thought. I don't actually have a grasp on what exactly the taxonomies look like.

rasmus-storjohann-PG: That would only work if taxonomy terms where globally unique, but they might not be. E.g. with the shopping for pants as an example, you might have a pair of pants with `colour:blue` and `liningColour:red`, and some other pair with `colour:red`, in which case having `taxonomyTerms :['red']` would be ambiguous. The data structure of taxonomy terms belonging to a taxonomy permits different taxonomy containing distinct terms with the same term id on them. Since the data structure can represent it, we have to make the references work with that too.

src/selectors/details/tasks.ts
rasmus-storjohann-PG: I don't know a context where "give me all tasks that have terms that are members of a particular taxonomy" would be useful, with many taxonomies this will be all the tasks.

src/fixtures/taxonomies.ts
rasmus-storjohann-PG: > it would allow for terms with the same names in different taxonomies

That doesn't make sense to me ... we'll need to sit down and talk through this, at some point, but it is not super urgent. Michael is working an initial pass of annotating the newcomers guide with taxonomy terms, so that will probably help.

bensinjin: I think I'm not being a clear as I'd hoped. We'll chat at some point I'm sure.

src/selectors/details/tasks.ts
bensinjin: Ok yeah initially I did misunderstand the intent of this function.

bensinjin:
I'm all for diving into functional programming :) With respect to the taxonomy work I'm ok with the structure you've outlined although my final thoughts are that a term id, with no connection to the actual name of the term, should be the means to disambiguate terms from one another and should be used to tag content.  I see the term name as a public facing representation of the term and the term taxonomy id as a grouping mechanism for terms.

src/fixtures/taxonomies.ts
bensinjin: See my final comment.

Issue #111, opened 2018-06-08
Surface related tasks and articles on task detail page


rasmus-storjohann-PG:
Make the related tasks and articles appear on the task detail page the same way as on the article page, see #110.

Issue #110, opened 2018-06-08
Article with related tasks and articles


rasmus-storjohann-PG:
![screenshot from 2018-06-08 14-22-12](https://user-images.githubusercontent.com/32307781/41181636-5d4645ea-6b27-11e8-9b83-8288c28b7308.png)

Use triangle/circle graphic for now.

Out of scope for this story: 
* Need help button, set language button
* Heart button
* Share button
* "This article is relevant for..."
* Recommendation star on tasks

robinpuga:
Link to your initial set of Articles from the Task details page, Learn More section...see https://projects.invisionapp.com/d/main/default/#/console/14398949/299441951/preview 

Also, when you are creating the look of the Learn More related article card please use the horizontal layout (with image beside text) as shown here: https://projects.invisionapp.com/d/main/default/#/console/14398949/300196080/preview#project_console and here https://projects.invisionapp.com/d/main/default/#/console/14398949/300196079/preview#project_console

PR #126, opened 2018-06-27
Feature/issue 110 articles

bensinjin:
Articles are not yet exposed any where through the UI so this will be visually easier to test once #111 is in which builds on #110. #111 Also adds some additional clean up around tasks work.


src/components/main/main_page_switcher.tsx
```diff
@@ -38,6 +39,9 @@ export const Component: React.StatelessComponent<Props & Actions> = (props: Prop
                 task={selectTaskById(props.locale, props.tasksStore, props.routeInProps.pageId)}
             />;
 
+        case store.Page.ArticleDetail:
+            return <ArticleDetailConnectedComponent />;
+
```
rasmus-storjohann-PG: I like how this file is shaping up, this should scale nicely. Aside: Why is `MyPlan` not a connected component?

src/components/tasks/connected_saved_tasks.ts
```diff
@@ -1,24 +1,18 @@
 import { Dispatch } from 'redux';
 import { connect } from 'react-redux';
-import { Component, Props, renderSavedTask } from './task_list';
-import { Actions } from './task';
+import { TaskListComponent, TaskListProps, TaskListActions } from './task_list';
```
rasmus-storjohann-PG: Nice

src/components/tasks/task_list.tsx
```diff
@@ -1,68 +1,27 @@
...
-    </List>
+export const TaskListComponent: React.StatelessComponent<AllTaskListProps> = (props: AllTaskListProps): JSX.Element => (
+    <View>
+        {R.map((task: Task) =>
+            <TaskListItemComponent
+                key={task.id}
+                listItemStyle={props.listItemStyle}
+                goToTaskDetail = {props.goToTaskDetail}
+                addToSavedList = {props.addToSavedList}
+                displayTaskInteractions={props.shouldDisplayTaskInteractions ? props.shouldDisplayTaskInteractions(task) : true}
```
rasmus-storjohann-PG: So I don't understand why you make these functions optional. I've seen this pattern in a few places now. You can make sure that the function is always set, so there is no reason to need to check if it's there.

src/components/tasks/task_list_item.tsx
```diff
@@ -3,34 +3,35 @@ import { ListItem, Text, Button, Icon, Grid, Col, Row } from 'native-base';
...
 
-export type Props = selector.Task;
+export interface TaskListItemStyleProps {
+    readonly listItemStyle?: object;
+}
+export interface TaskListItemProps extends TaskListItemStyleProps, Task {
+    readonly displayTaskInteractions: boolean;
+}
+export type TaskListItemActions = TaskActions;
+export type AllTaskListItemProps = TaskListItemProps & TaskListItemActions;
```
rasmus-storjohann-PG: Don't export

src/components/tasks/connected_suggested_tasks.ts
```diff
@@ -1,22 +1,21 @@
...
+import { TaskListItemStyleProps } from './task_list_item';
 import { Store } from '../../application/store';
 import { selectAllSuggestedTasks } from '../../selectors/tasks';
-import * as stores from '../../stores/tasks';
+import { Id, AddToSavedListAction, addToSavedList } from '../../stores/tasks';
 import { selectLocale } from '../../selectors/locale';
 import { SetTaskDetailPageAction, setTaskDetailPage } from '../../stores/page_switcher';
 
-const mapStateToProps = (store: Store): Props => ({
+const mapStateToProps = (store: Store, ownProps: TaskListItemStyleProps): TaskListProps => ({
```
rasmus-storjohann-PG: Any chance you could just use `TaskListItemProps` here instead of `TaskListItemStyleProps` to reduce the number of props types flying around?

src/fixtures/articles.ts
```diff
@@ -0,0 +1,63 @@
...
+
+export interface Article {
+    readonly id: Id;
+    readonly name: LocalizedText;
+    readonly content: LocalizedText;
+    readonly relatedTasks?: ReadonlyArray<TaskId>;
+    readonly relatedArticles?: ReadonlyArray<Id>;
+    readonly isRecommendedToAllUsers: boolean;
+    readonly isStarred: boolean;
+}
```
rasmus-storjohann-PG: `Articles` are very similar to `Tasks`, so let's use consistent naming. `name` -> `title`, `content` -> `description`, `isStarred` -> `starred`, and give it taxonomy terms just like tasks.

src/selectors/articles.ts
```diff
@@ -0,0 +1,60 @@
...
+    const articles = store.applicationState.articlesInStore.articles;
+    const article = findArticleById(articles, articleId);
+    return denormalizeArticle(locale, article);
+};
+
+export const findArticleById = (articles: model.ArticleMap, articleId: model.Id): model.Article => {
+    const validate = validateOneResultWasFound(articleId);
+    const getArticle = R.compose(validate, R.values, R.pickBy(R.propEq('id', articleId)));
+    return getArticle(articles);
+};
```
rasmus-storjohann-PG: You've followed the pattern from the task implementation here, but it can be simplified. All this jazz is needed in the tasks case to do the lookup from task to taskUserSettings, where there is a logical one-to-one relationship implemented in a data structure that can support a one-to-many relationship, so the logic is needed over there to ensure that it really is a one-to-one. Here, a different data structure is used that is incapable of representing this kind of invalid state, so the validation is not needed. Here, all you need is 
```
export const findArticleById = (articles: model.ArticleMap, articleId: model.Id): model.Article => (
    articles[articleId]
);
```
or you can inline it 
```
export const selectArticleAsListItem = (store: app.Store, articleId: model.Id): Article => {
    const locale = selectLocale(store);
    const articles = store.applicationState.articlesInStore.articles;
    const article = articles[articleId];
    return denormalizeArticle(locale, article);
};
```

src/stores/page_switcher.ts
```diff
@@ -1,13 +1,16 @@
 import * as constants from '../application/constants';
 import * as helpers from './helpers/make_action';
-import { Id } from './tasks';
+import { Id as TaskId } from './tasks';
+import { Id as SectionId } from './explore';
+import { Id as ArticleId } from './articles';
```
rasmus-storjohann-PG: My dream with all these different Id types is to figure out how to make them incompatible, i.e. an Id of one type cannot be assigned to an Id of a different type. Then we can have the compiler find all bugs where we're using the wrong Id to look things up. Could be cool. If it doesn't work, we can just make a single Id type and use that everywhere.

src/selectors/articles.ts
```diff
@@ -0,0 +1,60 @@
...
+import { Task, selectTaskById } from './tasks';
+import { Id as TaskId } from '../stores/tasks';
+
+export interface Article {
+    readonly id: model.Id;
+    readonly name: string;
+    readonly content: string;
+    readonly isStarred: boolean;
+    readonly relatedTasks?: ReadonlyArray<Task>;
+    readonly relatedArticles?: ReadonlyArray<Article>;
```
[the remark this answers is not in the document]
bensinjin: Fixed.

src/selectors/articles.ts
```diff
@@ -0,0 +1,60 @@
...
+export const selectCurrentArticle = (store: app.Store): Article => {
+    const locale = selectLocale(store);
+    const articles = store.applicationState.articlesInStore.articles;
+    const articleId = selectRoute(store).pageId;
+    const article = findArticleById(articles, articleId);
+    // TODO task selectors should take the whole store.
+    const relatedTasks = article.relatedTasks ?
+        R.map((id: TaskId) => selectTaskById(locale, store.applicationState.tasksInStore, id), article.relatedTasks) : undefined;
+    const relatedArticles = article.relatedArticles ?
+        R.map((id: model.Id) => selectArticleAsListItem(store, id), article.relatedArticles) : undefined;
```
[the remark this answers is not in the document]
bensinjin: Yeah I'll be doing this work in #111 as I've already fixed up the tasks selector there.

src/components/articles/article_detail.tsx
```diff
@@ -0,0 +1,92 @@
...
+        <ArticleListComponent
+            articles={props.article.relatedArticles}
+            goToArticleDetail={props.goToArticleDetail}
+        />
+    );
+};
+
+const renderRelatedTasks = (props: AllArticleDetailProps): JSX.Element => {
+    const shouldDisplayTaskInteractions = (task: Task): boolean => (
+        R.find(R.propEq('id', task.id))(props.savedTasks) === undefined
```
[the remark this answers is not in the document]
bensinjin: Fixed

src/components/articles/article_detail_connected_component.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { selectCurrentArticle } from '../../selectors/articles';
+import { SetTaskDetailPageAction, setTaskDetailPage } from '../../stores/page_switcher';
+import { selectAllSavedTasks } from '../../selectors/tasks';
+import { AddToSavedListAction, addToSavedList } from '../../stores/tasks';
+import { selectLocale } from '../../selectors/locale';
+
+const mapStateToProps = (store: Store): ArticleDetailProps => ({
+    article: selectCurrentArticle(store),
+    savedTasks: selectAllSavedTasks(selectLocale(store), store.applicationState.tasksInStore),
+});
```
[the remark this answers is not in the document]
bensinjin: No I agree. This is fixed in #111.

src/components/articles/article_list.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import { View } from 'native-base';
+import { Article } from '../../selectors/articles';
+import { ArticleListItemComponent, ArticleListItemActions } from './article_list_item';
+import R from 'ramda';
+
+export interface ArticleListProps {
+    readonly articles: ReadonlyArray<Article>;
+}
```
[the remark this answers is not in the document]
bensinjin: Fixed

src/components/articles/article_list.tsx
```diff
@@ -0,0 +1,22 @@
...
+export type ArticleListActions = ArticleListItemActions;
+export type AllArticleListProps = ArticleListProps & ArticleListActions;
```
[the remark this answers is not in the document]
bensinjin: `ArticleListProps` is already used for the props (line 7) - this is the combination of `ArticleListProps & ArticleListActions`

src/components/main/main_page_switcher.tsx
bensinjin: It should be. This will be an easy refactor that I'll do in #111 as I've clean up the tasks selectors there and this is related.

src/components/tasks/task_list.tsx
bensinjin: Defining functions unusable in a particular context (example:  "addToSavedList" when the task is already in your list) is unnecessary and to me makes things less clear. In the case of properties I am cleaning up areas I was a bit too liberal with the `?`.  In the particular case of showing/hiding task interactions it will only be relevant to task lists that have tasks that may or may not be already in your plan.  In some cases default values may be useful but I was under the impression we wanted to avoid that approach as well.

bensinjin: I'm not particularly set on this either way. We could always set unused properties to undefined to appease TS if that's preferable.

src/components/tasks/task_list_item.tsx
bensinjin: Fixed

src/components/tasks/connected_suggested_tasks.ts
bensinjin: Unfortunately it's not that simple as the `TaskListItemProps` expects a normalized `Task` shape (as well as some other properties). We wouldn't want to have to pass something that resembled an entire task when simply wanting to tweak the style of the list item.

src/fixtures/articles.ts
bensinjin: sounds good. fixed

src/selectors/articles.ts
bensinjin: Right .. yeah that wasn't totally clear to me initially. Fixed.

src/components/tasks/task_list.tsx
rasmus-storjohann-PG: The Proper Solution(TM) to entities with different sets of valid operations is to use different types. In some ways of talking about programming, that's precisely what a type is: A set of valid operations on values. Won't ask for that in this PR, but definitely look for ways to get us there going forward.

bensinjin: Yeahhh deep down I felt that myself. You're right. I'll be making note.

PR #128, opened 2018-06-28
Feature/issue 111 task related content

bensinjin:
This builds on #110 and should be merged after. Navigate to a task to view related tasks/articles.


rasmus-storjohann-PG:
I'll review this once #110 is merged

src/components/related_content_list/related_content_list.tsx
```diff
@@ -0,0 +1,27 @@
...
+    (props: AllRelatedContentListProps): JSX.Element => {
+        const ListComponent = props.component;
+        return (
+            <Col>
+                <Row style={applicationStyles.hr} />
+                <Row>
+                    <Text style={applicationStyles.bold}><Trans>{props.title}</Trans></Text>
+                </Row>
+                <ListComponent {...props.componentProps} />
+            </Col>
```
tnightingale: As we've discussed in-person, I think the usage of `react-native-easy-grid` causes more trouble than value (in the form of verbosity and confusion about flex-box).

This use of `<Row>` and `<Col>` is problematic as it assumes that the component will be put inside an existing `<Grid>` but there's no indication to the consumer that this will be the case (though nothing will actually break because all `<Col>` essentially does is return a `<View>` with `flexDirection: 'column'`)

Instead I suggest you just use `<View style={{ flexDirection: 'column' }}>...</View>`. `<Row>` is unnecessary because you're not trying to render multiple elements horizontally on each line, just a single `<Text>`.

src/components/tasks/task_detail.tsx
```diff
@@ -23,56 +29,96 @@ export const TaskDetail: React.StatelessComponent<Props & Actions> = (props: Pro
...
+    };
+    return (
+        <RelatedContentList
+            title={'RELATED TASKS'}
+            component={TaskListComponent}
+            componentProps={componentProps}
+        />
+    );
+};
+
```
tnightingale: Oh boy this is going to conflict badly with my pull req... ;-)

src/components/related_content_list/related_content_list.tsx
```diff
@@ -16,12 +16,10 @@ export const RelatedContentList: React.StatelessComponent<AllRelatedContentListP
...
-                <Row style={applicationStyles.hr} />
-                <Row>
-                    <Text style={applicationStyles.bold}><Trans>{props.title}</Trans></Text>
-                </Row>
+            <View>
+                <View style={applicationStyles.hr} />
+                <Text style={applicationStyles.bold}><Trans>{props.title}</Trans></Text>
                 <ListComponent {...props.componentProps} />
-            </Col>
+            </View>
```
tnightingale: 👍

src/components/related_content_list/related_content_list.tsx
bensinjin: Yeah I hear ya. Fixed.  I wont be using the `Row, Col, Grid` as freely moving forward. I'm also coming around to the idea of inline styles or just styles in `StyleSheet.create()` in the component file for none "application" styles.  I think keeping the styling so separated is more annoying and less useful than it is in a web context.

src/components/tasks/task_detail.tsx
bensinjin: We should probably tag team it :)

tnightingale: At this point it's probably not worth doing until this is merged. Otherwise we'd have to resolve two conflicts.

src/components/tasks/task_detail.tsx
```diff
@@ -23,56 +29,96 @@ export const TaskDetail: React.StatelessComponent<Props & Actions> = (props: Pro
...
+        goToTaskDetail: props.goToTaskDetail,
+        addToSavedList: props.addToSavedList,
+        shouldDisplayTaskInteractions: shouldDisplayTaskInteractions,
+    };
+    return (
+        <RelatedContentList
+            title={'RELATED TASKS'}
+            component={TaskListComponent}
+            componentProps={componentProps}
+        />
```
rasmus-storjohann-PG: Two functions, both calling `RelatedContentList` with completely different props, which are defined as `object`, which is equivalent to `any`. There is really two different kinds of content lists components, if you create both you can make this type safe.

rasmus-storjohann-PG: Or, you can try to do this: `componentProps` in `renderRelatedArticles` is the props that come with `ArticleListComponent`, while `componentProps` in `renderRelatedTasks` is the props that come with `TaskListComponent`. So, can you change `RelatedContentListProps` so that it is a  template class (in C++ terminology)? The idea that `RelatedContentListProps` is really `RelatedContentListProps<T>` where `T` can be `ArticleListComponent` or `TaskListComponent`, and then the type of `componentProps` is `T.Props`, which resolves at build time to either `ArticleListComponent.Props` or `TaskListComponent.Props`

src/components/articles/article_detail.tsx
```diff
@@ -71,22 +73,17 @@ const renderRelatedTasks = (props: AllArticleDetailProps): JSX.Element => {
     const shouldDisplayTaskInteractions = (task: Task): boolean => (
         R.find(R.propEq('id', task.id), props.savedTasks) === undefined
```
rasmus-storjohann-PG: This should be equivalent:
```ts
R.none(R.propEq('id', task.id), props.savedTasks)
```

src/components/tasks/task_detail.tsx
```diff
@@ -1,19 +1,25 @@
...
+import { RelatedContentList } from '../related_content_list/related_content_list';
 
-export interface Props {
-    readonly task: selector.Task;
+export interface TaskDetailProps {
+    readonly task: Task;
+    readonly savedTasks: ReadonlyArray<Task>;
 }
+export type TaskDetailActions = TaskListItemActions & ArticleListItemActions;
+export type AllTaskDetailProps = TaskDetailProps & TaskDetailActions;
```
rasmus-storjohann-PG: Don't export

src/components/tasks/task_detail_connected_component.tsx
```diff
@@ -0,0 +1,23 @@
...
+    savedTasks: selectAllSavedTasks(store),
+});
+
+const mapDispatchToProps = (dispatch: Dispatch<Store>): TaskDetailActions => ({
+    goToArticleDetail: (articleId: Id): SetArticleDetailPageAction => dispatch(setArticleDetailPage(articleId)),
+    goToTaskDetail: (taskId: Id): SetTaskDetailPageAction => dispatch(setTaskDetailPage(taskId)),
+    addToSavedList: (taskId: Id): AddToSavedListAction => dispatch(addToSavedList(taskId)),
+});
+
+export const TaskDetailConnectedComponent = connect(mapStateToProps, mapDispatchToProps)(TaskDetailComponent);
```
rasmus-storjohann-PG: This is what all our connected component files should look like

src/fixtures/tasks.ts
```diff
@@ -11,6 +12,8 @@ export interface Task {
     readonly tags: ReadonlyArray<string>;
     readonly category: string;
     readonly importance: number;
+    readonly relatedTasks?: ReadonlyArray<Id>;
+    readonly relatedArticles?: ReadonlyArray<ArticleId>;
```
rasmus-storjohann-PG: Make non-optional, use empty arrays to represent no related content

src/selectors/__tests__/tasks.test.ts
```diff
@@ -58,48 +58,11 @@ describe('tasks selector', () => {
...
-            expect(task.id).toEqual(id);
-        });
-
-        test('throws when select task by id parameter is invalid', () => {
-            expect(() => selector.selectTaskById(locale, store, aString())).toThrow();
-        });
-
-        test('throws when select task user settings by id parameter is invalid', () => {
-            expect(() => selector.findTaskUserSettingsByTaskId(store.taskUserSettingsMap, aString())).toThrow();
-        });
```
rasmus-storjohann-PG: Why are these tests no longer needed?

src/selectors/articles.ts
```diff
@@ -36,14 +35,15 @@ export const selectCurrentArticle = (store: app.Store): Article => {
     const articles = store.applicationState.articlesInStore.articles;
     const articleId = selectRoute(store).pageId;
     const article = articles[articleId];
-    // TODO task selectors should take the whole store.
-    const relatedTasks = article.relatedTasks ?
-        R.map((id: TaskId) => selectTaskById(locale, store.applicationState.tasksInStore, id), article.relatedTasks) : undefined;
-    const relatedArticles = article.relatedArticles ?
-        R.map((id: model.Id) => selectArticleAsListItem(store, id), article.relatedArticles) : undefined;
+    const relatedTasks = article.relatedTasks ? selectRelatedTasks(store, article.relatedTasks) : undefined;
+    const relatedArticles = article.relatedArticles ? selectRelatedArticles(store, article.relatedArticles) : undefined;
```
rasmus-storjohann-PG: Remove optional check

src/selectors/articles.ts
```diff
@@ -36,14 +35,15 @@ export const selectCurrentArticle = (store: app.Store): Article => {
...
     return denormalizeArticle(locale, article, relatedArticles, relatedTasks);
 };
 
+export const selectRelatedArticles = (store: app.Store, articleIds: ReadonlyArray<model.Id>): ReadonlyArray<Article> => (
+    R.map((id: model.Id) => selectArticleAsListItem(store, id), articleIds)
+);
+
 export const selectArticleAsListItem = (store: app.Store, articleId: model.Id): Article => {
```
rasmus-storjohann-PG: Why `AsListItem`? I was expecting that the return type was `ArticleAsListItem`, which could make sense if you have different properties on articles depending on where they appear, but it returns just straight `Article` so I don't see why this function is specific to list items.

src/selectors/tasks.ts
```diff
@@ -14,6 +16,8 @@ export interface Task {
     readonly taxonomyTerms: ReadonlyArray<TaxonomyTermReference>;
     readonly category: string;
     readonly importance: number;
+    readonly relatedTasks?: ReadonlyArray<Task>;
+    readonly relatedArticles?: ReadonlyArray<Article>;
```
rasmus-storjohann-PG: Remove optional

src/selectors/tasks.ts
```diff
@@ -22,57 +26,70 @@ export interface Task {
...
+     relatedArticles: ReadonlyArray<Article>, relatedTasks: ReadonlyArray<Task>): Task => (
         {
             id: task.id,
             title: selectLocalizedText(locale, task.title),
             description: selectLocalizedText(locale, task.description),
             taxonomyTerms: task.taxonomyTerms,
+            starred: taskUserSettings.starred,
+            relatedArticles: relatedArticles,
+            relatedTasks: relatedTasks,
+            taskUserSettingsId: taskUserSettings.id,
```
rasmus-storjohann-PG: Selecting data from the store should isolate components as much as possible from having to worry about store details like ids. I tried removing this attribute and it still builds, so you may be able to remove it... I may be missing something.

src/selectors/tasks.ts
```diff
@@ -22,57 +26,70 @@ export interface Task {
...
-export const selectTaskById = (locale: Locale, store: stores.Store, taskId: stores.Id): Task => {
-    const { taskMap }: stores.Store = store;
-    if (taskMap[taskId] === undefined) {
-        throw new Error(`Could not find Task for task id: ${taskId}`);
-    }
-    const taskUserSettings = findTaskUserSettingsByTaskId(store.taskUserSettingsMap, taskId);
-    return denormalizeTask(locale, taskMap[taskId], taskUserSettings);
+export const selectCurrentTask = (store: app.Store): Task => {
+    const locale = selectLocale(store);
+    const { taskMap, taskUserSettingsMap }: model.Store = store.applicationState.tasksInStore;
```
rasmus-storjohann-PG: Can't say I'm a big fan of this syntax myself, it can't be read easily from left to right.

src/selectors/tasks.ts
```diff
@@ -22,57 +26,70 @@ export interface Task {
...
+    const taskId = selectRoute(store).pageId;
+    const taskUserSettings = findTaskUserSettingsByTaskId(taskUserSettingsMap, taskId);
+    const task = taskMap[taskId];
+    const relatedTasks = task.relatedTasks ? selectRelatedTasks(store, task.relatedTasks) : undefined;
+    const relatedArticles = task.relatedArticles ? selectRelatedArticles(store, task.relatedArticles) : undefined;
```
rasmus-storjohann-PG: Remove optional

src/selectors/tasks.ts
```diff
@@ -22,57 +26,70 @@ export interface Task {
...
+    return denormalizeTask(locale, task, taskUserSettings, relatedArticles, relatedTasks);
+};
+
+export const selectRelatedTasks = (store: app.Store, taskIds: ReadonlyArray<model.Id>): ReadonlyArray<Task> => (
+    R.map((id: model.Id) => selectTaskAsListItem(store, id), taskIds)
+);
+
+export const selectTaskAsListItem = (store: app.Store, taskId: model.Id): Task => {
```
rasmus-storjohann-PG: Same confusion regarding the name of this function

src/selectors/tasks.ts
```diff
@@ -22,57 +26,70 @@ export interface Task {
...
+    const locale = selectLocale(store);
+    const { taskMap, taskUserSettingsMap }: model.Store = store.applicationState.tasksInStore;
+    const task = taskMap[taskId];
+    const taskUserSettings = findTaskUserSettingsByTaskId(taskUserSettingsMap, taskId);
+    return denormalizeTask(locale, task, taskUserSettings, [], []);
```
rasmus-storjohann-PG: Oh, I see, as task list item, you don't include related tasks and articles... so this really is a different type from `Task`. It would be better to have separate `Task` and `TaskListItem` types, I think... probably...

src/selectors/tasks.ts
```diff
@@ -91,7 +108,7 @@ export const selectTasksForCurrentExploreSection = (store: app.Store): ReadonlyA
     return denormalizedTasks;
 };
 
-const buildDenormalizedTask = R.curry((locale: Locale, userTasks: stores.TaskUserSettingsMap, task: stores.Task): Task => {
+const buildDenormalizedTask = R.curry((locale: Locale, userTasks: model.TaskUserSettingsMap, task: model.Task): Task => {
     const userTask = findTaskUserSettingsByTaskId(userTasks, task.id);
-    return denormalizeTask(locale, task, userTask);
+    return denormalizeTask(locale, task, userTask, [], []);
```
rasmus-storjohann-PG: We can find a better way to handle the "with related content" and "without related content" cases. Just leaving the related content collections empty some of the time and not other times is going to cause us confusion and bugs down the road. Usually the type system is the solution to this kind of problem...

src/stores/__tests__/helpers/tasks_helpers.ts
```diff
@@ -11,6 +12,8 @@ export class TaskBuilder {
     title: string = aString();
     description: string = aString();
     taxonomyTerms: TaxonomyTermReference[] = [];
+    relatedTasks: ReadonlyArray<ArticleId> = [aString(), aString()];
+    relatedArticles: ReadonlyArray<store.Id> = [aString(), aString()];
```
rasmus-storjohann-PG: Do you want to try to make a test helper so that you can do `arrayOf(aString)` or something like that? Fun to make that typesafe.

src/components/articles/article_detail.tsx
```diff
@@ -30,15 +28,14 @@ export const ArticleDetailComponent: React.StatelessComponent<AllArticleDetailPr
                         </Row>
                         {renderActions(props)}
                         <Row>
-                            <Text>{article.description}</Text>
+                            <Text>{props.article.description}</Text>
                         </Row>
-                        {article.relatedArticles ? renderRelatedContent('LEARN MORE', renderRelatedArticles(props)) : undefined}
-                        {article.relatedTasks ? renderRelatedContent('RELATED TASKS', renderRelatedTasks(props)) : undefined}
+                        {props.article.relatedArticles ? renderRelatedArticles(props) : undefined}
+                        {props.article.relatedTasks ? renderRelatedTasks(props) : undefined}
```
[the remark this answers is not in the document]
bensinjin: Yep fixed

src/components/tasks/task_detail.tsx
bensinjin: I did probably overgeneralize this piece in an attempt to reuse some of the related pieces of the components. I think I've come up with something nicer now by breaking this up into two components "related_tasks" & "related_articles" that retain type safety.

src/components/articles/article_detail.tsx
bensinjin: Yeah, nice. Updated.

src/components/tasks/task_detail.tsx
bensinjin: Fixed

src/fixtures/tasks.ts
bensinjin: Fixed

src/selectors/__tests__/tasks.test.ts
bensinjin: That particular test is OK so I've added it back in. The other selectors that had tests here now take the entire store and I thought we decided on not testing those?

src/selectors/articles.ts
bensinjin: Fixed

src/selectors/tasks.ts
bensinjin: Fixed

src/selectors/tasks.ts
bensinjin: Fixed. It didn't appear to be necessary at all

src/selectors/tasks.ts
bensinjin: Fixed

src/selectors/articles.ts
bensinjin: I've cleaned this up.

src/selectors/tasks.ts
bensinjin: Ok. Cleaned up.

src/selectors/tasks.ts
bensinjin: New types created

src/selectors/tasks.ts
bensinjin: Fixed (see below)

src/selectors/tasks.ts
bensinjin: Agreed and fixed

Issue #84, opened 2018-05-30
Search for BC211 service from the client


rasmus-storjohann-PG:
AC: A button in the Explore pane issues a hard-coded REST request to the server for some services using the search API, and displays a simple list of the resulting records. The point of this story is to get the infrastructure for REST calls in place.

tnightingale:
Mock of how services should look (not AC):
![task detail-window-map 3](https://user-images.githubusercontent.com/169635/42001550-293b9ac4-7a19-11e8-9b93-5120e50b9302.jpg)

PR #129, opened 2018-06-28
Query task services

tnightingale:
Addresses #84

Quite a lot going on here, a few notes:

#### Build-time configuration with `.env` file
I have added [.env configuration file](https://github.com/zetachang/react-native-dotenv) for tracking build-time environment config such as API host. The `.env` format is a convention in the web community stemming from [The Twelve Factor App](https://12factor.net/).

Unfortunately this module is lacking a bit in functionality:
- [updating config requires a change in source code](https://github.com/zetachang/react-native-dotenv#changes-to-env-file-is-not-updated)
- [default config](https://github.com/zetachang/react-native-dotenv/issues/10)
- [multiple environments](https://github.com/zetachang/react-native-dotenv/pull/28)
- [local overrides](https://github.com/zetachang/react-native-dotenv/issues/32)

But the alternative (https://github.com/luggit/react-native-config) [does not work with Expo/CRNA](https://github.com/luggit/react-native-config/issues/249).

*There is an example .env file (`.env.example`) that needs to be copied to `.env` before you build.*

#### Service entity & data model
As our front-end data model doesn't current include a Service entity, I have stubbed one in based on the equivalent backend entity in the 211 data. However due to issues with the API (described below) I do not fetch associated organization and location info.

#### Service search query is arbitrary and english-only
As there is not yet any definition for how a task's related services should be searched, we currently just construct the query by taking the first 3 words from the task's title. This also means that the query is locale-specific and as we only have English content in the backend, we only get results when viewing in english.

#### Backend API and data model
I have some inline comments that mention some potential shortcomings with the backend API and/or data model as it currently exists.


tnightingale:
This shouldn't be merged until #128 is merged and the inevitable conflicts in `src/components/task/task_detail.tsx` are resolved.

src/sagas/__tests__/services.test.ts
```diff
@@ -0,0 +1,54 @@
+// tslint:disable:no-expression-statement no-let no-null-keyword
+import { call, CallEffect, PutEffect, put } from 'redux-saga/effects';
+import { API } from '../../api';
+import { updateTaskServices } from '../services';
+import { TaskBuilder } from '../../stores/__tests__/helpers/tasks_helpers';
+import { Task } from '../../stores/tasks';
+import { aString } from '../../application/__tests__/helpers/random_test_values';
+import { updateTaskServicesAsync, UpdateTaskServicesAsync, serviceFromServiceData } from '../../stores/services';
+import { LocaleBuilder } from '../../stores/__tests__/helpers/locale_helpers';
+import { APIResponse } from '../../api/api_client';
```
rasmus-storjohann-PG: Hmm, this is a lot of import statements for a unit test, indicates that things are coupled too tightly if so much stuff is needed. Won't try to fix this now, but food for thought going forward

src/sagas/services.ts
```diff
@@ -0,0 +1,24 @@
...
+import { API } from '../api';
+import { APIResponse } from '../api/api_client';
+
+export function* watchUpdateTaskServices(): IterableIterator<ForkEffect> {
+    yield takeLatest(constants.UPDATE_SERVICES_REQUEST, updateTaskServices);
+}
+
+export function* updateTaskServices(
+    action: UpdateTaskServicesAsync.Request,
+): IterableIterator<CallEffect | PutEffect<UpdateTaskServicesAsync.Result>> {
```
rasmus-storjohann-PG: If you want to fix that very long line, you can do 
```
type UpdateResult = IterableIterator<CallEffect | PutEffect<UpdateTaskServicesAsync.Result>>;
```
 ... if you want

src/stores/__tests__/services.test.ts
```diff
@@ -0,0 +1,100 @@
...
+
+describe('services reducer', () => {
+    const loadedTask = new TaskServicesBuilder();
+    const loadingTask = new TaskServicesBuilder().withLoading(true);
+    const initialService = new ServiceBuilder();
+    const theStore = buildNormalizedServices([loadedTask, loadingTask], [initialService]);
+
+    it('returns a unmodified store when the action is unknown', () => {
+        expect(reducer(theStore, { type: '' })).toBe(theStore);
+    });
```
rasmus-storjohann-PG: I like this pattern!

src/stores/services/index.ts
```diff
@@ -0,0 +1,92 @@
...
+import { UpdateTaskServicesAsync, updateTaskServicesAsync } from './update_task_services';
+import { Task as constants } from '../../application/constants';
+import { Action } from 'redux';
+
+export { Id, Service, Store };
+export { UpdateTaskServicesAsync, updateTaskServicesAsync };
+
+export function serviceFromServiceData(data: any): Service { // tslint:disable-line:no-any
+    // TODO: Perform appropriate data validation.
+    //       Alternatively bring in a tool to do this for us, eg: Serializr
```
rasmus-storjohann-PG: We were talking about whether we can make assumptions about falsy values in our data. With validation/data massaging here, we can make those assumptions, and I think that is worth doing. Not in this PR, though. 

This is a place where the use of `any` is totally correct, is shows that we make no assumptions about the data coming from the server, and only after validation can we confidently return a `Service`.

src/stores/services/index.ts
```diff
@@ -0,0 +1,92 @@
...
+    const id: string = data.id || undefined;
+    // TODO: Determine how to handle localization from the API.
+    // TODO: Hard-coding 'en' for the meantime.
```
rasmus-storjohann-PG: A bit of whiteboarding of the store structure for service data is in order...

src/stores/services/index.ts
```diff
@@ -0,0 +1,92 @@
...
+}
+
+export function reducer(store: Store = buildDefaultStore(), action: Action): Store {
+    switch (action.type) {
+        case constants.UPDATE_SERVICES_REQUEST:
+            return updateServicesRequest(store, action as UpdateTaskServicesAsync.Request);
+        case constants.UPDATE_SERVICES_SUCCESS:
+            return updateServicesSuccess(store, action as UpdateTaskServicesAsync.Success);
+        case constants.UPDATE_SERVICES_FAILURE:
+            return updateServicesFailure(store, action as UpdateTaskServicesAsync.Failure);
```
rasmus-storjohann-PG: `makeAction()` should make these casts unnecessary I thought?

src/stores/services/update_task_services.ts
```diff
@@ -0,0 +1,35 @@
+import { Service } from './types';
+import { Id as TaskId } from '../tasks';
+import { Task as constants} from '../../application/constants';
+import * as helpers from '../helpers/make_action';
+import { Action as ReduxAction } from 'redux';
+
+interface UpdateTaskServicesAction<P extends {} = {}> extends ReduxAction {
+    readonly payload: P;
+}
```
rasmus-storjohann-PG: Thanks for exploring the use of `ReduxAction` as a possible alternative to `makeAction()`. Interested to know the trade offs.

rasmus-storjohann-PG:
I'll want to see this merged before #128, @bensinjin lost the rebase lottery this time

src/components/main/main_page_switcher.tsx
```diff
@@ -34,7 +34,7 @@ export const Component: React.StatelessComponent<Props & Actions> = (props: Prop
             return <ExploreSectionConnectedComponent />;
 
         case store.Page.TaskDetail:
-            return <TaskDetail
+            return <ConnectedTaskDetail
                 task={selectTaskById(props.locale, props.tasksStore, props.routeInProps.pageId)}
             />;
```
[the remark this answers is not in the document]
bensinjin: Fixed in my branch :P

src/api/api_client.ts
```diff
@@ -0,0 +1,57 @@
...
+import qs from 'query-string';
+
+export interface APIResponse {
+    readonly hasError: boolean;
+    readonly message: string;
+    readonly response?: Response;
+    readonly results?: any; // tslint:disable-line:no-any
+}
+
+interface Query {
```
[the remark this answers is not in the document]
tnightingale: `Response` already exists, so I guess we'll go with `APIQuery`.

src/api/api_client.ts
```diff
@@ -0,0 +1,57 @@
...
+        const queryString = qs.stringify(query);
+        // TODO: Should use something a bit more robust to build urls: nodejs's
+        //       URL module is a good example.
+        const url = `${this.host}/${endpoint}?${queryString}`;
+        const response = await fetch(url);
+        // TODO: Handle fetching paged results.
+        //       The API doesn't specify total number of results* or number of
+        //       pages so would need to make sequential requests until no
+        //       results are returned.
+        //       *Maybe it should? JSON-API is a good spec to follow
```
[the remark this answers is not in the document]
tnightingale: Oh cool I didn't beyond the DRF/Swagger docs, thanks!

src/components/tasks/connected_task_detail.ts
```diff
@@ -0,0 +1,51 @@
...
+import { Props } from './task_detail';
+import { TaskServices, selectTaskServices, createRelatedServicesQueryFromTask } from '../../selectors/services';
+import { Task } from '../../selectors/tasks';
+import { connect } from 'react-redux';
+import { selectLocale } from '../../selectors/locale';
+
+interface StateProps {
+    readonly query: string;
+    readonly taskServices: TaskServices;
+}
```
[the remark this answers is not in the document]
tnightingale: They're only used during the `connect()`ing process, in this file.

src/sagas/__tests__/services.test.ts
tnightingale: This is a known issue, and definitely needs some more careful thought.

Testing Sagas is broken; you can't avoid testing implementation.

It seems to stem from the fact that they're built with generators. There's a TON of 'helper' modules for testing Sagas, I briefly looked into some but they are mostly band-aids that risk further complicating/obscuring the tests.

src/stores/services/index.ts
tnightingale: 👍

src/stores/services/index.ts
tnightingale: I don't think so.

The way I see it, we have to make this cast from a runtime string (`action.type`) to a real type somewhere. So we might as well acknowledge and do it where _we_ can be sure it is safe. And also make it explicitly clear what's happening.

src/components/tasks/connected_task_detail.ts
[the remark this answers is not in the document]
tnightingale: `OwnProps` is what's exposed 'outside' the connected component (see `src/components/main/main_page_switcher.tsx`). `Props` in `src/components/task/task_detail.tsx` is the aggregated props that the HOC provides to `<TaskDetail>`.

Naming could be improved but this will be moot when Ben's code gets merged and we no-longer expose `<TaskDetail>`'s props to parent components.

src/components/tasks/task_detail.tsx
```diff
@@ -1,78 +1,131 @@
...
-import { Button, Content, Text, Icon, Container, Tab, Tabs, TabHeading } from 'native-base';
+import { Button, Content, Text, Icon, Tab, Tabs, TabHeading, ListItem, View } from 'native-base';
 import * as selector from '../../selectors/tasks';
 import { Col, Row, Grid } from 'react-native-easy-grid';
 import { applicationStyles } from '../../application/styles';
 import { taskDetailStyles } from './styles';
 import { Trans } from '@lingui/react';
+import { Service, TaskServices} from '../../selectors/services';
+import { UpdateTaskServicesAsync } from '../../stores/services';
+import { Service as ServiceComponent } from '../services/service';
```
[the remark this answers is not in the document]
tnightingale: Done

src/components/tasks/connected_task_detail.ts
```diff
@@ -0,0 +1,51 @@
...
+import { Store } from '../../application/store';
+import { updateTaskServicesAsync, UpdateTaskServicesAsync } from '../../stores/services';
+import { Props } from './task_detail';
+import { TaskServices, selectTaskServices, createRelatedServicesQueryFromTask } from '../../selectors/services';
+import { Task } from '../../selectors/tasks';
+import { connect } from 'react-redux';
+import { selectLocale } from '../../selectors/locale';
+
+interface StateProps {
+    readonly query: string;
```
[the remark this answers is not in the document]
tnightingale: Done

src/sagas/services.ts
tnightingale: Done

src/selectors/services.ts
```diff
@@ -0,0 +1,44 @@
...
+    readonly description: string;
+}
+
+export interface TaskServices {
+    readonly loading: boolean;
+    readonly message: string;
+    readonly services: ReadonlyArray<Service>;
+}
+
+export function selectTaskServices(locale: Locale, taskId: TaskId, store: Store): TaskServices {
```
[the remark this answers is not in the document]
tnightingale: Done

src/selectors/services.ts
```diff
@@ -0,0 +1,44 @@
...
+    };
+}
+
+export function createRelatedServicesQueryFromTask(task: Task): string {
+    // TODO: Determine better method of getting / creating a search query.
+    //       The current approach is to split task title into a list and take
+    //       the first three words. The intention being to provide some
+    //       (hopefully) relevant keywords but not so many that we don't get any
+    //       results.
+    return take(3, task.title.split(' ')).join(',');
```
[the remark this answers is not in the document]
tnightingale: 🥇

src/stores/services/index.ts
tnightingale: Yes, it definitely needs more thought than I gave it :)

src/stores/services/update_task_services.ts
tnightingale: I think the main benefit is that we're able to use the generic `Action` type where it's needed (in reducers). Downside is that it's a bit more verbose but as I point out in another comment, the verbosity is also more explicit about where the weak points are in the 'dynamic' type system.

PR #136, opened 2018-07-04
Feature/issue 86 tasks states

bensinjin:


bensinjin:
I'd actually like to write a couple tests outlining assumptions around task states .. moving to in progress.

PR #137, opened 2018-07-04
Feature/issue 86 tasks states

bensinjin:


src/components/tasks/task_states.ts
```diff
@@ -0,0 +1,17 @@
+export enum TaskStates {
+    CompletedInPlan,
+    CompletedNotInPlan,
+    InProgress,
+    Available,
+}
+
+export const getTaskState = (inPlan: boolean, completed: boolean): TaskStates => {
```
rasmus-storjohann-PG: An API like this is prone to bugs because it's very easy to pass in the arguments in the wrong order. If you passed in the task, it would be safer. If you don't want this code to depend on the file containing the task definition, you could define an interface in this file with the two properties that this function needs, I think TypeScript would be smart enough to recognize (even without you saying so explicitly) that the Task type implements that interface, so it would just work...

src/components/tasks/saved_tasks_connected_component.ts
```diff
@@ -12,6 +12,7 @@ const mapStateToProps = (store: Store): TaskListProps => ({
 
 const mapDispatchToProps = (dispatch: Dispatch<Store>): TaskListActions => ({
     goToTaskDetail: (taskId: stores.Id): SetTaskDetailPageAction => dispatch(setTaskDetailPage(taskId)),
+    addToSavedList: undefined,
```
rasmus-storjohann-PG: I will keep an eye out for a better way to do this...

src/components/tasks/task_detail.tsx
```diff
@@ -132,6 +116,78 @@ function ServiceListEmpty(): JSX.Element {
...
+    const inPlan = R.any((id: TaskId) => id === task.id, props.savedTasks);
+    const taskState = getTaskState(inPlan, task.completed);
+
+    const doneButton = (
+        <Button iconLeft rounded light style={[{marginTop: 10}]}
+                onPress={(): void => {props.toggleCompleted(task.id); props.removeFromSavedList(task.id); }}>
+            <Icon name='checkbox' />
+            <Text><Trans>Mark Done</Trans></Text>
+        </Button>
+    );
```
rasmus-storjohann-PG: Another (more testable) way of doing this is to have a function that takes the `taskState` and returns an object with properties `onPress`, `iconName` and `text`, and then you can have a single definition of the button that pulls its props out of the object. This would give you DRYer code too.

src/components/tasks/task_detail_connected_component.tsx
```diff
@@ -26,31 +26,35 @@ function mapStateToProps(store: Store): StateProps {
...
-type DispatchProps = Actions & TaskListItemActions & ArticleListItemActions;
-
 function mapDispatchToProps(dispatch: Dispatch<Store>): DispatchProps {
     return {
-        goToArticleDetail: (articleId: Id): SetArticleDetailPageAction => dispatch(setArticleDetailPage(articleId)),
-        goToTaskDetail: (taskId: Id): SetTaskDetailPageAction => dispatch(setTaskDetailPage(taskId)),
-        addToSavedList: (taskId: Id): AddToSavedListAction => dispatch(addToSavedList(taskId)),
+        goToArticleDetail: (articleId: ArticleId): SetArticleDetailPageAction => dispatch(setArticleDetailPage(articleId)),
+        goToTaskDetail: (taskId: ArticleId): SetTaskDetailPageAction => dispatch(setTaskDetailPage(taskId)),
+        addToSavedList: (taskId: ArticleId): AddToSavedListAction => dispatch(addToSavedList(taskId)),
```
rasmus-storjohann-PG: Type should be `TaskId`. I want to figure out a way to make this a compile error...

src/components/tasks/task_states.ts
```diff
@@ -0,0 +1,17 @@
...
+    if (inPlan && completed) {
+        return TaskStates.CompletedInPlan;
+    } else if (!inPlan && completed) {
+        return TaskStates.CompletedNotInPlan;
+    } else if (inPlan && !completed) {
+        return TaskStates.InProgress;
+    }
+    return TaskStates.Available;
+};
```
rasmus-storjohann-PG: Many ways to skin this cat, this is maybe easier to follow?
```
if (inPlan) {
    return completed ? TaskStates.CompletedInPlan : TaskStates.InProgress;
}
return completed ? TaskStates.CompletedNotInPlan : TaskStates.Available;
```

src/components/tasks/task_states.ts
bensinjin: Good point. Currently, however, the task itself is not sufficient we'd also need the saved tasks list. Once we refactor our store for tasks this shouldn't be an issue. Regardless I've created an interface for this that should ensure it's more difficult to accidentally misuse the api.

PR #138, opened 2018-07-05
Feature/issue 112 recommendation engine

rasmus-storjohann-PG:


rasmus-storjohann-PG:
@bensinjin , @tnightingale (?) and @trisapeace for review. This implements the simplest possible recommendation mechanism. Note that my plan now is empty unless there are some selected answers in the questionnaire that cause the recommendation engine to recommend tasks. It will also be possible to tag tasks and articles with a taxonomy term that causes it to be recommended to everybody: `recommendation:recommendToAll`

src/selectors/tasks.ts
```diff
@@ -115,14 +115,20 @@ const buildDenormalizedTask = R.curry((locale: Locale, userTasks: model.TaskUser
     return denormalizeTask(locale, task, userTask, [], []);
 });
 
-export const selectRecommendedTasks = (taskMap: model.TaskMap): ReadonlyArray<model.Task> => {
-    const recommendToAll = (ref: TaxonomyTermReference): boolean => (
-        ref.taxonomyId === 'recommendation' && ref.taxonomyTermId === 'recommendToAll'
-    );
+export const selectRecommendedTasks =
```
bensinjin: I'm starting to feel like select* functions should always return content that's intended to be displayed. I think functions that query/filter "model" data should probably live in the associated store file.

Issue #142, opened 2018-07-12
Move this info to a README in the front end repo: Release publishing information


tnightingale:
After a bunch of digging, I now have a better understanding of how React Native and Expo work:

### React Native (RN)

The `react-native-cli` includes Facebook's "Metro Bundler"; a custom javascript bundler responsible for compiling & bundling your javascript for deployment in a RN application.

When you compile/build an platform binary for _development_ (eg: `react-native start` or `react-native run-[ios|android]`), a couple of things happen (this is not a exhaustive list):

- Metro Bundler is used to compile a javascript bundle (optionally, the bundler can watch the filesystem for changes and re-compile the javascript if there are any changes detected).
- All assets (images, fonts, etc) are collected
- A local server is started. It listens for connections from RN apps and serves the latest javascript bundle and any other assets required.

Once the native binary is deployed and run on a device (or emulator/simulator), it kicks off execution by making a request to the local server for the javascript bundle and assets and then runs your application.

_Production_ builds function in essentially the same way except that the Javascript bundle goes through an extra step where the debug features (source-maps, etc) are removed and the source code is minified/obfuscated. The resulting bundle is then embedded into the production binary and deployed to the device so that when run, it is loaded from the device and not via the development server.

### Expo

Expo follows a very similar process to RN (it wraps `react-native-cli`, Metro Bundler, the local dev server, etc) but adds some more infrastructure and steps to the system.

The Expo Client (aka ExpoKit) does several things (not an exhaustive list):

- Provides a bunch of native components and APIs and exposes them to the RN javascript environment (this is the Expo SDK).
- Replaces/Wraps the RN runtime with the Expo runtime

The Expo runtime functions very similarly to the RN runtime described above except that it is configured to [always*](https://docs.expo.io/versions/latest/guides/offline-support.html#bundle-your-assets-inside-your-standalone-binary) look for the compiled javascript bundle & other assets via a external URL. This URL either points to a local dev server (similar to RN's dev server) for _development_ or to a storage bucket on Amazon Web Services (AWS) or Google Cloud Platform (GCP) for _production_ releases.

_The main benefit of this is that it makes it possible to release (javascript) application updates without having to go through either platform's app store._

Publishing an Expo bundle is done with the `exp publish` command. This will create a production bundle of your javascript (minified/obfuscated, just like RN) and all of your assets and upload it to AWS or GCP. It will then return a url which can be used to load your application into _any device running the Expo application_.

[...]

PR #143, opened 2018-07-12
#142 Fix publishing issues

tnightingale:
There's still some icons that don't make it into the build (most obviously those used in the bottom tabbar on iOS). Someone needs to track these down and make sure they're included in the bundle. See #142 for more info.


tnightingale:
@bensinjin specifically this change: https://github.com/pg-irc/pathways-frontend/pull/143/commits/15f294c2542e19b082b649248a5c59f90441e0a8#diff-7765904c0fb44044594e919a5dd53488R25

The Javascript minification step during publish/production mode appears to perform an unsafe transformation to the `props` parameter's properties. Making `i18n` a local variable prevents this breakage from happening (the translation function is now referenced like this `i18n.t(...)` instead of `props.i18n.t(...)`). I've had similar problems with minification scripts like UglifyJS before; there's likely a configuration that prevents it but unfortunately I think Expo/CRNA/RN abstracts it away from us.

bensinjin:
@rasmus-storjohann-PG feel free to merge this when you're ready

Issue #145, opened 2018-07-17
MVP polish: Welcome page


rasmus-storjohann-PG:
ACs:

- [x] Needs some indication that you can select your language. Field label? See mocks for solution.
- [x] Can’t scroll on smaller screen: Make the widgets fill up the screen with no scrolling regardless of screen size.
- [x] “Get Started” button should go to Home screen. 
- [x] Hide the Pathways Header and Nav Footer on Landing screen
- [x] Some additional orientation text explaining what the app is/does under the name: 

```
Arrival Advisor

For immigrants and refugees new to Canada. Arrival Advisor is here to
 help you start your new life in Canada, every step of the way.
```

rasmus-storjohann-PG:
Changing the behaviour of the language picker on the welcome screen requires that the locale as picked by the fallback logic in `/home/rasmus/git/pathways-frontend/src/sagas/locale.ts#loadCurrentLocale()` (which should appear in the welcome screen as "Select your language") be distinguishable somehow from the locale set explicitly by the user. Not sure what would be a clean way to do this...

tnightingale:
If I recall correctly, it needs to support a `currentLocale === undefined` state (or something similar). 

Currently when the application loads, if no `currentLocale` is found in storage, the default/fallback is  automatically set and persisted. This currently prevents the `undefined` state from being possible.

PR #146, opened 2018-07-17
Issue #145 Hide header and footer on welcome screen

rasmus-storjohann-PG:


rasmus-storjohann-PG:
@bensinjin for review

PR #164, opened 2018-07-18
Feature/issue 145 polish welcome page

rasmus-storjohann-PG:


rasmus-storjohann-PG:
@bensinjin for review

rasmus-storjohann-PG:
I was not able to repro the scrolling problem on the welcome screen, can you give me some pointers on how to see the problem?

bensinjin:
Try setting the height of the star image to something larger. You'll notice the page does not scroll once the content goes off the screen. One way to fix this is to wrap the component in the `<content>` component so that it can scroll. Another way would be to use percentages for heights although this approach only works for heights (not widths) and doesn't fix the fact that font sizes are hardcoded. I'd probably do a bit of research around how to tackle these issues.

bensinjin:
I think in general you want to try and avoid hardcoding as much as possible. Ideally get familiar with flex box and use it as much as you can. This tool is kinda neat: https://yogalayout.com/playground (what RN uses for its flexbox implementation)

bensinjin:
I’ll get a screenshot of it on an iPhone 5 as that was device I saw the issue on. It’d be worth setting up an Android emulator with a smaller screen for these sort of issues.

bensinjin:
screen shot 2018-07-19 at 9 56 15 am

bensinjin:
screen shot 2018-07-19 at 10 00 38 am

rasmus-storjohann-PG:
Thanks for digging into this, I have it running on a tiny screen on an android simulator now.

rasmus-storjohann-PG:
The image on the welcome screen now crops in some weird ways depending on the screen size, but all the widgets are visible on the screen, so it is fully usable. Postponing the final styling of this page until we have our design.

Issue #147, opened 2018-07-17
MVP polish: Home screen


rasmus-storjohann-PG:
Updated mockups here: https://projects.invisionapp.com/share/H7JJSDYKW5B#/screens/308729231 

Should include the block for “You haven’t personalized your plan…” This should be in the My Plan section  above the recommended task list
https://projects.invisionapp.com/share/3VJJSE3NG28#/screens/299441955 
This directs you to the Questions 
Remove the “Not Now” (which would dismiss) for MVP 
Make the URL link in the copyright message open the web page

List of Recommended tasks should have some introduction in the preceding text, including a link to the questionnaire, matching MyPlan section on home page. 
Limited to the top 3 recommended tasks

Change `How can we help you today?`  to `Arrival Advisor helps you start your new life in Canada, every step of the way.`

For “You haven’t personalized your plan…”
```
My Plan
Plan everything you need to do as a newcomer to Canada. 
Want to know what next steps you need to take? Answer 
some questions to get tasks and tips recommended for you.
```

PR #167, opened 2018-07-20
Issue #147 Localize strings

rasmus-storjohann-PG:


rasmus-storjohann-PG:
@bensinjin for review

PR #171, opened 2018-07-26
Feature/issue 152 polish my pan

rasmus-storjohann-PG:


rasmus-storjohann-PG:
@bensinjin for review, this takes care of the My Plan page. There is one outstanding bug but I'd like to get this in and track that bug separately, I've got too many big PRs in play.

rasmus-storjohann-PG:
This PR requires a server side change, see https://github.com/pg-irc/pathways-backend/pull/211/files

src/components/tasks/task_list_item.tsx
```diff
@@ -45,8 +45,8 @@ export const TaskListItemComponent: React.StatelessComponent<AllTaskListItemProp
 const renderTaskInteractions = (props: AllTaskListItemProps): JSX.Element => {
     const addToSavedList = (): AddToSavedListAction => props.addToSavedList(props.id);
     return (
-        <Col size={10}>
-            {props.addToSavedList ? renderAddButton(addToSavedList) : renderDragButton()}
+        <Col size={15}>
+            {props.addToSavedList ? renderAddButton(addToSavedList) : renderUnchecedkBox()}
```
bensinjin: Fix typo, should be `renderUncheckedBox()`

src/components/tasks/task_list_item.tsx
```diff
@@ -57,8 +57,8 @@ const renderAddButton = (onPress: () => void): JSX.Element => (
     </Button>
 );
 
-const renderDragButton = (): JSX.Element => (
+const renderUnchecedkBox = (): JSX.Element => (
```
bensinjin: So I'm noticing I still see the "+" when the task is in my "TASKS I PLAN TO DO" section. Is this the bug you're referring to? It stems from the fact that the "+" is rendered whenever `props.addToSavedList` is defined.

src/components/tasks/task_list_item.tsx
rasmus-storjohann-PG: Fixed

src/components/tasks/task_list_item.tsx
rasmus-storjohann-PG: Yeah, looking at it briefly it seems to me that the fix won't be completely simple, so I'd rather get this PR merged and fix it later, probably under the same story.

rasmus-storjohann-PG: There are also some strings that aren't translated, we can do a full sweep of the app for that later.

rasmus-storjohann-PG:
Rebased. Remember to merge the server side piece at the same time

Issue #149, opened 2018-07-17
MVP polish: Questions


rasmus-storjohann-PG:
ACs:
* Remove the Questions link in Nav Footer.
* At top of questionnaire there is instructions that you are not required to answer any questions you don’t want to answer.
* Remove the Continue button 
* Change label of Skip to “Next”
* After final question is answered and Next button is pressed we should redirect to your My Plan page and see your Recommended Tasks.

PR #174, opened 2018-07-27
Issue #149 Clean up questions in questionnaire

bensinjin:


src/components/questionnaire/question.tsx
```diff
@@ -7,42 +7,43 @@ import { applicationStyles } from '../../application/styles';
...
 import { Collapser } from '../collapser/collapser';
+import { RouterProps, goToRouteWithoutParameter, Routes } from '../../application/routing';
+import { History } from 'history';
 
 export interface Props {
     readonly question: selector.Question;
 }
-
 export type Actions = QuestionnaireActions;
+type AllQuestionProps = Props & Actions & RouterProps;
```
rasmus-storjohann-PG: I don't know if you'll like this, but I am starting to rename these `AllQuestionProps` types (that are not exported) to just `Props`. The types `Props` and `Actions` here should be named to something more specific anyway, since they *are* exported.

src/components/questionnaire/question.tsx
```diff
@@ -7,42 +7,43 @@ import { applicationStyles } from '../../application/styles';
...
         <Collapser
-            collapsedHeader={getQuestionCollapsedHeader(question.number, question)}
-            expandedHeader={getQuestionExpandedHeader(question.number, question)}
-            content={getQuestionContent(question, selectAnswer)}
+            collapsedHeader={renderQuestionCollapsedHeader(props)}
+            expandedHeader={renderQuestionExpandedHeader(props)}
+            content={renderQuestionContent(props)}
             // TODO - This should likely be the next unaswered question in the questionnaire
-            initiallyCollapsed={question.number === 1 ? false : true }
+            initiallyCollapsed={props.question.number === 1 ? false : true }
```
rasmus-storjohann-PG: I think the proper solution to encoding the ordering of the questions is to have them stored in an array rather than a map in the store. Then either the selector can set number properties on them or the component can walk the array and figure out reliably what is the first and last question. This would likely be too big a change for the MVP, let's see what we can come up with that's simpler to implement for now.

src/fixtures/hard_coded/questionnaire.ts
```diff
@@ -19,6 +19,7 @@ export const buildQuestionnaireFixture = (): Store => {
                     'fr': 'Test du champ d\'explication',
                 },
                 acceptMultipleAnswers: false,
+                isFinalQuestion: false,
```
rasmus-storjohann-PG: This is too hacky. I'd either change the selector to have it compute the index of each question and set the property on the selector question, or (probably better) I'd put some logic in the component to figure out what is the last question. Ramda will have all the functions you need to do that simply.

src/components/questionnaire/question.tsx
bensinjin: Works for me. Fixed

src/components/questionnaire/question.tsx
bensinjin: This part is referring to what section to initially open when you navigate back to the questionnaire. It likely falls under #150.

bensinjin: The numbering stuff could probably be improved but we have something in place already that's working fine. What are you getting at exactly?

src/fixtures/hard_coded/questionnaire.ts
bensinjin: OK, fixed ... at least to my liking. I'd appreciate if feedback was kept less subjective. If something really is hacky you need to explain why otherwise nothing is gained.

src/fixtures/newcomers_guide/README.md
```diff
@@ -1 +0,0 @@
-This folder is here to contain generated content from newcomers guide. This would be generated at build time and never checked in.
```
rasmus-storjohann-PG: Don't delete this file

src/components/questionnaire/question.tsx
rasmus-storjohann-PG: No, what you have in this PR is what I had in mind. I was just thinking more long term for what we will want to do with the questionnaire.

PR #178, opened 2018-07-30
Feature/issue 150 accordion for questions

bensinjin:


src/components/questionnaire/question.tsx
```diff
@@ -1,77 +1,54 @@
...
 import { Trans } from '@lingui/react';
-import { Collapser } from '../collapser/collapser';
 import { RouterProps, goToRouteWithoutParameter, Routes } from '../../application/routing';
 import { History } from 'history';
+import { SetActiveQuestionAction } from '../../stores/questionnaire';
 
 export interface QuestionProps {
     readonly question: selector.Question;
     readonly isFinalQuestion: boolean;
+    readonly activeQuestion: number;
```
rasmus-storjohann-PG: We are using strings for question.id in the store, so it would be natural to use those strings to identify the question here too. You may have to set a property `nextQuestionId` on the questions in the selector, in which case `nextQuestionId === undefined` would identify the last question.

src/components/questionnaire/question.tsx
```diff
@@ -1,77 +1,54 @@
...
-        <Collapser
-            collapsedHeader={renderQuestionCollapsedHeader(props)}
-            expandedHeader={renderQuestionExpandedHeader(props)}
-            content={renderQuestionContent(props)}
-            // TODO - This should likely be the next unaswered question in the questionnaire
-            initiallyCollapsed={props.question.number === 1 ? false : true }
-        />
+        <Grid style={questionStyles.questionWrapper}>
+            <Row>
+                <Col>
```
rasmus-storjohann-PG: I thought you said you were going to move away from grid and use FlexBox directly...

src/components/questionnaire/questionnaire.tsx
```diff
@@ -1,32 +1,82 @@
...
+
 type Props = QuestionnaireProps & QuestionnaireActions & RouterProps;
 
 export const Component: React.StatelessComponent<Props> = (props: Props): JSX.Element => (
     <Content padder>
         <Text style={applicationStyles.pageTitle}><Trans>Personalize My Plan</Trans></Text>
         <Text style={questionnaireStyles.introText}>
-            <Trans>There is no requirement to answer any of the following questions but in doing so you help us recommend tasks right for you.</Trans>
+            <Trans>There is no requirement to answer any of the following questions
+                but in doing so you help us recommend tasks right for you.</Trans>
```
rasmus-storjohann-PG: Phrasing: "recommend tasks right for you."

src/components/questionnaire/questionnaire.tsx
```diff
@@ -1,32 +1,82 @@
...
     </Content>
 );
+
+interface AccordionSection {
+    readonly title: string;
+    readonly content: JSX.Element;
+}
+
+// tslint:disable-next-line:readonly-array
+type AccordionSectionArray = Array<AccordionSection>;
```
rasmus-storjohann-PG: Why can't you use `ReadonlyArray` here?

src/components/questionnaire/question.tsx
```diff
@@ -1,77 +1,54 @@
...
+                {question.explanation ? <Text style={questionStyles.explanationText}>{question.explanation}</Text> : undefined}
```
rasmus-storjohann-PG: Use `<EmptyComponent/>` rather than undefined here.

src/components/questionnaire/question.tsx
bensinjin: Switched this to a question Id in the store. Some helpers in `questionnaire.tsx` now manage finding the correct index for a given questions id (we're required to use an index for the `Accordion` component). This file no longer knows about the active question but is instead just passed an OnPress function for the next button.

src/components/questionnaire/question.tsx
bensinjin: Sure. Fixed. For the record I'm not apposed to Grid it's just in some scenarios the styles it sets can make doing what you need to do more difficult. I think though we might as well be consistent throughout and clean up Grid implementations as we see them.

src/components/questionnaire/questionnaire.tsx
bensinjin: The Accordion component expects a `push` function on the sections prop.

src/components/questionnaire/question.tsx
bensinjin: Fixed

src/components/questionnaire/questionnaire.tsx
bensinjin: Explain?

bensinjin:
@rasmus-storjohann-PG ready for re-review

src/components/questionnaire/questionnaire.tsx
rasmus-storjohann-PG: The sentence is not really English, "in doing so you help us recommend tasks right for you" should probably be "in doing so you help us recommend tasks and articles for you"

bensinjin: Pretty sure that's English ... There are a collection of tasks but not all of them are right for you ... we'll find the ones that are if you answer the questions.

bensinjin: Maybe it's sorta slang?

bensinjin: I'll change it, mine is probably confusing

bensinjin: Fixed

src/components/questionnaire/index.ts
```diff
@@ -1,19 +1,21 @@
...
+import { QuestionnaireActions } from './questionnaire';
 import { Store } from '../../stores';
-import { Id, SelectAnswerAction, selectAnswer } from '../../stores/questionnaire';
+import { Id, SelectAnswerAction, selectAnswer, SetActiveQuestionAction, setActiveQuestion } from '../../stores/questionnaire';
 import { selectQuestionnaire } from '../../selectors/questionnaire';
 
 const mapStateToProps = (store: Store): QuestionnaireProps => {
     return {
         questionnaire: selectQuestionnaire(store),
+        activeQuestion: store.questionnaireInStore.activeQuestion,
```
rasmus-storjohann-PG: should use a selector here

PR #188, opened 2018-08-03
Feature/issue 151 recommended task notifications

bensinjin:
So I've deviated from the original AC slightly to deliver what I think is a better UX. Instead of flashing notifications that were often the same (ie. "you have 7 recommended tasks") I've gone with a floating counter which Robin and I agreed on was an improvement. I think ideally if you wanted to do this with a notification you'd flash something like "2 new tasks were recommended" unfortunately the effort to build that with our current model is somewhat significant and this is arguably just as good if not better. Let me know your thoughts.


src/components/notifications/notification.tsx
```diff
@@ -1,44 +1,76 @@
...
 
-    private renderContentForNotification(notification: model.Notification): JSX.Element {
-        switch (notification.type) {
+    private renderContentForNotification(): JSX.Element {
+        switch (this.props.notification.type) {
             default:
-            case model.NotificationType.TaskAddedToPlan:
-                return <Trans>Task added to plan</Trans>;
+            case NotificationType.TaskAddedToPlan:
+                return this.renderContentForTaskAddedNotification();
```
rasmus-storjohann-PG: This switch statement is an example of writing dead code in anticipation of what you think will come in in the future. It will be trivial to write that swtich statement once we need it, for now it adds noise. It's better to write just the code you need. I suggest you remove the `NotificationType` type altogether.

src/components/notifications/notifications.tsx
```diff
@@ -1,62 +1,36 @@
...
 
-export const NotificationsComponent: React.StatelessComponent<AllNotificationsProps> = (props: AllNotificationsProps): JSX.Element => {
-    // tslint:disable-next-line:no-null-keyword
-    return R.keys(props.notifications).length > 0 ? renderNotifications(props) : null;
+export const NotificationsComponent: React.StatelessComponent<Props> = (props: Props): JSX.Element => {
+    return R.keys(props.notifications).length > 0 ? <Notifications {...props} /> : <EmptyComponent />;
 };
 
-const renderNotifications = (props: AllNotificationsProps): JSX.Element => {
+const Notifications: React.StatelessComponent<Props> = (props: Props): JSX.Element => {
```
rasmus-storjohann-PG: Having different symbols in our codebase whose name differ by just one letter (`Notification` and `Notifications`) makes the code harder to read. I suggest `NotificationListComponent` here.

src/components/notifications/notifications.tsx
```diff
@@ -1,62 +1,36 @@
...
+    readonly removeNotification: (notificationId: Id) => RemoveNotificationAction;
 }
-type AllNotificationsProps = NotificationsProps & NotificationsActions;
+type Props = NotificationsProps & NotificationsActions;
 
-export const NotificationsComponent: React.StatelessComponent<AllNotificationsProps> = (props: AllNotificationsProps): JSX.Element => {
-    // tslint:disable-next-line:no-null-keyword
-    return R.keys(props.notifications).length > 0 ? renderNotifications(props) : null;
+export const NotificationsComponent: React.StatelessComponent<Props> = (props: Props): JSX.Element => {
+    return R.keys(props.notifications).length > 0 ? <Notifications {...props} /> : <EmptyComponent />;
```
rasmus-storjohann-PG: `R.isEmpty()` works on objects

src/components/questionnaire/index.ts
```diff
@@ -5,13 +5,13 @@ import { QuestionnaireActions } from './questionnaire';
...
-    return {
-        questionnaire: selectQuestionnaire(store),
-        activeQuestion: selectActiveQuestion(store),
-    };
-};
+const mapStateToProps = (store: Store): QuestionnaireProps => ({
+    questionnaire: selectQuestionnaire(store),
+    activeQuestion: selectActiveQuestion(store),
+    recommendedTaskCount: selectRecommendedTasks(store).length,
+});
```
rasmus-storjohann-PG: Nice. Could you rename this file while you're here? It should be `questionnaire_connected_component.ts`

src/selectors/notifications.ts
```diff
@@ -14,8 +12,8 @@ export const denormalizeNotification = (notification: model.Notification): Notif
     }
 );
 
-export const selectNotifications = (store: Store): ReadonlyArray<Notification> => {
+export const selectNotifications = (store: Store): NotificationList => {
     const notifications = store.notificationsInStore.notifications;
     return R.map((notificationId: model.Id) =>
         denormalizeNotification(notifications[notificationId]), R.keys(notifications));
```
rasmus-storjohann-PG: So here there is not actually anything to denormalize, you're just returning the data straight from the model. So just do that, there is no need to write pass through function that doesn't do anything. If we need to do data manipulation in the future, we can write the function then.

src/selectors/tasks.ts
```diff
@@ -63,22 +63,27 @@ export const denormalizeTaskListItem = (locale: Locale, task: model.Task, isReco
     }
 );
 
-export const selectSavedTasks = (store: Store): ReadonlyArray<TaskListItem> => {
+export type TaskList = ReadonlyArray<Task>;
+export type TaskListItemList = ReadonlyArray<TaskListItem>;
+export type TaskModelList = ReadonlyArray<model.Task>;
+export type TaskIdList = ReadonlyArray<model.Id>;
```
rasmus-storjohann-PG: I see some naming problems here, `TaskListItemList` is definitely awkward. So you have two kinds of task objects, the small ones for list items, and the big one for details page. How about calling those `Task` and `TaskDetail`? Then we get `TaskList` and `TaskDetailList`. Or maybe better `TaskPartial` and `TaskFull`, with `TaskPartialList` and `TaskFullList`? 

I would also avoid `TaskModelList` and just use `model.TaskList` instead. Avoid having different definitions with different names for the same thing. I would also expect the model to contain a type for the task id list, so you should just reuse it here.

Also, I think we should probably prefer `store` to `model`, but that's a small thing, easy to fix later.

High level question: Are you sure that `FooList` is actually more readable than `ReadonlyArray<Foo>`?

src/components/notifications/notification.tsx
bensinjin: Yeah good point. Fixed. Ideally we'd probably not even have the notion of types at this point but we can maybe let that slide.

src/components/notifications/notifications.tsx
bensinjin: Fixed

src/selectors/tasks.ts
bensinjin: My thoughts/adjustments based on this feedback:
- For the time being I've reverted back to ReadonlyArray<type>. I'm not sure if typing these lists is an improvement or not ... On the one hand you probably have to think less when writing other components if you know you can just import the "list" type from the selector, on the other hand it may be difficult to build convention around this and a ReadonlyArray of some type is more clear
- For this file I've renamed the `store` argument to `appStore` and `model` to `store`. I think ideally we should have an `AppStore` type as apposed to our `Store` (clean up task?) type so we know when we're talking about a `store` it's always a sub store.

src/selectors/notifications.ts
bensinjin: So now in `notification_list.tsx` we have `import { Notification } from '../../stores/notifications';` .. is this what you'd expect? Should I be exporting the type from the selector?

bensinjin:
@rasmus-storjohann-PG please check my feedback/changes, cheers.

src/selectors/tasks.ts
bensinjin: Oh - regarding the `ListItem` name - I actually think `ListItem` is more informative than say `TaskPartial` as it does a better job of describing its application and doesn't require a whole lot of digging to understand the purpose of the type. `TaskDetail` and `TaskListItem` to me is pretty clear and I'd opt for using this convention elsewhere.

src/components/notifications/notification_list.tsx
```diff
@@ -15,7 +15,7 @@ export interface NotificationsActions {
 type Props = NotificationsProps & NotificationsActions;
 
 export const NotificationListComponent: React.StatelessComponent<Props> = (props: Props): JSX.Element => {
-    return R.empty(props.notifications) ? <EmptyComponent /> :
+    return R.isEmpty(props.notifications) ? <EmptyComponent /> :
```
rasmus-storjohann-PG: Yup, I did that too once

src/selectors/notifications.ts
rasmus-storjohann-PG: So a purist might say that there should be a distinct selector type even when the store type has the same fields of the same types. Others would see that as over-engineering. I think it's fine to reuse the store type in a selector when it looks the same.

Issue #185, opened 2018-08-01
MVP polish: Style bullet lists


rasmus-storjohann-PG:
AC: Bullet lists look good. Currently, the bullet character is a tiny dot.

PR #189, opened 2018-08-03
Issue #185 Fix styling for bullets

bensinjin:


rasmus-storjohann-PG:
The code changes are fine, but I had a look at the bullets as they appear now, and they still look funny, I think we need to see  if we can make this better before we merge.

Issue #123, opened 2018-06-19
Post-MVP cleanups


rasmus-storjohann-PG:
* Simplify task store data structure
* Think about reducers not ever actually being called with undefined actions, but they are called with `any` action.
* Maybe replace locale builder with locale generator functions similar to `aString()`, i.e. `aLocale()`, `aLocaleWithCode(code)`, etc.

# Selectors

Reorgnanize selectors, by creating the following categories of functions (where not far from this now, it just needs more consistency). 

Given that these functions may be composed in many different ways, the hope is that many of them will be shared, the most flexible arrangement is to have each function in its own file, named to match the function name. Given the consistent function naming scheme, I think discovering and reusing existing functions will work well and we should avoid needless duplication of functionality. 

* selectors: named `selectFoo()`, all accept `app.Store` as their only argument, and return types such as `selectors.Task` (as opposed to `stores.Task`). Can be named `selectCurrentFoo()` when they also accept the route and return the item with id from the current route. Their implementations consist exclusively of calls to the helper functions below, and they contain no loops or branches --- this is so that they need not be unit tested. Selector functions need to be aggressively broken up into smaller selector functions and other helpers, because many are too big. Big bonus: we avoid unit tests that work with the entire app store. They are called from `mapStateToProps()` functions. They are closely coupled to the view, since they build data structures exactly as the views need them to be. Another big bonus: since there is no massaging or other logic in `mapStateToProps()` functions, we don't need to unit test those either.
* pickers: named `pickFoo()`, all accept `app.Store` as their only argument, and return a collection from some sub store, or pickFooById(id: string, store: app.Store)` to return a single entity by its `id`. These functions are the only ones outside of reducers that know the details of how the store is structured. They are all one-line functions and are not unit tested.
* filters: named `filterFooByBar()`, they contain the logic for taking one or more collections of store data and returning the subset needed for the view, such as the tasks tagged with a given taxonomy. They accept and return types such as `stores.Task` (as opposed to `selectors.Task`. These functions are unit tested.
* builders, named `toSelectorFoo()` or `toSelectorFooFromBarAndBaz()`, they accept one or more collection of `stores` data structures such as `stores.Task` and `stores.TaskUserSettings`, and return data structures as needed for the view, i.e. collections of `selector` data structures such as `selectors.Task`. These functions are unit tested. 
* All functions should work with the "model" flavour of objects, only the selectors and buiders should return the "selector" flavour of objects.
[...]

PR #194, opened 2018-08-09
Feature/issue 183 persist more data

rasmus-storjohann-PG:


rasmus-storjohann-PG:
@bensinjin I would suggest that you review this PR by looking at the diff for all the commits except the last one. The last commit touches a lot of code that the rest of the PR doesn't, so it's better to review the last commit separately. I just created a tmp branch at the second-to-last commit and diffed against master.

rasmus-storjohann-PG:
Never mind, I removed the last commit from this PR, I'll put it up separately after this is merged.

src/sagas/questionnaire.ts
```diff
@@ -0,0 +1,59 @@
...
+        console.error(`Failed to load active questions (${error.message})`);
+        yield put(Persistence.loadFailure(error.message));
+    }
+}
+
+export async function loadActiveQuestionsAsync(): Promise<string> {
+    return await AsyncStorage.getItem(ACTIVE_QUESTIONS_STORAGE_KEY);
+}
+
+const deserialize = (ids: string): ReadonlyArray<Id> | undefined => (
```
rasmus-storjohann-PG: Remove undefined from return type

src/sagas/__tests__/questionnaire.test.ts
```diff
@@ -0,0 +1,118 @@
...
+
+    it('should dispatch select effect to get active question ids from the store', () => {
+        const saga = saveActiveQuestions();
+
+        const result = saga.next().value;
+
+        expect(result).toEqual(select(selectIdsOfActiveQuestions));
+    });
+
+    describe('after selectig active question ids from store', () => {
```
bensinjin: Fix selectig typo

src/sagas/questionnaire.ts
```diff
@@ -0,0 +1,59 @@
+// tslint:disable:no-expression-statement
+
+import { call, CallEffect, PutEffect, put, ForkEffect, takeLatest, select, SelectEffect } from 'redux-saga/effects';
+import { AsyncStorage } from 'react-native';
+import { ACTIVE_QUESTIONS_STORAGE_KEY } from '../application/constants';
+import { Persistence, Id } from '../stores/questionnaire';
+import * as constants from '../application/constants';
+import { selectIdsOfActiveQuestions } from '../selectors/select_ids_of_active_questions';
```
bensinjin: This name isn't great to me it looks as if it's going to return an array of question ids.  The term "Active" is somewhat vague I'd probably opt for something like "selected" or "chosen". For this function `selectChosenAnswerIds` or `getSelectedAnswerIds` would likely be an improvement.

src/sagas/questionnaire.ts
```diff
@@ -0,0 +1,59 @@
...
+export function* watchAnswerChangesToSaveAllAnswerStates(): IterableIterator<ForkEffect> {
+    yield takeLatest(constants.SELECT_ANSWER, saveActiveQuestions);
+}
+
+type SaveActions = IterableIterator<SelectEffect | CallEffect | PutEffect<Persistence.SaveSuccessAction | Persistence.SaveFailureAction>>;
+
+export function* saveActiveQuestions(): SaveActions {
+    try {
+        const ids = yield select(selectIdsOfActiveQuestions);
+        const serializedIds = serialize(ids);
```
bensinjin: Is serialization a requirement here?

src/sagas/questionnaire.ts
rasmus-storjohann-PG: Yeah, I have a commit waiting that changes all instances of "selected" and "active" to "chosen", it was a mess.

bensinjin: Also for my knowledge. Did you have a conventions around naming selectors? It seems we follow select<some-format><some-context> or select<some-context><some-format> did you have a preference?

rasmus-storjohann-PG: That's all summarized in https://github.com/pg-irc/pathways-frontend/issues/123, please have a look and let me know if you disagree with any of it.

PR #205, opened 2018-08-10
Feature/issue 183 handle loading and error states

rasmus-storjohann-PG:


rasmus-storjohann-PG:
The store has properties that only have meaningful values in some states and not others: error message in the error state, progress in the loading state. Using classes we can express this, and we avoid the problems of stale values sitting in error and progress fields when they are not relevant. Also, being able to restore the last known good state from an error state is nice. We could do that with other approaches too but here we use the type system to ensure that the state that comes back after clearing the error really is valid.

bensinjin:
I like it.

rasmus-storjohann-PG:
Done

PR #220, opened 2018-08-31
Feature/issue 219 persist my saved tasks

rasmus-storjohann-PG:


rasmus-storjohann-PG:
See related server side change here https://github.com/pg-irc/pathways-backend/pull/235/

Issue #211, opened 2018-08-21
Mark some articles and tasks as recommended to all


rasmus-storjohann-PG:
ACs: Some articles and tasks are recommended to all users, using the taxonomy term `recommendation:recommendToAll`

rasmus-storjohann-PG:
Three PRs related to this story, but they're all trivial: https://github.com/pg-irc/pathways-backend/pull/236, https://github.com/pg-irc/pathways-frontend/pull/221, https://github.com/PeaceGeeksSociety/content/pull/109/files

rasmus-storjohann-PG:
@bensinjin please have a look

PR #221, opened 2018-09-04
Issue #211 Remove a unused field, a taxonomy term is used to indicate…

rasmus-storjohann-PG:
… that a task is recommended to all


bensinjin:
Works for me

Issue #157, opened 2018-07-18
MVP polish: Replace hard-coded data in Find services page


rasmus-storjohann-PG:
In the Find Services tab on a task details page, replace hard-coded address and opening hours with live data from the service response. Limit scope to what can be done without any server side work and without adding `O(n)` calls to the server.

bensinjin:
In my opinion this issue is blocked on the approval and implementation of the aforementioned spec. Once the data we require from services_at_location endpoint is further defined we can make the necessary adjustments in the app to pull in the correct fields.

PR #229, opened 2018-09-11
Issue #157 Remove hardcoded service data

bensinjin:
This PR has an associated backend PR: https://github.com/pg-irc/pathways-backend/pull/242

This PR pulls in phone number data from the services_at_locations endpoint and removes anything hardcoded. Visually it will need further work. Implementation wise it is very basic as the AC was somewhat light. I expect this to evolve and also feel it's likely worth looking into tools like serialzr to better convert between our API responses and their associated TS Objects.

### DIFF
--- a/src/stores/services/index.ts
+++ b/src/stores/services/index.ts
@@ -1,6 +1,7 @@
+import * as R from 'ramda';
 import { LocalizedText } from '../../locale';
 
-import { Id, Service, ServiceStore, ServiceMap, TaskServices } from './types';
+import { Id, Service, ServiceStore, ServiceMap, TaskServices, PhoneNumber, APIPhoneNumber } from './types';
 import { UpdateTaskServicesAsync, updateTaskServicesAsync } from './update_task_services';
 import * as constants from '../../application/constants';
 import { Action } from 'redux';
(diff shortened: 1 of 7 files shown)


src/stores/services/index.ts
```diff
@@ -11,10 +12,19 @@ export { UpdateTaskServicesAsync, updateTaskServicesAsync };
...
+    const phoneNumbers: ReadonlyArray<PhoneNumber> = data.location.phone_numbers ?
+        APIPhoneNumbersToStorePhoneNumbers(data.location.phone_numbers) : [];
+    return { id, name, description, phoneNumbers };
+}
+
+function APIPhoneNumbersToStorePhoneNumbers(phoneNumbers: ReadonlyArray<APIPhoneNumber>): ReadonlyArray<PhoneNumber> {
+    return R.map((phoneNumber: APIPhoneNumber): PhoneNumber => ({
+        type: phoneNumber.phone_number_type,
+        phoneNumber: phoneNumber.phone_number,
+    }), phoneNumbers);
```
rasmus-storjohann-PG: Do you want to look into using serializr for this? I think it could go into this PR. I'd also want to see this function validate the data (what data is valid), and see unit tests around that.

src/stores/services/types.ts
```diff
@@ -2,10 +2,21 @@ import { LocalizedText } from '../../locale';
 
 export type Id = string;
 
+export interface APIPhoneNumber {
```
rasmus-storjohann-PG: Naming: the thing that distinguishes this type from `PhoneNumber` is that it has not been validated yet. So while it's cumbersome, I suggest `NonValidatedPhoneNumber`. The type won't be used widely so the name shouldn't be a big problem.

src/stores/services/index.ts
```diff
@@ -11,10 +12,19 @@ export { UpdateTaskServicesAsync, updateTaskServicesAsync };
 export function serviceFromServiceData(data: any): Service { // tslint:disable-line:no-any
     // TODO: Perform appropriate data validation.
     //       Alternatively bring in a tool to do this for us, eg: Serializr
-    const id: string = data.id || undefined;
-    const name: LocalizedText = { 'en': data.name || '' };
-    const description: LocalizedText = { 'en': data.description || '' };
-    return { id, name, description };
+    const id: string = data.service.id || undefined;
+    const name: LocalizedText = { 'en': data.service.name || '' };
```
rasmus-storjohann-PG: So there was a time when we thought the BC211 data would be translated. Looking at the amount of data involved, this is unlikely to happen (unless the BC211 organization takes that on, I have not heard that they will...), so perhaps we should just be honest and use `string` rather than `LocalizedText` for the service name and description... not sure

src/stores/services/index.ts
```diff
@@ -11,10 +12,19 @@ export { UpdateTaskServicesAsync, updateTaskServicesAsync };
...
+    const description: LocalizedText = { 'en': data.service.description || '' };
+    const phoneNumbers: ReadonlyArray<PhoneNumber> = data.location.phone_numbers ?
+        APIPhoneNumbersToStorePhoneNumbers(data.location.phone_numbers) : [];
```
rasmus-storjohann-PG: I would do the check against falsy inside `APIPhoneNumbersToStorePhoneNumbers`. It would just be a little tidier here, and there is also the clean-code principle of not  [mixing levels of abstraction](http://principles-wiki.net/principles:single_level_of_abstraction) which sort-of applies here.

src/stores/services/index.ts
bensinjin: I'm going to do that - moving this back to in progress...

bensinjin:
A new PR will be opened to address the comments here around api response validation this work will be included in that PR.

PR #233, opened 2018-09-14
Feature/issue 216 update libraries

rasmus-storjohann-PG:


src/selectors/questionnaire/get_ids_from_answer_map.ts
```diff
@@ -1,6 +1,6 @@
-import { Id, AnswersMap } from '../../stores/questionnaire';
+import { Id, Answer, AnswersMap } from '../../stores/questionnaire';
 import * as R from 'ramda';
 
 export const getIdsFromAnswerMap = (answers: AnswersMap): ReadonlyArray<Id> => (
-    R.map(R.prop('id'), R.values(answers))
+    R.map<Answer, Id>(R.prop('id'), R.values(answers))
```
bensinjin: It'd be nice to clean up this syntax to something like: `R.map<Answer, Id>(answer => answer.id, R.values(answers))`

PR #251, opened 2018-09-28
Feature/issue 237 hide back button on main pages

rasmus-storjohann-PG:


rasmus-storjohann-PG:
@bensinjin for review, can you also take this for a spin on an android device and let me know if you think the behaviour is correct? An iDevice maybe too.

PR #252, opened 2018-09-28
Feature/issue 238 change sort order of tasks and articles

rasmus-storjohann-PG:


rasmus-storjohann-PG:
@bensinjin for review

src/selectors/__tests__/helpers/task_helpers.ts
```diff
@@ -0,0 +1,46 @@
+// tslint:disable:readonly-keyword no-this no-expression-statement readonly-array no-class
+
+import { aString, aBoolean } from '../../../application/__tests__/helpers/random_test_values';
+import { TaxonomyTermReference } from '../../../stores/taxonomies';
+import { ExploreSection } from '../../explore/types';
+import { Task } from '../../tasks/task';
+import { TaskListItem } from '../../tasks/task_list_item';
+import { ArticleListItem } from '../../articles/article_list_item';
+
+export class ViewTaskBuilder {
```
bensinjin: If you're just using this to test your sort function could you not just use the existing `TaskBuilder`?

bensinjin: Ah yeah I see that now, makes sense.

PR #253, opened 2018-09-29
Feature/issue 239 polish nav bar

rasmus-storjohann-PG:


src/components/main/footer.tsx
```diff
@@ -1,9 +1,11 @@
 import React from 'react';
+import * as ReactNative from 'react-native';
```
bensinjin: This feels awkward. I think we should shy away from import * when at all possible as you would in most environments.

rasmus-storjohann-PG: Fixed

PR #260, opened 2018-10-10
Feature/issue 243 questionnaire popup

bensinjin:
This is currently set to show after three minutes. You can manually set the value in `modal_notification.tsx` to test.  A couple file renames (`notifications_list` -> `notifications` and `notification` -> `expiring_notification`) make this PR look larger than it is.


rasmus-storjohann-PG:
One way to make PRs with file renames easier to review is to do the renames in its own commit.

src/components/notifications/modal_notification.tsx
```diff
@@ -0,0 +1,159 @@
...
+type Props = ModalNotificationProps & ModalNotificationActions & RouterProps;
+
+interface State {
+    readonly timeHasElapsed: boolean;
+    readonly modalBottomPosition: number;
+}
+
+export class ModalNotificationComponent extends React.Component<Props, State> {
+    timer: number;
+    displayAfterMilliseconds: number = 180000; // Three minutes in milliseconds
```
rasmus-storjohann-PG: I usually make this clear by doing 
```
displayAfterMilliseconds: number = 3 * 60 * 1000;
```
and/or use the variable name `threeMinutesInMilliseconds`

src/components/notifications/modal_notification.tsx
```diff
@@ -0,0 +1,159 @@
...
+    private getComponentForContent(): JSX.Element {
+        if (this.shouldDisplayModal()) {
+            if (this.props.notification.type === NotificationType.QuestionnaireInformation) {
+                return this.withModalHeader(this.getContentForQuestionnaireInformation());
+            }
+        }
+        return <EmptyComponent />;
+    }
+
+    private shouldDisplayModal(): boolean {
```
rasmus-storjohann-PG: The name is a little confusing, does a false return value imply don't display, or display non-modally?

src/components/notifications/modal_notification.tsx
```diff
@@ -0,0 +1,159 @@
...
+                    <Button onPress={onPress} small>
+                        <Icon name='close' />
+                    </Button>
+                </View>
+                <View>{content}</View>
+            </View>
+        );
+    }
+
+    private getContentForQuestionnaireInformation(): JSX.Element {
```
rasmus-storjohann-PG: This function should not be part of this class, since it breaks the single responsibility principle for this class.

src/components/notifications/modal_notification.tsx
bensinjin: Fixed

src/components/notifications/modal_notification.tsx
bensinjin: Fixed the name. This check basically indicates whether or not we show anything so I've renamed it `shouldDisplay()`

src/components/notifications/modal_notification.tsx
bensinjin: Fixed for this as well as the other notifications. Much cleaner now.

bensinjin:
@rasmus-storjohann-PG this is ready for another check.

Issue #246, opened 2018-09-26
Make questionnaire status message clickable


rasmus-storjohann-PG:
ACs:
* Clicking the "N tasks recommended" label in the questionnaire screen navigates to the My Plan page
* After a question is selected or deselected so that the number of recommended tasks changes, the label expands to take up more screen space, containing more explanation of what's going on. After a short delay it shrinks back down to the small size.

PR #261, opened 2018-10-10
Issue #246 Make task counter clickable and add indication of tasks be…

bensinjin:
…ing added and removed

### DIFF
--- a/src/components/questionnaire/floating_task_counter.tsx
+++ b/src/components/questionnaire/floating_task_counter.tsx
@@ -19,25 +19,24 @@ interface State {
 type Props = FloatingTaskCounterProps;
 
 export class FloatingTaskCounter extends React.Component<Props, State> {
-    readonly taskCountChangeDefault: number = 0;
     timer: number;
 
     constructor(props: Props) {
         super(props);
         this.state = {
-            taskCountChange: this.taskCountChangeDefault,
+            taskCountChange: 0,
         };
     }
 
     componentDidUpdate(previousProps: Props): void {
         const difference = this.props.taskCount - previousProps.taskCount;
-        if (difference !== this.taskCountChangeDefault) {
+        if (difference !== 0) {
             this.setState({
                 taskCountChange: difference,
             });
             this.timer = setTimeout(() => {
                 this.setState({
-                    taskCountChange: this.taskCountChangeDefault,
+                    taskCountChange: 0,
                 });
             }, 1500);
         }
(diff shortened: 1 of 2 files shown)


src/components/questionnaire/floating_task_counter.tsx
```diff
@@ -0,0 +1,111 @@
...
+                <Text style={styles.largeFloatingText}>{this.props.taskCount} </Text>
+            </View>
+        );
+    }
+
+    private getTaskCountText(): JSX.Element {
+        return (
+            <Text style={styles.largeFloatingText}>{this.props.taskCount} </Text>
+        );
+    }
```
rasmus-storjohann-PG: These two functions divide up their work in a way that is confusing, and also violates SRP and DRY. If you had one function that returns either the difference `Text` component or an `EmptyComponent` component, and a second function that returns just the task count `Text` componnent, it will be better, I think.

src/components/questionnaire/floating_task_counter.tsx
```diff
@@ -0,0 +1,111 @@
...
+                </Text>
+                <Text style={styles.smallFloatingText}>
+                    <Trans>recommended</Trans>
+                </Text>
+            </TouchableOpacity>
+        );
+    }
+
+    private shouldDisplayTaskCountChange(): boolean {
+        return this.state.taskCountChange !== this.taskCountChangeDefault;
```
rasmus-storjohann-PG: I see that you want to avoid magic numbers in your code, but I think using just `0` for `this.taskCountChangeDefault` throughout will actually make it easier to follow.

src/components/questionnaire/floating_task_counter.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+    componentWillUnmount(): void {
+        clearTimeout(this.timer);
+    }
+
+    render(): JSX.Element {
+        return (
+            <TouchableOpacity
+                style={styles.floatingCount}
+                onPress={goToRouteWithoutParameter(Routes.MyPlan, this.props.history)}
```
rasmus-storjohann-PG: I'm working on finding a way to make the callback set in `mapDispatchToProps()` take the history from `ownProps` so that it doesn't need to be passed around so much. Try it out if you like.

src/components/questionnaire/floating_task_counter.tsx
bensinjin: Fixed

src/components/questionnaire/floating_task_counter.tsx
bensinjin: Yeah this did feel somewhat awkward .. fixed.

src/components/questionnaire/floating_task_counter.tsx
bensinjin: Yeah I'll take a crack at this. Moving back to in progress.

bensinjin: We've decided to hold off on this as similar work is being done in other branches.

bensinjin:
@rasmus-storjohann-PG ok I think this is ready for a final once over.

PR #266, opened 2018-10-12
Feature/issue 244 business logic

rasmus-storjohann-PG:


src/selectors/__tests__/tasks.test.ts
```diff
@@ -88,58 +90,130 @@ describe('tasks selector', () => {
...
+            incompleteTask = new TaskBuilder().withTaxonomyTerm(aTaxonomyTerm).withCompleted(false).build();
+            incompleteTasks = { [incompleteTask.id]: incompleteTask };
         });
 
-        it('includes tasks that are not saved in my plan', () => {
-            const task = new TaskBuilder().withLocaleCode(locale.code).build();
-            const noSavedTaskIds: ReadonlyArray<string> = [];
-            const result = rejectTasksWithIdsInList(noSavedTaskIds, [task]);
-            expect(result).toEqual([task]);
+        it('should include a tasks that was not recommended and is recommended', () => {
```
bensinjin: These could use a reword. For this instance maybe something like `'returns tasks that were previously not recommended but are now recommended'`.

src/selectors/questionnaire/pick_questionnaire.ts
```diff
@@ -1,6 +1,6 @@
 import { Store } from '../../stores';
 import { toValidOrThrow, ValidQuestionnaireStore } from '../../stores/questionnaire/stores';
 
-export const pullQuestionnaire = (appStore: Store): ValidQuestionnaireStore => (
+export const pickQuestionnaire = (appStore: Store): ValidQuestionnaireStore => (
```
bensinjin: Can you clarify when you chose pick over pull?

src/selectors/tasks/get_newly_recommended_tasks.ts
```diff
@@ -0,0 +1,13 @@
...
+import * as R from 'ramda';
+import { getRecommendedTasks } from './get_recommended_tasks';
+
+export const getNewlyRecommendedTasks =
+    (oldAnswers: AnswersMap, newAnswers: AnswersMap, tasks: TaskMap, savedTaskIds: ReadonlyArray<Id>): ReadonlyArray<Task> => (
+        R.difference(
+            getRecommendedTasks(newAnswers, tasks, savedTaskIds),
+            getRecommendedTasks(oldAnswers, tasks, savedTaskIds),
+        )
+    );
```
bensinjin: Nice :)

src/fixtures/types/questionnaire.ts
```diff
@@ -26,10 +26,18 @@ export interface AnswersMap {
     readonly [key: string]: Answer;
 }
 
+export enum State {
+    InQuestionnaire,
+    NotInQuestionnaire,
+    PopupNeeded,
+}
```
bensinjin: On first glance these states seem odd for a questionnaire. Maybe add some indication that you're referring to the questionnaire being viewed or the that the user is on the questionnaire route? PopupNeeded seems unrelated to the other states here and should likely be clarified further as well.

src/stores/questionnaire/reduce_valid_store.ts
```diff
@@ -27,6 +32,28 @@ export const reduceValidStore = (store: ValidQuestionnaireStore, action?: Questi
...
+            ...store,
+            oldAnswers: store.answers,
+            state: State.InQuestionnaire,
+        });
+    }
+
+    if (store.state === State.InQuestionnaire && !isNewRouteToQuestionnaire) {
+        return new ValidQuestionnaireStore({
+            ...store,
+            state: State.PopupNeeded,
```
bensinjin: So off the top of my head I think what we'd like to do is try and figure out a way to not have the `PopupNeeded` state at all and actually have the notifications store listen for the `RouteChangedAction` and then make a call to `addNotificationToStore`. We'd naturally have to tweak `modal_notification.tsx`'s props to allow a milliseconds value so you can have the modal display immediately.

src/fixtures/types/questionnaire.ts
bensinjin: This also ties into my last comment in this PR.

src/stores/questionnaire/reduce_valid_store.ts
rasmus-storjohann-PG: My intuition is that this popup dialog will potentially become pretty complex so I don't think using the notification for it will be appropriate. It will cover the whole screen or almost, it will probably have functionality to tick off which tasks to add to my plan and which to leave out, etc...

src/fixtures/types/questionnaire.ts
rasmus-storjohann-PG: I've renamed the enum values.

src/selectors/__tests__/tasks.test.ts
rasmus-storjohann-PG: Fixed

src/components/main/main.tsx
```diff
@@ -7,35 +7,46 @@ import { NotificationsConnectedComponent } from '../notifications/notifications_
...
 }
 
 export type Props = HeaderProps & FooterProps & RouterProps;
 
-export class Component extends React.Component<Props, DrawerState> {
+export interface Actions {
+    readonly routeChanged: (location: Location, action: Action) => RouteChangedAction;
+}
+
+export type MainComponentProps = Props & Actions;
```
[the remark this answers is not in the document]
rasmus-storjohann-PG: Fixed in a different PR I just put up... there may be some rebase conflicts between them, but shouldn't be too bad regardless of their merge order

bensinjin:
@rasmus-storjohann-PG this is looking good, can you do a rebase and fix up your commit messages too?

PR #271, opened 2018-10-15
Feature/issue 264 combine tasks and articles

bensinjin:
This: https://github.com/pg-irc/pathways-backend/pull/256 should be merged first and the ts files need to be regenerated locally otherwise you'll likely get errors using outdated fixtures.


src/stores/__tests__/helpers/tasks_helpers.ts
```diff
@@ -75,8 +69,7 @@ export class TaskBuilder {
             title: this.createLocalizedText(this.title),
             description: this.createLocalizedText(this.description),
             taxonomyTerms: this.taxonomyTerms,
-            relatedArticles: this.relatedArticles,
-            relatedTasks: this.relatedArticles,
```
rasmus-storjohann-PG: If we had made the planned change to make it impossible to assign one Id type to another (type safe id types), we would have caught this bug. I still want to do that sometime...

Issue #259, opened 2018-10-05
Polish service list in task view


rasmus-storjohann-PG:
ACs:
* List of services in the task view shows service physical address and phone numbers with the phone number types.

From email conversation about this work with App-Scoop:

We have a screen that contains a list of service providers. Currently it only shows the names and phone numbers, and it doesn't look great.

It would be very helpful if you could update this list screen to show the name, physical address and phone numbers of each service provider, and find a better way to show the phone number type. The phone number type can be phone, fax, toll-free and a bunch of other strings. Don't show the postal address.

PR #272, opened 2018-10-16
Clean up task service implementation

bensinjin:
This is some clean up around the address piece for task services. The goal here was to keep consistent with our already established pattern which maybe wasn't as obvious as it could be so I'm putting it here:

`request JSON from server -> validate JSON (throw if invalid) -> convert TS type representing valid JSON to TS type representing valid store object`

In this particular scenario each address from the server comes in the form: `{address_type: string, address: object}` once validated our TS type: `ValidatedAddressWithTypeJSON` is then converted to its store representation: `Address` which has its type field embedded.

### DIFF
--- a/src/components/services/service.tsx
+++ b/src/components/services/service.tsx
@@ -12,7 +12,7 @@ interface Props {
 
 export function ServiceComponent(props: Props): JSX.Element {
     const mapWithIndex = R.addIndex(R.map);
-
+    const physicalAddresses = R.filter(R.propEq('type', 'physical_address'), props.service.addresses);
     const capitalizeFirstLetter = (phoneType: string): string => (
         phoneType.charAt(0).toUpperCase() + phoneType.slice(1)
     );
@@ -29,7 +29,7 @@ export function ServiceComponent(props: Props): JSX.Element {
                 mapWithIndex((address: Address, index: number) =>
                     <View key={index}>
                         <Text>{address.address}</Text>
-                    </View>, props.service.addresses)
+                    </View>, physicalAddresses)
             }
             {
                 mapWithIndex((phoneNumber: PhoneNumber, index: number) =>
--- a/src/components/services/service.tsx
+++ b/src/components/services/service.tsx
@@ -1,10 +1,9 @@
 import React from 'react';
 import * as R from 'ramda';
 import { applicationStyles, colors } from '../../application/styles';
-import { Service, Address } from '../../stores/services';
+import { Service, PhoneNumber, Address } from '../../stores/services';
 import { View } from 'native-base';
 import { Text } from 'react-native';
-import { PhoneNumber } from '../../stores/services';
 import { TextWithPhoneLinks } from '../link/text_with_phone_links';
 
 interface Props {
@@ -18,10 +17,6 @@ export function ServiceComponent(props: Props): JSX.Element {
         phoneType.charAt(0).toUpperCase() + phoneType.slice(1)
     );
 
-    const getAddress = (address: Address): string => (
-        address === null ? '' : address.address
-    );
-
     return (
         <View>
             <Text style={[
@@ -30,7 +25,12 @@ export function ServiceComponent(props: Props): JSX.Element {
             ]}>
                 {props.service.name}
             </Text>
-            <Text>{getAddress(props.service.physicalAddress)}</Text>
+            {
+                mapWithIndex((address: Address, index: number) =>
+                    <View key={index}>
+                        <Text>{address.address}</Text>
+                    </View>, props.service.addresses)
+            }
             {
                 mapWithIndex((phoneNumber: PhoneNumber, index: number) =>
                     <View key={index}>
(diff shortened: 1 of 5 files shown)


rasmus-storjohann-PG:
Holding off on this PR until we can talk in person

bensinjin:
My reasonings about the changes I introduced:

-`AddressWithType` was unnecessary and added no real value. The only reference to `AddressWithType` was from the `AddressBuilder` which was never used.
- Putting the `type` right on the `Address` makes filtering addresses simple
- Using `addresses` on `Service` keeps our mapping between JSON type and store type simpler (see `serviceFromValidatedJSON`) The `postalAddress` was unused so it was removed.
- Previously `Address` was being used to represent both our JSON address and our store address but actually only reflected the JSON version so I've renamed it to `ValidatedAddressJSON` and the created `Address` for our store type.

rasmus-storjohann-PG:
My thinking for handling addresses and phone numbers differently is that there are a million different phone number types, but only two address types that I know of. However, I'm not going to insist on that, there also value to be consistent between phone numbers and addresses.

Generally, there can be value to having both `AddressWithType` and `Address`. When the store address type as `postal_address` and `physical_address`, we need an `Address ` type without the type field on it, since it would be redundant and we could get bugs where the `postal_address` could well end up with a type of `physical_address`, ugh. However, the json parser would output an address with the type field on it, so we'd need both. But if we go with the same approach as phone numbers, then we need the type field. We end up with more business logic in the presentation to pick out the address, where you can get bugs like not just having a missing physical address, but you could have multiple addresses with that type. (My reasoning around what data structure to use is really to find a solution where invalid situations are impossible to represent).

src/components/services/service.tsx
```diff
@@ -30,7 +25,12 @@ export function ServiceComponent(props: Props): JSX.Element {
             ]}>
                 {props.service.name}
             </Text>
-            <Text>{getAddress(props.service.physicalAddress)}</Text>
+            {
+                mapWithIndex((address: Address, index: number) =>
+                    <View key={index}>
+                        <Text>{address.address}</Text>
+                    </View>, props.service.addresses)
+            }
```
rasmus-storjohann-PG: We should just show physical address, I don't think we have a use case yet for postal address, let's keep things simple for now.

src/stores/__tests__/helpers/services_helpers.ts
```diff
@@ -55,43 +50,30 @@ export class PhoneNumberBuilder {
 
 export class AddressBuilder {
     type: string = aString();
-
     address: string = aString();
     city: string = aString();
-    state_province: string = aString();
-    postal_code: string = aString();
+    province: string = aString();
```
rasmus-storjohann-PG: The HSDS standard uses `state_province`, so we should be consistent and use `stateProvince`
https://openreferral.readthedocs.io/en/latest/hsds/reference/#physical_address

src/stores/services/index.ts
```diff
@@ -16,30 +16,23 @@ export function serviceFromValidatedJSON(data: ValidatedServiceAtLocationJSON):
...
-    var physicalAddress: Address = null;
-    var postalAddress: Address = null;
-    addressesWithType.forEach(element => {
-        if (element.type.match(ADDRESS_TYPE_PHYSICAL)) {
-            physicalAddress = element.address;
-        }else if (element.type.match(ADDRESS_TYPE_POSTAL)) {
-            postalAddress = element.address;
-        }
-    });
+    const physicalAddresses = R.filter(R.propEq('type', 'physical_address'), addresses);
```
rasmus-storjohann-PG: If the field on the DTO is called `addresses`, the strong implication is that it will contain all the addresses, so there should be no such filtering. This problem has now been transferred to the presentation layer.

src/stores/services/types.ts
```diff
@@ -6,25 +6,20 @@ export interface PhoneNumber {
 }
 
 export interface Address {
+    readonly type: string;
     readonly address: string;
     readonly city: string;
-    readonly state_province: string;
-    readonly postal_code: string;
+    readonly province: string;
```
rasmus-storjohann-PG: Fix naming here too

rasmus-storjohann-PG:
Can you update all commit messages to reference #259?

src/stores/__tests__/helpers/services_helpers.ts
```diff
@@ -55,43 +50,30 @@ export class PhoneNumberBuilder {
...
-        this.type = type;
-        return this;
-    }
-
     build(): Address {
         return {
+            type: this.type,
             address: this.address,
             city: this.city,
-            state_province: this.state_province,
```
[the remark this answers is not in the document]
bensinjin: Fixed

src/components/services/service.tsx
bensinjin: This is filtered now in this component so it's clearer we're only showing physical addresses.

src/stores/__tests__/helpers/services_helpers.ts
bensinjin: Fixed

src/stores/services/index.ts
bensinjin: Fixed

src/stores/services/types.ts
bensinjin: Fixed

bensinjin:
@rasmus-storjohann-PG Cheers for the feedback. I do see your points as well. I've addressed the issues you've outlined.

PR #282, opened 2018-10-25
Issue #281 Install missing package for lingui string extraction

rasmus-storjohann-PG:


rasmus-storjohann-PG:
@bensinjin for review

PR #284, opened 2018-10-26
Issue pg-irc/pathways-backend#251 Add missing schema validation tests…

bensinjin:
… for address fields, organize tests into logical groups, allow null values for address, state_province, and postal_code

### DIFF
--- a/src/stores/__tests__/helpers/services_schema_helpers.ts
+++ b/src/stores/__tests__/helpers/services_schema_helpers.ts
@@ -72,7 +72,7 @@ export class PhoneNumberJSONBuilder {
 
 export class AddressJSONBuilder {
     id: any = anInteger();
-    address: any | number = aString();
+    address: any = aString();
     city: any = aString();
     state_province: any = aString();
     postal_code: any = aString();
--- a/src/stores/__tests__/helpers/services_schema_helpers.ts
+++ b/src/stores/__tests__/helpers/services_schema_helpers.ts
@@ -1,6 +1,9 @@
 // tslint:disable:no-class readonly-keyword no-expression-statement no-this no-let no-any
 import { aString, anInteger } from '../../../application/__tests__/helpers/random_test_values';
 
+// The following "JSON" types exist purely to allow us to test our schemas with invalid data.
+// The optional and any properties on these types ensure we can create invalid objects.
+
 interface PhoneNumberJSON {
     readonly phone_number_type?: any;
     readonly phone_number?: any;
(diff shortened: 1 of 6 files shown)


bensinjin:
This fixes our crash on tasks that pull in services with null values for their postal code.

bensinjin:
I likely should have put the reorganization of the schema tests into its own commit. I'll be sure to remember to better isolate noisy refactoring commits.

src/stores/__tests__/helpers/services_schema_helpers.ts
```diff
@@ -71,34 +71,39 @@ export class PhoneNumberJSONBuilder {
...
-    id: number = anInteger();
-    address: string | null = aString();
-    city: string | null = aString();
-    state_province: string | null = aString();
-    postal_code: string | null = aString();
-    country: string | null = aString();
-
-    withAddress(address: string | null): AddressJSONBuilder {
+    id: any = anInteger();
+    address: any | number = aString();
```
rasmus-storjohann-PG: `any | number` type is equivalent to just `any`

src/stores/__tests__/helpers/services_schema_helpers.ts
```diff
@@ -1,29 +1,29 @@
-// tslint:disable:no-class readonly-keyword no-expression-statement no-this no-let
+// tslint:disable:no-class readonly-keyword no-expression-statement no-this no-let no-any
 import { aString, anInteger } from '../../../application/__tests__/helpers/random_test_values';
 
 interface PhoneNumberJSON {
-    readonly phone_number_type?: string | null;
-    readonly phone_number?: string | null;
+    readonly phone_number_type?: any;
+    readonly phone_number?: any;
```
rasmus-storjohann-PG: So why are you changing `string | null` to `any`?

src/stores/__tests__/helpers/services_schema_helpers.ts
bensinjin: Find and replace error. Fixed.

src/stores/__tests__/helpers/services_schema_helpers.ts
bensinjin: `null` used to be used to represent the invalid type. Now `null` is valid in some cases so we'd need to pick a new invalid type. Naturally we could tack on ` | number` but I felt that would confuse things. These types exist for the soul purpose of letting us test the schemas with missing values and values of incorrect types maybe the naming could be improved to better reflect this?

rasmus-storjohann-PG: OK, the reason is good but the situation is unusual. I would add a comment explaining this. I would probably break out the `no-any` from 
```ts
// tslint:disable:no-class readonly-keyword no-expression-statement no-this no-let no-any
```
to have it on its own with the comment explaining, but there are other ways to do it too.

bensinjin: I've added a comment.

PR #286, opened 2018-10-30
Feature/issue 248 persist completed flag

rasmus-storjohann-PG:


bensinjin:
I like 8c4cd7516c9beda64ae3082b16dcad0f5b319702 these 

> @bensinjin for review. Have a close look at the first refactor commit. I tried out writing a different style of functions with ramda, let me know if you think this makes things better.

Looks nice and potentially more reusable so likely better :)

PR #292, opened 2018-11-02
Feature/issue 61 clean up recommended tasks selector

rasmus-storjohann-PG:


rasmus-storjohann-PG:
@bensinjin a PR with some more Ramda cleanup in the same spirit as the one a couple of days ago. Not a rush to get this in...

PR #298, opened 2018-11-13
Fix broken syntax causing nav labels to not show up

bensinjin:


rasmus-storjohann-PG:
Can you change the commit message to contain the story #? Whichever of the current stories that fits the closest.

PR #303, opened 2018-11-15
Feature/issue 299 button for clearing local data

rasmus-storjohann-PG:


rasmus-storjohann-PG:
@bensinjin for review. Glad I got something working today!!

src/components/help/help_component.tsx
```diff
@@ -34,46 +35,85 @@ const fixture: ReadonlyArray<HelpContact> = [
     },
 ];
 
-export const HelpComponent: React.StatelessComponent = (): JSX.Element => {
+export interface HelpComponentProps {
```
bensinjin: I've been thinking ... how would you feel about just having one type we export so we can mix actions and props together? It's a weird distinction we make between actions and props when they're all just props really. In this case we'd just export something called `type HelpComponentProps`. Do we gain something from having two separate interfaces?

src/components/help/help_component.tsx
```diff
@@ -34,46 +35,85 @@ const fixture: ReadonlyArray<HelpContact> = [
...
+            'answers are chosen in the questionnaire, which tasks are in My Plan and which ' +
+            'tasks are marked as completed';
+
+        // tslint:disable-next-line:no-expression-statement
+        Alert.alert(alertHeading, alertMessage,
+            [
+                // tslint:disable-next-line:no-empty
+                { text: 'Cancel', onPress: (): void => { }, style: 'cancel' },
+                // tslint:disable-next-line:no-expression-statement
+                { text: 'Delete all user data', onPress: (): void => { props.clearAllUserState(); } },
```
bensinjin: I think this syntax: `onPress: (): void => props.clearAllUserState()` will remove the need to disable the ts rule.

src/components/help/help_connectede_component.tsx
```diff
@@ -0,0 +1,13 @@
...
+import { clearAllUserData, ClearAllUserDataAction } from '../../stores/questionnaire/actions';
+
+const mapStateToProps = (_: Store): HelpComponentProps => ({
+});
+
+const mapDispatchToProps = (dispatch: Dispatch<Store>): HelpComponentActions => ({
+    clearAllUserState: (): ClearAllUserDataAction => dispatch(clearAllUserData()),
+});
+
+export const HelpConnectedComponent = connect(mapStateToProps, mapDispatchToProps)(HelpComponent);
```
bensinjin: Fix typo in this file's name

src/components/help/help_component.tsx
rasmus-storjohann-PG: Hey, I'd love to do that, but the `connect()` API needs this separation so I don't see a way to do this cleanly.

src/components/help/help_component.tsx
rasmus-storjohann-PG: Fixed

src/components/help/help_connectede_component.tsx
rasmus-storjohann-PG: Fixed

rasmus-storjohann-PG:
@bensinjin addressed your comments

src/components/help/help_component.tsx
bensinjin: Oh right, yeah unfortunate

Issue #304, opened 2018-11-15
No longer navigate from questionnaire floating popup to My Plan


rasmus-storjohann-PG:
It's easy to click this by mistake and it brings up the Newly Recommended popup which is just really confusing.

AC: Clicking the floating popup in the questionnaire does nothing

PR #305, opened 2018-11-15
Issue #304 No longer navigate to My Plan when clicking the floating p…

rasmus-storjohann-PG:
…opup in the questionnaire


rasmus-storjohann-PG:
@bensinjin for review

Issue #232, opened 2018-09-12
Upgrade to Redux 4


rasmus-storjohann-PG:
See https://dev.to/resir014/redux-4--typescript-29-a-type-safe-approach-2lf4

PR #307, opened 2018-11-20
Issue #232 Upgrade Redux to 4.0.0 and fix TS errors

bensinjin:
So this PR upgrade Redux to 4.0.0 and updates our Dispatch types to fix all of our TS errors. There are likely gains to be had in changes to syntax introduced in this version but I've only focused on ensuring we can build without errors and the app functions with this new version of Redux.


bensinjin:
Also, I've ensured we're using the `Dispatch` type definition from redux, not react-redux, which looks to be out of date. I wonder if we actually need the react-redux types as redux ships when its own.

Issue #301, opened 2018-11-14
Update heroku deployment of fresh 211 data with similarity data


rasmus-storjohann-PG:
Lots of useful stuff in https://github.com/pg-irc/pathways-backend/issues/272

PR #312, opened 2018-11-28
Issue #301 Change client API call to pass just the task id, remove se…

rasmus-storjohann-PG:
…rvice query property


rasmus-storjohann-PG:
@bensinjin for review

Issue #224, opened 2018-09-07
Make it possible to select text in the app for copy-paste


rasmus-storjohann-PG:
ACs: 
* The user can select and copy text from text in the app, not including the tasks description.

Edit: changed the AC to explicitly exclude task description, previously this was explicitly included.

PR #316, opened 2018-11-29
Issue #224 Create selectable text component and use it in a few places

rasmus-storjohann-PG:


rasmus-storjohann-PG:
@bensinjin for review

bensinjin:
Looks good.

Issue #313, opened 2018-11-29
Add top level exception handler that reports a crash to the server


rasmus-storjohann-PG:
ACs: 
* Look into using Sentry
* To trigger a crash: change the definition of the schema so that validation fails, this throws in a saga, we need to catch that exception.

PR #317, opened 2018-11-30
Issue #313 Clean up services saga and document Sentry integration

bensinjin:
Service validation errors are now yielded, like http errors, so we get useful error messages from our services saga when validation fails. As it stands we're not currently surfacing these errors but I did look into how to do that with Sentry.

I was able to get it working so our exceptions can be passed from our error boundary over to sentry. Which is pretty neat. This is what it looks like in Sentry when validation fails on a service from our API:

screen shot 2018-11-29 at 5 01 28 pm

The stack trace is also really nice and we also can assign these errors to people kind of like Github tickets.

Anyways, to actually get this in place we need to do a couple things:

- Set up a Sentry account for Peacegeeks
- Follow the set up here: https://docs.expo.io/versions/latest/guides/using-sentry
- Move our error boundary into App.js. I can only get errors to show up in Sentry when done like this, although, I think if we can get this: https://github.com/expo/sentry-expo/blob/master/index.js file's version of Sentry into our ErrorComponent we may be able to keep them separate. Something to try again, anyways, for a reference, my working App.js looked like:

```
import React from 'react';
import { Application } from './lib/application';
import Sentry from 'sentry-expo';

Sentry.config('Public DSN goes here').install();

export default () => (
    <ErrorBoundary>
        <Application />
    </ErrorBoundary>
);

class ErrorBoundary extends React.Component {
    componentDidCatch(error, errorInfo) {
        Sentry.captureException(error, {
            extra: errorInfo
        });
    }
    render() {
        return this.props.children;
    }
}
```
- Test in production mode

### DIFF
--- a/App.js
+++ b/App.js
@@ -1,3 +1,7 @@
 import { Application } from './lib/application';
+import Sentry from 'sentry-expo';
+
+// DSN for PeaceGeek's pathways project.
+Sentry.config('https://de23d08a2e2b44a49849c045c2a6fc0c@sentry.io/256284').install();
 
 export default Application;
--- a/app.json
+++ b/app.json
@@ -15,6 +15,17 @@
       "assets/images/*",
       "assets/fonts/*",
       "node_modules/native-base/Fonts"
-    ]
+    ],
+    "hooks": {
+      "postPublish": [
+        {
+          "file": "sentry-expo/upload-sourcemaps",
+          "config": {
+            "organization": "peacegeeks",
+            "project": "pathways"
+          }
+        }
+      ]
+    }
   }
 }
--- a/.env.example
+++ b/.env.example
@@ -1,1 +1,2 @@
 API_URL=https://pathways-production.herokuapp.com/v1
+SENTRY_AUTH_TOKEN=your auth token
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ Install dependencies
 yarn install
 ```
 
-Copy example configuration file and modify accordingly (set your own API host, etc).
+Copy example configuration file and modify accordingly (set your own API host, Sentry auth token etc).
 ```
 cp .env.example .env
 ```
--- a/package.json
+++ b/package.json
@@ -91,6 +91,7 @@
     "redux": "^4.0.0",
     "redux-persist": "^5.9.1",
     "redux-saga": "^0.16.0",
+    "sentry-expo": "^1.10.0",
     "uuid": "^3.3.2"
   },
   "lingui": {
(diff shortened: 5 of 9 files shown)


rasmus-storjohann-PG:
I like this, I think we should proceed with this. It's unlikely to cost us more than USD26/month (unless we have a bug causing our app to spew millions of errors, let's not do that). @robinpuga for review, and Arti as well.

rasmus-storjohann-PG:
@bensinjin can you proceed to test this out for production use. I created a PeaceGeeks account and invited you as Admin.

bensinjin:
@rasmus-storjohann-PG I've finished the set up (see my last commit). Currently I'm using my `authToken` in order to be able to write to Sentry. This could be changed but not a big deal right now. Other than that it looks to work well, I tested "in production" by commenting out `enableInExpoDevelopment` and publishing a release to the development channel.

App.js
```diff
@@ -1,3 +1,9 @@
 import { Application } from './lib/application';
+import Sentry from 'sentry-expo';
+
+// Uncomment to log to Sentry while developing.
+// Sentry.enableInExpoDevelopment = true;
+
+Sentry.config('https://de23d08a2e2b44a49849c045c2a6fc0c@sentry.io/256284').install();
```
rasmus-storjohann-PG: Both `enableInExpoDevelopment` and the API key should be defined in `.env`

app.json
```diff
@@ -15,6 +15,18 @@
...
-    ]
+    ],
+    "hooks": {
+      "postPublish": [
+        {
+          "file": "sentry-expo/upload-sourcemaps",
+          "config": {
+            "organization": "peacegeeks",
+            "project": "pathways",
+            "authToken": "4d99cd8514c2471c8ddde0acb3937d9ad6dd5a66add549a8aeb537f5a50dc6a7"
```
rasmus-storjohann-PG: This should definitely not go into git, it should be in `.env`

bensinjin: Yeah, of course, I've already squashed this out. Working on another solution.

App.js
bensinjin: See my final comment.

bensinjin:
@rasmus-storjohann-PG This should be ready. After digging through the sentry-expo project I was able to find that they look for an env var called `SENTRY_AUTH_TOKEN` so I've added that to our `.env.example` and it looks to work when it's set in a local `.env` file. From what I can tell the only sensitive setting. The DSN looks to be public and I did try setting it in .env and then referencing it in `App.js` this didn't appear to work and errors stopped logging so I've left it hardcoded for now. The setting for enabling is development does not need to be set and was there initially as a sort of documentation comment but I've since removed it.

rasmus-storjohann-PG:
Invalidate the current Sentry API key since it's included in this story and hence public.

PR #319, opened 2018-11-30
Feature/issue 315 make nlp metadata accessible in the app

rasmus-storjohann-PG:


rasmus-storjohann-PG:
@bensinjin for review. Just make more info available in the service list

bensinjin:
Nice.

PR #333, opened 2018-12-13
WIP Feature/issue 310 client side data migration

rasmus-storjohann-PG:


src/fixtures/buildFixtures.ts
```diff
@@ -1,5 +1,4 @@
-export { buildTasksFixture } from './newcomers_guide/tasks';
-export { buildTaxonomyFixture } from './newcomers_guide/taxonomies';
-
+export { buildTasksFixture } from './hard_coded/tasks';
```
bensinjin: What are the implications of these changes? Presumably it'll still grab newcomers fixture is they exist?

PR #343, opened 2018-12-14
Feature/issue 320 screen per question questionnaire

bensinjin:


rasmus-storjohann-PG:
@DarwishMenna Here's a PR for you to review, if you see things that don't make sense or could be improved, let us know.

src/components/questionnaire/questionnaire_component.tsx
```diff
@@ -25,94 +25,142 @@ export interface QuestionnaireActions {
...
-    </TouchableOpacity>
-));
-
-const renderContent = (section: AccordionSection): JSX.Element => (
-    section.content
-);
+    private renderNextButton(): JSX.Element {
+        const activeQuestionIndex = this.findIndexForQuestion(this.props.activeQuestion);
+        const nextQuestionIndex = activeQuestionIndex + 1;
+        const nextQuestion = this.fetchQuestionWithFallbackIndex(nextQuestionIndex, activeQuestionIndex);
```
rasmus-storjohann-PG: Picky: The code would be clearer if you could express here the situation in which you'd get the `activeQuestionIndex` value back, rather than bury that inside `fetchQuestionWithFallbackIndex()`

src/components/questionnaire/questionnaire_component.tsx
```diff
@@ -25,94 +25,142 @@ export interface QuestionnaireActions {
...
+        );
+    }
+
+    private renderPreviousButton(): JSX.Element {
+        const activeQuestionIndex = this.findIndexForQuestion(this.props.activeQuestion);
+        const previousQuestionIndex = activeQuestionIndex - 1;
+        const previousQuestion = this.fetchQuestionWithFallbackIndex(previousQuestionIndex, activeQuestionIndex);
+        const onPress = (): void => { this.props.setActiveQuestion(previousQuestion.id); this.scrollToTop(); };
+        const icon = I18nManager.isRTL ? 'arrow-forward' : 'arrow-back';
+        const text = <Trans>Back</Trans>;
```
rasmus-storjohann-PG: The renderPrevious and renderNext functions could be DRYed up a bit

src/components/main/header.tsx
```diff
@@ -25,6 +25,11 @@ type Props = HeaderProps & UiActions;
 
 export const HeaderComponent: React.StatelessComponent<Props> = (props: Props): JSX.Element => {
     const path = props.location.pathname;
+    const isOnQuestionnaireScreen = pathMatchesRoute(path, Routes.Questionnaire);
```
rasmus-storjohann-PG: This function should be in `/pathways-frontend/src/application/routing.ts` with its brethren

src/components/questionnaire/questionnaire_component.tsx
```diff
@@ -3,17 +3,19 @@ import React from 'react';
...
 import { Trans } from '@lingui/react';
 import { RouterProps } from '../../application/routing';
 import { Id, ChooseAnswerAction, SetActiveQuestionAction } from '../../stores/questionnaire';
 import { goToRouteWithoutParameter, Routes } from '../../application/routing';
 import { arrivalAdvisorGlyphLogo } from '../../application/images';
+import { EmptyComponent } from '../empty_component/empty_component';
 
 export interface QuestionnaireProps {
-    readonly questions: ReadonlyArray<selector.Question>;
+    readonly questions: ReadonlyArray<SelectorQuestion>;
```
rasmus-storjohann-PG: Either way is fine for this

src/components/questionnaire/questionnaire_component.tsx
```diff
@@ -123,44 +120,60 @@ export class QuestionnaireComponent extends React.Component<Props> {
...
+    private fetchActiveQuestionFromProps(): SelectorQuestion {
         return this.props.questions[this.findIndexForQuestion(this.props.activeQuestion)];
     }
 
-    private fetchQuestionWithFallbackIndex(indexToFetch: number, fallbackIndex: number): selector.Question {
+    private fetchQuestionRelativeToActiveQuestion(offset: number): SelectorQuestion {
+        const activeQuestionIndex = this.findIndexForQuestion(this.props.activeQuestion);
+        const targetQuestionIndex = activeQuestionIndex + offset;
+        return this.fetchQuestionWithFallbackIndex(targetQuestionIndex, activeQuestionIndex);
+    }
```
rasmus-storjohann-PG: This function is fine, but the name is long and cumbersome to use elsewhere, I would create two very simple functions
```
private fetchPreviousQuestion() { return fetchQuestionRelativeToActiveQuestion(-1); }
private fetchNextQuestion() { return fetchQuestionRelativeToActiveQuestion(1); }
```
to make the call sites easier to read.

src/components/questionnaire/questionnaire_component.tsx
```diff
@@ -15,9 +15,7 @@ import { arrivalAdvisorGlyphLogo } from '../../application/images';
 import { EmptyComponent } from '../empty_component/empty_component';
 
 export interface QuestionnaireProps {
-    readonly questions: ReadonlyArray<SelectorQuestion>;
-    readonly activeQuestion: Id;
-    readonly recommendedTaskCount: number;
+    readonly activeQuestion: SelectorQuestion;
```
rasmus-storjohann-PG: Nice

src/components/questionnaire/questionnaire_component.tsx
```diff
@@ -68,10 +66,9 @@ export class QuestionnaireComponent extends React.Component<Props> {
     }
 
     private renderProgress(): JSX.Element {
-        const activeQuestion = this.fetchActiveQuestionFromProps();
-        const currentQuestionCount = activeQuestion.number;
-        const totalQuestionCount = this.props.questions.length;
-        const barWidth = totalQuestionCount > 0 ? Math.round(currentQuestionCount / totalQuestionCount * 100) : 100;
+        const activeQuestionPosition = this.props.activeQuestion.positionInQuestionnaire;
+        const lengthOfQuestionnaire = this.props.activeQuestion.lengthOfQuestionnaire;
+        const barWidth = lengthOfQuestionnaire > 0 ? Math.round(activeQuestionPosition / lengthOfQuestionnaire * 100) : 100;
```
rasmus-storjohann-PG: Nice cleanup. I think `index` is just as clear as `activeQuestionPosition`, but it's your call

src/components/questionnaire/questionnaire_component.tsx
```diff
@@ -84,32 +81,29 @@ export class QuestionnaireComponent extends React.Component<Props> {
...
-        const isFirstQuestion = activeQuestion.number === 1;
+        const hasNextQuestion = this.props.activeQuestion.nextQuestionId !== undefined;
+        const hasPreviousQuestion = this.props.activeQuestion.previousQuestionId !== undefined;
         return (
-            <View style={{ flexDirection: 'row', justifyContent: 'space-around' }}>
-                {isFirstQuestion ? <EmptyComponent /> : this.renderPreviousButton()}
-                {isFinalQuestion ? this.renderGotoPlanButton() : this.renderNextButton()}
+            <View style={{ flexDirection: 'row', justifyContent: 'space-around', marginBottom: 10 }}>
+                {hasPreviousQuestion ? this.renderPreviousButton() : <EmptyComponent />}
+                {hasNextQuestion ? this.renderNextButton() : this.renderGotoPlanButton()}
```
rasmus-storjohann-PG: Nice

src/selectors/__tests__/questionnaire.test.ts
```diff
@@ -7,63 +8,80 @@ import { aLocale } from '../../stores/__tests__/helpers/locale_helpers';
...
         });
-    });
 
-    it('should return all the questions', () => {
-        const questionCount = anInteger();
-        const questions = new Array(questionCount).fill(0).map(() => (
-            new QuestionBuilder().withLocaleCode(locale.code)),
-        );
-        const normalizedData = new ValidStoreBuilder().withQuestions(questions).build();
+        it('position in questionnaire', () => {
```
rasmus-storjohann-PG: `it()` should take a string so that it reads like a sentence starting with "it", it should return bla bla

src/selectors/questionnaire/to_selector_question.ts
```diff
@@ -0,0 +1,21 @@
...
+    const questionIndex = R.findIndex(R.propEq('id', question.id), questionsList);
+    return {
+        id: question.id,
+        text: getLocalizedText(locale, question.text),
+        explanation: question.explanation ? getLocalizedText(locale, question.explanation) : undefined,
+        answers: toSelectorAnswerListForQuestion(locale, question, answers),
+        positionInQuestionnaire: questionIndex + 1,
+        lengthOfQuestionnaire: questionsList.length,
+        nextQuestionId: questionsList[questionIndex + 1] ? questionsList[questionIndex + 1].id : undefined,
+        previousQuestionId: questionsList[questionIndex - 1] ? questionsList[questionIndex - 1].id : undefined,
```
rasmus-storjohann-PG: `questionsList[questionIndex + 1]` cast to bool here is just a complicated way of saying `questionIndex < questionsList.length()`, and `questionsList[questionIndex - 1]` can likewise be replaced with simply `questionIndex > 0`

src/components/questionnaire/questionnaire_component.tsx
rasmus-storjohann-PG: You addressed this in a later commit

src/components/questionnaire/questionnaire_component.tsx
rasmus-storjohann-PG: You already fixed this up too

src/application/styles.ts
```diff
@@ -73,6 +82,15 @@ export const textStyles = StyleSheet.create({
         textAlign: 'left',
         color: colors.black,
     },
+    headlineH3StyleBlackCenter: {
+        fontFamily: fontFamily,
```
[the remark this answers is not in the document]
bensinjin: Right, fixed

src/components/questionnaire/questionnaire_component.tsx
```diff
@@ -25,94 +25,142 @@ export interface QuestionnaireActions {
...
-            question={question}
-            nextButtonOnPress={nextButtonOnPress}
-            isFinalQuestion={question.number === questionnaire.length} />
-    );
-};
+    private renderHeading(): JSX.Element {
+        const logoSize = Dimensions.get('screen').width / 6;
+        const closeButtonOnPress = goToRouteWithoutParameter(Routes.MyPlan, this.props.history);
+        return (
+            <View ref={(component: object): object => this.headingComponent = component as JSX.Element}>
```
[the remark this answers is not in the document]
bensinjin: Added some clarity around this. This is necessary so we can scroll the next question to the top of the page. Without it if a question is long and you go next the next question scroll position is still set at the bottom of the page from the previous question.

src/application/styles.ts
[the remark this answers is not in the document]
bensinjin: Fixed

src/components/main/header.tsx
bensinjin: So currently we have what I refer to as a "parent" and "child" set of screens and their associated functions to determine if you're on one. The naming is pretty awkward and can likely be improved but I can't think of anything better at this time, anyways, what I'm getting at here is this is an even more specific piece of functionality and I wasn't intending to write one of these for every screen. Maybe it can be justified if we find ourselves writing this code again but it feels a bit premature so I think we can leave it.

src/selectors/questionnaire/to_selector_question.ts
bensinjin: Fixed

bensinjin:
@rasmus-storjohann-PG I think I've addressed everything here

src/components/main/header.tsx
rasmus-storjohann-PG: I understand your thinking, but I think we can simplify by not worrying about parent and child screens, but just have predicate functions for testing what screen you're on. How those screens relate to other screens is a different concern. As the app grows we will encounter this again, so we can address it then.

PR #353, opened 2018-12-19
Feature/issue 328 ux changes

bensinjin:


src/components/tasks/bookmark_button_component.tsx
```diff
@@ -25,22 +43,43 @@ export const BookmarkButtonComponent: React.StatelessComponent<Props> = (props:
...
                 />
             </Button>
         </View>
     );
 };
+
+type ButtonConfiguration = {
+    readonly onPress: () => void;
+    readonly icon: string;
+};
```
rasmus-storjohann-PG: Usually we've been using `interface` for this, wonder what the difference is

src/components/tasks/bookmark_button_component.tsx
```diff
@@ -25,22 +43,43 @@ export const BookmarkButtonComponent: React.StatelessComponent<Props> = (props:
...
+    const isBookmarked = R.contains(props.taskId, props.savedTasksIdList);
+    return {
+        onPress: isBookmarked ? getRemoveBookmarkOnPress(props) : getAddBookmarkOnPress(props),
+        icon: isBookmarked ? 'bookmark' : 'bookmark-o',
+    };
+};
+
+const getAddBookmarkOnPress = (props: Props): () => AddToSavedListAction => (
+    (): AddToSavedListAction => props.addBookmark(props.taskId)
+);
```
rasmus-storjohann-PG: I've seen you start using this pattern of having functions returning functions. In some cases it was in a tricky context, so it was the right tool for the job, but it is another level of complexity and I don't think it should become the default way of doing things. I would suggest something like this is simpler:
```ts
const getButtonConfiguration = (props: Props): ButtonConfiguration => {
    const addBookmark = () => props.addBookmark(props.taskId);
    const removeBookmark = () => props.removeBookmark(props.taskId);
    const isBookmarked = R.contains(props.taskId, props.savedTasksIdList);
    return {
        onPress: isBookmarked ? addBookmark : removeBookmark,
        icon: isBookmarked ? 'bookmark' : 'bookmark-o',
    };
};
```

src/components/explore/explore_detail_component.tsx
```diff
@@ -17,6 +17,7 @@ export interface ExploreDetailProps {
 
 export interface ExploreDetailActions {
     readonly addToSavedList: (taskId: TaskId) => AddToSavedListAction;
+    readonly removeFromSavedList: (taskId: TaskId) => RemoveFromSavedListAction;
```
rasmus-storjohann-PG: A UX issue: I see the notification for adding a bookmark, but there is no corresponding notification for removing one, can we have that?

src/components/help/help_component.tsx
```diff
@@ -91,8 +91,8 @@ const ClearAppMemoryButton: React.StatelessComponent<Props> = (props: Props): JS
     const alertToClearAllUserData = (): void => {
         const alertHeading = 'Delete user data';
         const alertMessage = 'Do you want to delete all user data from this phone? This includes which ' +
-            'answers are chosen in the questionnaire, which tasks are in My Plan and which ' +
-            'tasks are marked as completed';
+            'answers are chosen in the questionnaire, which topics are in bookmarked, and which ' +
+            'topics are marked as completed';
```
rasmus-storjohann-PG: "which topics are bookmarked" (remove "in"), and remove reference to completed

src/components/tasks/bookmark_button_component.tsx
bensinjin: Primarily it means I can mix these types (action and property) and in this instance the separation is not necessary as Redux is not involved.

src/components/tasks/bookmark_button_component.tsx
bensinjin: Fixed

src/components/explore/explore_detail_component.tsx
bensinjin: Added

src/components/help/help_component.tsx
bensinjin: Fixed

src/components/explore/explore_detail_component.tsx
bensinjin: I'm reminded of what seems to be a bit of an architectural issue with the notifications in general. The process of responding to the same action in two reducers is somewhat troublesome in that you don't really want to create a notification if say the task removal operation fails for some reason. This is obviously out of scope for this story but it's something we may want to look into.

rasmus-storjohann-PG: I guess so ... generally, though, I think having more than one reducer respond to one action is one of the killer features of redux. It decouples the source of the actions from the recipients in a very elegant way, so I'm all for using that.

PR #356, opened 2018-12-20
Issue #351 Fix Arabic translations in Introduction sections

DarwishMenna:


src/fixtures/hard_coded/explore.ts
```diff
@@ -140,7 +140,7 @@ export const buildExploreFixture = (): ExploreStore => (
...
+                    'ar': 'عندما تأتي إلى كندا، يجب عليك فهم كيفية عمل القوانين الكندية لأنها قد تكون مختلفة عن القوانين في بلدك الأصلي. تتولى الحكومة الفدرالية (الكندية) مسؤلية القوانين التي تخص الدولة بأكملها. أما الحكومة الإقليمية وحكومة المقاطعة فانها تتولى مسؤلية القوانين المتعلقة بالمدن والبلدات (مثل فيكتوريا أو مترو فانكوفر). ينتخب الكنديون جميع هذه المسؤليات الحكومية والتي لها حق إصدار القوانين. أما المحاكم، فتتولى تطبيق القوانين وتكون مستقلة عن الحكومة. يوجد في كندا قانون يسمى الميثاق الكندي للحقوق والحريات (Canadian Charter of Rights and Freedoms) تعتبر هذه الحقوق والحريات مهمة جداً للكنديين وتشمل على: الحق في حرية التعبير وحرية العقيدة وحرية العيش والعمل في أي مكان في كندا والحق في المشاركة في النشاط السياسي السلمي laws-lois.justice.gc.ca/eng/Const/page-15.html',
```
rasmus-storjohann-PG: Actually I confused myself: For the introduction sections (which don't go through the import tool, it would be nice if they did): fix links like this: `[link](https://laws-lois.justice.gc.ca/eng/Const/page-15.html)`.

src/fixtures/hard_coded/explore.ts
```diff
@@ -12,15 +12,15 @@ export const buildExploreFixture = (): ExploreStore => (
...
-                    'ar': ' على دليل القادمين الجدد في بريتش كولومبيا يحتوي ستقرار في بريتش كولومبيا في إ معلومات ستساعدك على ال ولى من وصولك. أ شهر القليلة ال أال ت التي تتضمن معظم الفصول على قائمة بأسماء الوكال يمكنك الحصول منها على مزيد من المعلومات. سماء وأرقام قد تكون بعض المعلومات (مثل ال الهواتف والعناوين) قد تغيرت منذ وقت طباعة هذا الدليل. يجاد إستخدم الملحق في نهاية هذا الكتاب ل المعلومات في هذا الدليل. نترنت مكتوبة باللغة إ جميع المواقع على شبكة ال نجليزية. قد تقدم بعضها معلومات مكتوبة إ ال بلغات أخرى',
+                    'ar': 'يحتوي دليل القادمين الجدد في بريتش كولومبيا على معلومات ستساعدك على الإ ستقرار في بريتش كولومبيا في الأشهر القليلة الأولى من وصولك. تتضمن معظم الفصول على قائمة بأسماء الوكالات التي يمكن الحصول منها على مزيد من المعلومات. قد تكون بعض المعلومات (مثل الأسماء وارقام الهواتف والعناوين) قد تغيرت منذ وقت طباعة هذا الدليل. إستخدم الملحق في نهاية هذا الكتاب لايجاد المعلومات في هذا الدليل. جميع المواقع على شبكة الانترنت مكتوبة باللغة الانجليزية. قد تقدم بعضها معلومات مكتوبة بلغات أخرى. www.welcomebc.ca/newcomersguides',
```
rasmus-storjohann-PG: Change link from `www.welcomebc.ca/newcomersguides'` to `[link](https://www.welcomebc.ca/newcomersguides'`

src/fixtures/hard_coded/explore.ts
```diff
@@ -40,7 +40,7 @@ export const buildExploreFixture = (): ExploreStore => (
...
+                    'ar': 'في بريتش كولومبيا، يجب على جميع الأطفال من سن ال 5 إلى 16 الذهاب إلى المدرسة. ينقسم النظام المدرسي إلى مدارس عامة ومدارس مستقلة (خاصة). المدارس العامة هي مدارس مدعومة بشكل تام من قبل حكومة بريتش كولومبيا. ولا يتوجب دفع رسوم لكي يدرس طفلك في مدرسة عامة. أما المدارس المستقلة، فهي تحصل على تمويل جزئي من الحكومة. وعليه يتوجب على أولياء الأمور دفع رسوم لكي يدرس أطفالهم في معظم المدارس المستقلة. في بريتش كولومبيا، تقوم حكومة المقاطعة (وزارة التربية) ومجالس التربية المحلية وسلطات المدارس المستقلة بإدارة النظام المدرسي العام (من رياض الأطفال إلى الصف 12). تقوم حكومة المقاطعة .بتمويل النظام التعليمي وتسن التشريعات والقوانين والسياسات لضمان توفر المعايير الإقليمية في كافة المدارس وتلقي كافة الطلاب تعليماً ذو كفأة عالية.[www.gov.bc.ca](http://www.gov.bc.ca/bced) يمكن الأطفال أيضاً الحصول على التعليم في منازلهم.',
```
rasmus-storjohann-PG: This one contains a link of a form that we used to use, with `[www.gov.bc.ca]`, we decided to go with just `[link]` instead, so if you could fix that in Arabic and the other languages too, that would be great. As to whether we put http or https it doesn't really matter, but we prefer https.

src/fixtures/hard_coded/explore.ts
bensinjin: Currently these are not actually run through the markdown filter in the app. Whether or not they should be is likely another discussion. Naturally if we don't care to run them through it the un "marked down" format is preferable.

src/fixtures/hard_coded/explore.ts
bensinjin: Currently these are not actually run through the markdown filter in the app. Whether or not they should be is likely another discussion. Naturally if we don't care to run them through it the un "marked down" format is preferable.

bensinjin: It's probably fine. If we want to make improvements to the "Read More" functionality, specifically adding a gradient, that may not be possible on text that is using the markdown component. My general struggles with the markdown component have lead me to use it sparingly and specifically only currently when rendering Topic content.

PR #358, opened 2018-12-20
Feature/issue 346 client version appears in about

rasmus-storjohann-PG:


rasmus-storjohann-PG:
@bensinjin for review. With this PR, the version string is in four places (bad) but a unit test ensures that they all match (better). `app.json` and `package.json` are pretty unavoidable, `.env` makes it nice and easy to read the version string in during the build, and `VERSION.txt` is standard but I can make the change to not have a `VERSION.txt`. Thoughts?

PR #362, opened 2019-01-03
Feature/issue 327 google maps

bensinjin:
Some notes:
- I've done a fair amount of testing to confirm physical addresses match lat/lng values coming from the server (they're two separate fields) from what I can tell this appears to be the case. This point is important as we currently are showing the physical location on services and want to make sure when pressing "Open in map" and navigating to the provided lat/lng we're navigating to the same location. No it does not look like we can just navigate to the address as a string.
- Some services do not provided lat/lng values for these no "Open in map" button is rendered.
- Once published we will need to recreate the IPA for iOS users in order for them to open the map using Google Maps as this is a change to `app.json`, Apple Maps should work fine

### DIFF
--- a/app.json
+++ b/app.json
@@ -6,7 +6,12 @@
     "version": "1.0.0",
     "slug": "arrival_advisor",
     "ios": {
-      "bundleIdentifier": "org.peacegeeks.arrivalAdvisor"
+      "bundleIdentifier": "org.peacegeeks.arrivalAdvisor",
+      "infoPlist": {
+        "LSApplicationQueriesSchemes": [
+          "comgooglemaps"
+        ]
+      }
     },
     "android": {
       "package": "org.peacegeeks.arrivalAdvisor"
--- a/src/components/maps_application_popup/maps_application_popup_component.tsx
+++ b/src/components/maps_application_popup/maps_application_popup_component.tsx
@@ -1,7 +1,7 @@
 // tslint:disable no-class no-this no-expression-statement
 import React from 'react';
 import { View, Text, Icon, Button } from 'native-base';
-import { Popup } from 'react-native-map-link';
+import { showLocation } from 'react-native-map-link';
 import { Trans } from '@lingui/react';
 import { textStyles, colors, values, applicationStyles } from '../../application/styles';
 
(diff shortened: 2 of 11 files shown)


app.json
```diff
@@ -6,7 +6,12 @@
     "version": "1.0.0",
     "slug": "arrival_advisor",
     "ios": {
-      "bundleIdentifier": "org.peacegeeks.arrivalAdvisor"
+      "bundleIdentifier": "org.peacegeeks.arrivalAdvisor",
+      "infoPlist": {
+        "LSApplicationQueriesSchemes": [
+          "comgooglemaps"
```
rasmus-storjohann-PG: Should we allow users to use some of these other mapping apps too? https://www.npmjs.com/package/react-native-map-link

rasmus-storjohann-PG: Default iPhones don't have google maps, so we need to support apple maps at a minimum probably

src/components/maps_application_popup/maps_application_popup_component.tsx
```diff
@@ -0,0 +1,69 @@
...
+                <Popup
+                    isVisible={this.state.isVisible}
+                    onCancelPressed={this.close}
+                    onAppPressed={this.close}
+                    onBackButtonPressed={this.close}
+                    modalProps={{
+                        animationIn: 'slideInUp',
+                        onBackdropPress: this.close,
+                    }}
+                    appsWhiteList={['apple-maps', 'google-maps']}
```
rasmus-storjohann-PG: Oh OK, so I guess this takes care of apple maps

bensinjin:
Can you clarify how the dialogue should look? I've just used what the library provides although we do have control over styling. Translation of strings in the dialogue may not be possible as the library wont accept a component for string values. Regarding supporting other apps I personally don't think they're super relevant to BC for example we don't have Uber, Lyft etc...

app.json
bensinjin: Yeah from what I understand we don't include apple-maps here but whitelisting it allows us to use it.

bensinjin:
If you'd prefer we can add all the possible options for map apps and remove the whitelist?

app.json
bensinjin: This is an Apple specific setting to white list url scheme for third party apps. It doesn't look like we have to include anything about apple maps here maybe because it's iOS specific.

rasmus-storjohann-PG:
About the dialog: On my phone when I click on a location in e.g. OpenTable, google maps opens up directly without any dialog, so that is what I would expect our app would do too. I don't know if that is hard to do...

bensinjin:
@rasmus-storjohann-PG I've opted to use the more simple api (`showLocation`) that appears to work as you'd expect. I've also ensured that if no applicable applications are installed we get a sensible alert back. Please take another look

src/components/maps_application_popup/maps_application_popup_component.tsx
```diff
@@ -11,59 +11,30 @@ interface MapsApplicationPopupProps {
...
+    (props: MapsApplicationPopupProps): JSX.Element => (
+        <View>
+            <Button
+                onPress={(): Promise<void> =>
+                    showLocation({
+                        title: props.locationTitle,
+                        latitude: props.latitude,
+                        longitude: props.longitude,
+                        appsWhiteList: ['apple-maps', 'google-maps'],
+                    }).catch((): void => alert('Supported applications include Apple or Google maps.'))
```
rasmus-storjohann-PG: Can that error message be localizable?

src/components/maps_application_popup/maps_application_popup_component.tsx
```diff
@@ -1,7 +1,7 @@
 // tslint:disable no-class no-this no-expression-statement
 import React from 'react';
 import { View, Text, Icon, Button } from 'native-base';
-import { Popup } from 'react-native-map-link';
+import { showLocation } from 'react-native-map-link';
```
rasmus-storjohann-PG: So would you expect this change to alter the UX at all? Because it looks the same to me

src/components/maps_application_popup/maps_application_popup_component.tsx
```diff
@@ -11,59 +11,30 @@ interface MapsApplicationPopupProps {
...
-                    appsWhiteList={['apple-maps', 'google-maps']}
-                    options={{
-                        title: this.props.locationTitle,
-                        latitude: this.props.latitude,
-                        longitude: this.props.longitude,
+export const MapsApplicationPopupComponent: React.StatelessComponent<MapsApplicationPopupProps> =
+    (props: MapsApplicationPopupProps): JSX.Element => (
+        <View>
+            <Button
+                onPress={(): Promise<void> =>
```
rasmus-storjohann-PG: I'd pull the onPress event handler out into its own function

src/components/maps_application_popup/maps_application_popup_component.tsx
```diff
@@ -11,59 +11,30 @@ interface MapsApplicationPopupProps {
...
-    }
-
-    close(): void {
-        this.setState({ isVisible: false });
-    }
-}
+                <Text style={textStyles.button}><Trans>Open in map</Trans></Text>
+            </Button>
+        </View>
+    );
```
rasmus-storjohann-PG: Nice simplification, less than half the number of lines

rasmus-storjohann-PG: One bug: Go to Settling in -> Personal safety -> Find nearest services -> Scroll to PEACE in Williams Lake (there are several PEACE entries) -> Open in map: Some times I get three markers on the map, sometimes none.

src/components/maps_application_popup/maps_application_popup_component.tsx
bensinjin: The popup feature is a more nicely styled version of showLocation and is more suited to providing many options for map apps. The only case in which you'd notice a difference was if you were on iOS and had both Google maps and Apple maps installed and had to pick one, in which case your dialog may not be as pretty as the popup but will still work fine. In order to actually see this we need to publish another binary due to the changes in app.json, I can do that once this is in master.

bensinjin: The only case in which you'd see a difference in UX is if you had iOS and both Apple maps and Google maps installed in which case your dialog styles would differ. In order to see this we need to publish a new ipa due to the changes in app.json. I can do so once this is in master.

src/components/maps_application_popup/maps_application_popup_component.tsx
bensinjin: Fixed

src/components/maps_application_popup/maps_application_popup_component.tsx
bensinjin: I've found issues with localizing strings, we need to fix this in other places too ie. the "clear app data" dialog. I spent time poking around to no real avail (`yarn extract` is not detecting strings marked as translatable using techniques we've previously used). I've opened an issue around this here: https://github.com/pg-irc/pathways-frontend/issues/375

src/components/maps_application_popup/maps_application_popup_component.tsx
bensinjin: Nice find. So this doesn't occur on Apple maps but indeed on Google maps. It seems Google maps wants to make us aware of a closer location with the same/similar location title. I've made the location title stronger by using all parts of the address this seems to appease the Google maps gods and only show one marker.

bensinjin:
@rasmus-storjohann-PG I've addressed your comments here. Please take one last look

src/components/maps_application_popup/maps_application_popup_component.tsx
bensinjin: Fixed

Issue #376, opened 2019-01-09
Update Right Drawer Menu with About the App info


cherrielam:
See mock: [https://zpl.io/adrwJGe](https://zpl.io/adrwJGe) or open in app: zpl://screen?pid=5bc7bfa42e72260acb03aa86&sid=5c35398881e80429857c679a **Happy to review this as a team before it goes into development**

The right drawer has been split into two sections: Top being the language selection section and Bottom being the About the App section. 

Specs if you need:
![right menu drawer](https://user-images.githubusercontent.com/21990736/50868430-60035800-1364-11e9-8b9b-7b1c347e7bd3.png)

bensinjin:
Included under: https://github.com/pg-irc/pathways-frontend/issues/374

Issue #374, opened 2019-01-08
Reorganize bottom nav and top nav


cherrielam:
See mock: [Open in Web - zpl.io/beLgK1W](https://zpl.io/beLgK1W) or zpl://screen?pid=5bc7bfa42e72260acb03aa86&sid=5bc7c591c5588f18fa4caaae to open in app

**Bottom nav** should have three icons: Recommended, Learn, Bookmarks -- **EDITED JAN 8, 4:52 PM**
_Recommeded (left) is your landing/active page, Learn (middle) - should have higher importance and hierarchy than Bookmarks, Bookmarks is on the right - least priority. Onboarding should take you directly to the Recommended page if they complete the questionnaire._

**Top nav** should have a right hamburger menu icon and a left "help" icon. Left "help" icon opens to the Help page, Right menu icon opens to a menu with language and other items. 

New mock for the Right Menu to come, see: #376

bensinjin:
This will include: https://github.com/pg-irc/pathways-frontend/issues/376

PR #386, opened 2019-01-10
Feature/issue 374 reorganize nav

bensinjin:
This includes #374 & #376. At some point I'll need direction on where all the "About the app" links should navigate to. I made "Send feedback" take you to the old options page just so we don't forget it exists as button in the header is now a help button.

Note: The new "header_menu" files are copied from the old "language_switcher" (now deleted) directory and cleaned up - there isn't a much new code here as it appears.


src/components/header_menu/header_menu_button_component.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
```
rasmus-storjohann-PG: I get a warning "Warning: Failed prop type: Invalid prop `name` of value `md-arrow-back` supplied to `Icon`, expected one of ["glass","music","search"..."

src/components/main/header.tsx
```diff
@@ -32,29 +31,29 @@ export const HeaderComponent: React.StatelessComponent<Props> = (props: Props):
...
+                onPress={props.onHeaderMenuButtonPress}
                 locale={props.currentLocale}
                 textColor={colors.white}
             />;
-        return renderHeader(colors.topaz, renderArrivalAdvisorLogo(props), languageSwitcher);
+        return renderHeader(colors.topaz, renderHelpButton(props), headerMenuButton);
     }
 
     if (isOnChildScreen(path)) {
-        const languageSwitcher =
```
rasmus-storjohann-PG: This function is long. You can easily break out the onParentScreen and onChildScreen sections into their own functions

src/components/header_menu/header_menu_component.tsx
```diff
@@ -0,0 +1,188 @@
...
+                icon={<AboutIcon name='file' fontSize={20} marginRight={7}/>}
+                text={<Trans>Disclaimer</Trans>}
+                onPress={buildAboutItemOnPress(props, Routes.About)}
+            />
+            <AboutItem
+                icon={<AboutIcon name='comments' fontSize={22} marginRight={4}/>}
+                text={<Trans>Send Feedback</Trans>}
+                onPress={buildAboutItemOnPress(props, Routes.Home)}
+            />
+        </View>
```
rasmus-storjohann-PG: Are my eyes playing tricks on me or are the entries in this menu not evenly spaced vertically? Maybe because the icons are not the same height?

src/components/header_menu/header_menu_button_component.tsx
```diff
@@ -0,0 +1,21 @@
...
+export interface Props {
+    readonly locale: Locale;
+    readonly textColor: string;
+}
+
+export interface Actions {
+    readonly onPress: () => void;
+}
+
+export const HeaderMenuButtonComponent = (props: Props & Actions): JSX.Element => {
```
rasmus-storjohann-PG: This is old code you just moved, but might as well fix the naming here, `HeaderMenuButtonProps` and `HeaderMenuButtonActions`

src/components/main/header.tsx
rasmus-storjohann-PG: Also, on Android, the question mark icon looks different on the topic detail page (which is a child page) than on the other pages, it's just black with no button, elsewhere it's a round button in white/teal. Is this as designed? It could make sense anticipating it will have a photo behind it I guess

bensinjin: Fixed function length, need to see an example/how to recreate second comment as I cannot on an Android emulator

src/components/header_menu/header_menu_button_component.tsx
bensinjin: Please show how to recreate

src/components/header_menu/header_menu_button_component.tsx
bensinjin: Fixed

src/components/main/header.tsx
bensinjin: On topic details you shouldn't see the question mark in the header, you should see a back button. The styling of the back button will be addressed in another issue.

src/components/header_menu/header_menu_button_component.tsx
bensinjin: Ok - I understand the issue now although still haven't been able to recreate on an emulator - trying on a physical Android device

bensinjin: Still cannot recreate - we're going to have to debug this in person

bensinjin:
@rasmus-storjohann-PG we've fixed the bug we were seeing, this should be good to merge

PR #387, opened 2019-01-11
Feature/issue backend 289 translation pipeline

rasmus-storjohann-PG:


bin/strings.sh
```diff
@@ -0,0 +1,129 @@
...
+    done
+
+    echo "Please send files locale/*/messages.csv or locale/*/messages.po for translation"
+}
+
+
+build_all() {
+    for locale in "${locales[@]}"
+    do
+        echo "Converting CVS to PO files for ${locale} ..."
```
bensinjin: Use `CSV` instead of `CVS` (same below)

PR #388, opened 2019-01-14
#369 & #383  Update read more  & topic detail components

bensinjin:
Note: I've intentionally left out the last point on ticket #383 as we don't have clear requirements for breadcrumbs.

Issue #377, opened 2019-01-09
Small UX issues: Remove bottom nav on questionnaire, etc.


cherrielam:
Small issues:
- remove bottom nav on the questionnaire screen. Confusing because the X should close it anyway and makes page feel more busy
- on topic items, increase tap area for bookmark (easy to miss and open the topic instead)

Issue #369, opened 2019-01-04
Update the ReadMore component


rasmus-storjohann-PG:
See mock: [https://zpl.io/2pnXkR8](https://zpl.io/2pnXkR8) or in app zpl://screen?pid=5bc7bfa42e72260acb03aa86&sid=5c363c1a0b5f05bdc987351c

ACs:
* Read more button looks good
* The read more component is never smaller when expanded than collapsed
* Mocks needed
* "Read more" is 16 px below the text and 24 px above the grey line; when text is expanded the arrow will point upwards and text will say "Read Less", position with stay the same (still be 16px below the bottom of the text and 24 px above the grey line.)

More spacing reference on top of Zeplin, if need

![topic detail 1x](https://user-images.githubusercontent.com/21990736/50922359-2a5f7d00-13ff-11e9-91c9-450b5e190292.png)

cherrielam:
@bensinjin this should be ready to go! Small issue, I will create a separate issue for the other Topic Detail stuff

bensinjin:
This will include #383 as this is a small story

bensinjin:
Bug: This has been merged but we need to look into how to deal with the new bookmark button triggering the "reset" ie. re closing the readmore.

PR #390, opened 2019-01-15
Issue #377 Remove nav on questionnaire

bensinjin:
Note: The point about making the bookmark button larger was done under #369 (pr #388)

### DIFF
--- a/src/components/main/footer.tsx
+++ b/src/components/main/footer.tsx
@@ -2,7 +2,7 @@ import React from 'react';
 import { StyleProp, TextStyle } from 'react-native';
 import { Footer, FooterTab, Button, Icon } from 'native-base';
 import { History, Location } from 'history';
-import { Routes, goToRouteWithoutParameter, pathMatchesRoute, isOnStartScreen } from '../../application/routing';
+import { Routes, goToRouteWithoutParameter, pathMatchesRoute } from '../../application/routing';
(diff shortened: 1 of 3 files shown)


src/components/main/footer.tsx
```diff
@@ -13,11 +13,16 @@ export interface FooterProps {
...
     const path = props.location.pathname;
+    const isOnQuestionnaireScreen = pathMatchesRoute(path, Routes.Questionnaire);
 
     if (isOnStartScreen(path)) {
         return <EmptyComponent />;
     }
 
+    if (isOnQuestionnaireScreen) {
+        return <EmptyComponent />;
+    }
```
rasmus-storjohann-PG: Inconsistent with the code directly above

bensinjin: I've used this pattern for the header so keeping it consistent. I wasn't planning on making an `isOnXScreen` for every path as I've previously stated when this has been brought up. `isOnStartScreen` `isOnParentScreen` and `isOnChildScreen` are meant to be more general.

rasmus-storjohann-PG: I agree, I'm not a fan of `isOnStartScreen()`, so I would take that out in this PR.

bensinjin: Ah, gotcha, fixed

Issue #297, opened 2018-11-07
Figure out versioning for client and server


rasmus-storjohann-PG:
ACs: Have a plan for how we will use versions for user testing, git, server, client, etc

Update: This is mostly done, one last, optional piece is to have the server version string appear in the client About box, this would be all client side to make the network call and show the returned string on the screen.

robinpuga:
You should take a look at 	https://semver.org/

rasmus-storjohann-PG:
# Suggested rules: 
* Client and server are versioned independently.
* Client specifies the minimum server version required for proper client operation.
* We use [semantic versioning](https://semver.org/) of the form X.Y.Z
* Versioning happens at the end of each sprint.
* We bump the Z each sprint, except for sprints where no changes went in
* We bump the Y when we add features that are backwards compatible
* We bump the X when we add features that are backwards incompatible, i.e. when the client would fail to operate correctly with an older server version.
* Versions are stored as tags on the master branch in git, as well as in [__init__.py](https://stackoverflow.com/a/16055291) on the server and in [package.json](https://docs.npmjs.com/files/package.json#version) on the client.

# Questions?
When would we ever bump X on the client?

bensinjin:
@rasmus-storjohann-PG Using Release Channels will likely be related to this work: https://docs.expo.io/versions/latest/distribution/release-channels

rasmus-storjohann-PG:
@robinpuga for review, or if you don't have time, let me know

rasmus-storjohann-PG:
@bensinjin for review and feedback with specifics on release channels, avoiding accidental push of dev builds to production, etc.

robinpuga:
@rasmus-storjohann-PG This looks pretty good to me. Thanks for pulling this together. I think it would be good to include these Rules in both project's ReadMe files or wherever seems appropriate. 

I would also add a note for Hotfixes (emergency deployments outside of sprints) for both the client and server. These should also bump the Z value. 

RE: When to bump the X on the client? I would say this would be bumped when major API changes to the server happen which means it is no longer compliant with previous versions of the server. Additionally, there may be times we want to bump the X for marketing purposes because MARKETING!

rasmus-storjohann-PG:
TODO: Make a postinstall script on the client to validate that the version id in package.json, app.json and VERSION.txt all match. Something similar can be done on the server.

PR #392, opened 2019-01-16
Issue #297 Show server version in client about screen

rasmus-storjohann-PG:


rasmus-storjohann-PG:
@DarwishMenna for review since I just remembered @bensinjin is out of the office the rest of this week

src/components/about/about_component.tsx
```diff
@@ -24,9 +24,13 @@ const disclaimerTextP2 = <Trans>Users of the Arrival Advisor app do so at their
 const disclaimerParagraphs: ReadonlyArray<JSX.Element> = [disclaimerTextP1, disclaimerTextP2];
 
 const versionTitle = <Trans>Version</Trans>;
-const versionText = <Text><Trans>This is Arrival Advisor version</Trans> {VERSION}</Text>;
 
-export const AboutComponent: React.StatelessComponent = (): JSX.Element => {
+export interface Props {
```
rasmus-storjohann-PG: Oops, that should be `AboutComponentProps`.

rasmus-storjohann-PG: Fixed

src/components/about/about_component_with_server_version.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { API_URL } from 'react-native-dotenv';
+
+interface Props {
+}
+
+interface State {
+    readonly serverVersion: string;
+}
+
+export class AboutComponentWithServerVersion extends React.Component<Props, State> {
```
bensinjin: I'd call this `ServerVersionComponent` and just have it render its server version in a `<Text>` component. `AboutComponent` can render it but the fact that it does probably doesn't need to be advertised. You could even move the "version" components (client/server) to their own directory but that's your call.

src/components/about/about_component.tsx
bensinjin: I've actually been naming these (just Props and Actions) without the "Component" part to keep things a little shorter. Not sure how you feel about that or we can decide on this later.

src/components/about/about_component_with_server_version.tsx
rasmus-storjohann-PG: But your suggestion also makes a lot of sense

rasmus-storjohann-PG: But then the server version component would control its own styling, which is definitely a separate concern from making a network call ... this feels wrong after all.

bensinjin: I think technically our connected component are also container components. Maybe ideally you'd follow our patterns around sagas for this but it's probably not entirely necessary. Anyways I'm not too worried just something maybe we can standardize on better in the future. Maybe we start calling our connected components container components as the fact that we use "connect" may not be that important.

rasmus-storjohann-PG: I wrote a saga version of this, it was about 10 times as many lines of code than this version. Sagas are better, but not that much better. I don't want to be dogmatic and end up over-engineering things.

I agree that connected components are "containers", but they are clearly separated from the non-connected components, so that is a correct pattern to use. I don't like the term "container" because there's a grey zone between presentation and container. Connected is clearer, because either they contain a call to `connect()` or they don't.

bensinjin:
Ok I'm happy with this

Issue #342, opened 2018-12-13
Client: Upgrade all libraries to the latest version


rasmus-storjohann-PG:
Including new version of expo

PR #393, opened 2019-01-16
Issue #342 Upgrade all packages except react native, native base and …

rasmus-storjohann-PG:
…typescript


rasmus-storjohann-PG:
@bensinjin and @DarwishMenna for review and QA. Does the app load up OK in iDevices?

PR #401, opened 2019-01-17
Issue #369 Fix bug that caused Read more component to flicker when upd…

bensinjin:
…ating component state, Fix bug that caused crash on task services

### DIFF
--- a/src/components/expandable_content/expandable_content_component.tsx
+++ b/src/components/expandable_content/expandable_content_component.tsx
@@ -10,6 +10,7 @@ import { toggleExpandedState, ExpandableContentStates, shouldShowReadMoreButton,
 import { values } from '../../application/styles';
 
 export interface ExpandableContentProps {
+    readonly contentId?: string | number;
     readonly content: JSX.Element;
     readonly contentBackgroundColor?: string;
 }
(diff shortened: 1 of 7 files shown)


src/components/services/service_list_connected_component.tsx
```diff
@@ -10,13 +10,15 @@ import {
...
 } from './service_list_component';
+import { Routes, getMatchParamsFromPathAndRoute } from '../../application/routing';
 
 type OwnProps = {
     readonly location: Location;
 };
 
 const mapStateToProps = (store: Store, ownProps: OwnProps): ServiceListProps => {
-    const task: Task = selectCurrentTask(store, ownProps.location);
+    const matchParams = getMatchParamsFromPathAndRoute(ownProps.location.pathname, Routes.Services);
```
rasmus-storjohann-PG: The one thing I feel makes this PR hard to understand is the name of this function. It's also confusing that it needs that second parameter. However, at the call sites, I care more about what it does than how it does it, so the name should focus on that. How about `getParametersFromPath()`. Maybe it would also help if the first argument was passed in like this.
```
getParametersFromPath(ownProps.location, Routes.Services);
```
it's less to read and makes sense that it takes the location, at the call site we don't care that it actually needs the pathname from within the location.

src/components/expandable_content/expandable_content_component.tsx
```diff
@@ -10,6 +10,7 @@ import { toggleExpandedState, ExpandableContentStates, shouldShowReadMoreButton,
 import { values } from '../../application/styles';
 
 export interface ExpandableContentProps {
+    readonly contentId?: string | number;
```
rasmus-storjohann-PG: Broad types like this is a code smell. Can it just be `string` and not optional?

src/components/services/service_list_connected_component.tsx
bensinjin: Ah yeah, I like that, fixed

bensinjin:
@rasmus-storjohann-PG I think this is ready

src/components/expandable_content/expandable_content_component.tsx
rasmus-storjohann-PG: Yeah, taking the number out helps, it makes it look less like "just throw any old value in there".

PR #403, opened 2019-01-17
Feature/issue 367 photos for content

DarwishMenna:


bin/generateTopicsImages.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+path='src/application/topicsImages.ts'
+
+echo "export const topicsImages: any = {" >> $path
+for f in assets/images/topics/*.png; do
+    filename=$(basename "$f" | cut -d. -f1)
```
rasmus-storjohann-PG: Well done figuring that out, it's tricky. If you look at `man basename` it looks like there might be ways to get rid of the file extension without `cut`.

src/components/tasks/task_detail_heading_component.tsx
```diff
@@ -1,10 +1,16 @@
...
 import { View } from 'native-base';
-import { arrivalAdvisorGlyphLogo } from '../../application/images';
 import { colors } from '../../application/styles';
+import { topicsImages } from '../../application/topicsImages';
+export interface TaskDetailHeadingComponent {
+    readonly taskId: string;
+}
 
-export const TaskDetailHeadingComponent: React.StatelessComponent = (): JSX.Element => {
+type Props = TaskDetailHeadingComponent;
```
rasmus-storjohann-PG: Hmmm, this doesn't look quite right. Have a look at our existing components and how the props interfaces are defined and named. `BookmarkButtonComponent`, `HeaderMenuComponent` and `HelpComponent` are all examples to follow there.

src/components/tasks/task_detail_component.tsx
```diff
@@ -25,18 +25,26 @@ export interface TaskDetailActions {
```
rasmus-storjohann-PG: One more thing: We don't "merge master", we prefer to rebase onto master, it makes the git history a lot easier to follow. We can talk about that in person next time your in.

Also, in our commit messages, we put a # in front of the issue numbers, it makes it clickable so we can easily go from a commit to the corresponding issue, which is super helpful.

bin/generateTopicsImages.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+echo "export const topicsImages: any = {"
+for f in $1; do
+    filename=$(basename $f .png)
+    echo "'$filename' : require('../../assets/images/topics/$filename.png'),"
+done
+echo "};"
```
rasmus-storjohann-PG: Getting there, still some outstanding issues: I want to call the script with a list of files, the output will include just the first file. Like this (see `man touch` for what that does):
```
$ touch assets/images/topics/foo.png
$ touch assets/images/topics/bar.png
$ touch assets/images/topics/baz.png
$ ./bin/generateTopicsImages.sh assets/images/topics/*png
export const topicsImages: any = {
'bar' : require('../../assets/images/topics/bar.png'),
};
```
Have a look at using `$@` or `"$@"`

rasmus-storjohann-PG: Secondly, the hard-coded path to the images in the script is a problem. If I call the script with images in some other location, I get the wrong result. The file is in `/tasks/`, but the output points to the file in `/topics/`:
```ts
$ touch assets/images/tasks/foo.png
$ ./bin/generateTopicsImages.sh assets/images/tasks/foo.png 
export const topicsImages: any = {
'foo' : require('../../assets/images/topics/foo.png'),
};
```

rasmus-storjohann-PG: The name `topicsImages` that you export from the generated ts code is also very specific, and makes it hard to use the script for other kinds of images, such as task images. You could pass the name to use as a command line argument to the script. However, a simper solution is to just export it as the generic name `images`. Where the file is imported, we can do this
```ts
import { images as topicsImages } from '../../application/topicsImages';
import { images as taskImages } from '../../application/taskImages';
```

src/components/tasks/task_detail_heading_component.tsx
```diff
@@ -3,8 +3,15 @@ import { Image, Dimensions } from 'react-native';
 import { View } from 'native-base';
 import { arrivalAdvisorGlyphLogo } from '../../application/images';
 import { colors } from '../../application/styles';
+import { images as topicsImages } from '../../application/topicsImages';
```
bensinjin: `topicImages` is a nicer name

PR #421, opened 2019-01-25
Feature/issue 367 photos for content

DarwishMenna:


bin/generateTopicsImages.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+echo "export const images: any = {"
+for f in $@; do
+    filename=$(basename $f .png)
+    echo -e "\t'$filename' : require('$f'),"
```
rasmus-storjohann-PG: The -e is causing a problem on my linux system. Not sure why it is needed. Other than that, this PR is ready. I suggest squishing the commits into one though, just because of the amount of churn on this. One more thing, rename the script from `generateTopicsImages.sh` to `generateImagesFile.sh`.

PR #422, opened 2019-01-25
Issue pg-irc/pathways-backend#300 Replace packages with known vulnera…

rasmus-storjohann-PG:
…bilities, found using retire.js


package.json
```diff
@@ -79,8 +79,12 @@
     "@types/ramda": "^0.25.38",
     "expo": "28.0.0",
     "expo-analytics": "^1.0.7",
+    "extend": "^3.0.2",
     "history": "^4.7.2",
     "intl": "^1.2.5",
+    "jquery": "^1.12.4",
```
bensinjin: Not sure I follow, what are these new libraries for? jQuery? surely we don't need that...

rasmus-storjohann-PG: It's a dependency of one of our dependencies, but by default it pulls in a version of jquery that has a vulnerability, this version does not.

Issue #424, opened 2019-01-29
Investigate solutions to Content + Flatlists


bensinjin:
Currently if you have a `Content` component with a nested `FlatList` and that list uses infinite scroll (`onEndReached`) the outer content component will load all data without waiting for the scroll conditions of the flatlist to be met. This can be recreated by Nesting our `TaskListComponent` inside a `Content` component.  This is a known issue but we should try and define how we'd expect this to work and then remove the need to pass `loadOnDemand` to our `TaskListComponent`.

bensinjin:
By using the `FlatList`'s `HeaderComponent` prop we've resolved this problem.

PR #427, opened 2019-01-29
Feature/issue 396 app performance improvements

rasmus-storjohann-PG:
From @bensinjin 

A lot of the work on this story was digging and guessing. My main take away is the point I make about our NewlyRecommendedTasks component. Here's an overview of what I've done:

Replaced areas where we'd loop over articles with <Flatlist/>. This speeds up page loads. Up to 20 + frames faster in some cases.

Allowed large article lists to be loaded on demand using the infinite scroll technique. This greatly increases performance on our recommended and Bookmarked article pages. Previously if these lists were large you'd see very choppy animations when opening the menu or a complete freeze for around 1.5s when trying to navigate to a particular task.

Note: This is not yet in place for article and explore detail screens due to issues around having nested scroll views on the same screen. I hope to address this in the future (Issue stubbed in here: #424).

Only render NewlyRecommendedTasks when navigating away from the questionnaire. This is actually a pretty noticeable one, the questionnaire should feel a lot faster now and this has actually improved performance throughout the app. In general we need to think twice about rendering connected components on every page as it turns out it can be a hefty performance hit as they recalculate their values every time the store changes even if they're not visible on the screen.

Animate the read more component. I put this in place to ease in the read more content so there isn't that "jank" we sometimes see. I'm not super stoked on the end result but in some cases it's better. Feel free to give me your thoughts.


bensinjin:
@rasmus-storjohann-PG Thx for cleaning this up, appreciate your contribution. Let me know if there are more changes that need to done here.

PR #433, opened 2019-01-30
Feature/issue 396 app performance improvements

bensinjin:
Original PR here: https://github.com/pg-irc/pathways-frontend/pull/425

Here are some tests demonstrating performance improvements. Generally speaking all our lists will scale much better now as they have a constant load time and data in lists is loaded on demand. This means there is more time in a frame for other work to be done.   Loosely speaking the more frames we lose in a test the more we wait.

Max frame rate is 60. 
Before is on the `master` branch
After is on the `feature/Issue_396_app_performance_improvements` branch

**Navigate to recommended screen with 60 topics:**

Before: 20fps
After:  42fps (this number should not grow)

_22 frames gained_

**Answer question in questionnaire:**

Before: 52fps  
After: 58fps 

_6 frames gained_

 **Answer another question in questionnaire:**

Before: 35fps, drops dramatically while answering questions in succession (as low as 5fps)
After: 58fps

_23 - 50 frames gained_ 

**Navigate to next question in questionnaire:**

Before: 47fps, drops dramatically while answering questions in succession (as low as 5fps)
After: 57fps

_10 - 50 frames gained_

**Bookmark topic in topic list with 60 tasks:**

Before: 15fps
After: 40fps

_25 frames gained_

   **Navigate to settling in explore detail screen (picked due to its somewhat large number of topics):**

Before: 32fps
After:  48fps 

_8 frames gained_


bensinjin:
@rasmus-storjohann-PG  This PR grew, I did however need to find a solution to: https://github.com/pg-irc/pathways-frontend/issues/424, which I've done in my later commits (I can explain if need b), in order to properly move ahead with the styling stories coming.

Issue #396, opened 2019-01-16
Improve app performance


rasmus-storjohann-PG:
Understand why the app is slow on some phones.
Confirm that the bundles used are production bundles.

PR #435, opened 2019-01-31
Issue #396 For operations that remove or add items to task lists simp…

bensinjin:
…ly load the entire data set

### DIFF
--- a/src/components/tasks/task_list_component.tsx
+++ b/src/components/tasks/task_list_component.tsx
@@ -73,14 +73,29 @@ export class TaskListComponent extends React.PureComponent<Props, State> {
     }
 
     private getFreshState(): State {
-        const sections = R.splitEvery(this.numberOfItemsPerSection, this.props.tasks);
-        const sectionIndex = this.state ? sections.length - 1 : 0;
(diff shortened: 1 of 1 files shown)


src/components/tasks/task_list_component.tsx
```diff
@@ -77,6 +72,17 @@ export class TaskListComponent extends React.PureComponent<Props, State> {
         );
     }
 
+    private getFreshState(): State {
+        const sections = R.splitEvery(this.numberOfItemsPerSection, this.props.tasks);
+        const sectionIndex = this.state ? sections.length - 1 : 0;
+        const data = this.state ? this.props.tasks : sections[0];
```
rasmus-storjohann-PG: Why this dependence on `this.state`? Normally this kind of code goes
```
const data = this.state ? this.state.tasks : sections[0];
```
but you don't reference state after the `?`. You can make it clearer with a well named local variable assigned to `this.state`, or perhaps an early return on `this.state` not being null.

bensinjin: Good points, I've cleaned this up.

Issue #408, opened 2019-01-22
Service listing: "null" appears for missing postal code


rasmus-storjohann-PG:
Better At Home Program - North Shore is missing the postal code, which appears as "null" in the app.

ACs:
* When the address is missing the postal code, it should not show anything for the postal code, i.e. render as `""`, not `"null"`.

PR #440, opened 2019-02-01
Issue #408 Fix postal code value when it's null

DarwishMenna:


DarwishMenna:
@rasmus-storjohann-PG

Issue #450, opened 2019-02-05
UX Polish


rasmus-storjohann-PG:
- [x] Topic details page: String "Recommended for me" with the star icon is not styled
- [x] Home page call to action block, should have an image, see Zepplin
- [x] Service listing page, header should not be sticky 
- [x] Server side: Add a hash to everything in the NCG content, see [here](https://github.com/pg-irc/pathways-backend/issues/304)
- [ ] When newly recommended topics appears, hide the nav bar 
- [x] Add map pin icon to the "Find related services near me" button
- [ ] About, disclaimer: Don't breadcrumb these, they have an X button which takes you back to the
most recent non-about non-disclaimer page
- [x] Help page: Find a settlement worker near me doesn't work
- [x] Go to the questionnaire, answer some questions, clear all data from the app, go back to the questionnaire, observe that the app remembers the current question you were on, expected is that the current question resets to the first question in the questionnaire. (We have other flows than delete all data that also ought to reset the current question)
- [x] Add new AA icon

PR #452, opened 2019-02-05
Feature/issue 450 ux polish

bensinjin:
This currently excludes two items from the list here: https://github.com/pg-irc/pathways-frontend/issues/450. These are broken out into new stories: #454 #455


app.json
```diff
@@ -2,7 +2,7 @@
   "expo": {
     "sdkVersion": "28.0.0",
     "name": "Arrival Advisor",
-    "icon": "./assets/images/aa_logoglyph.png",
+    "icon": "./assets/images/aa_icon.png",
```
rasmus-storjohann-PG: At some point we may need different icons for apple and android, not yet...

src/components/tasks/task_detail_content_component.tsx
```diff
@@ -91,7 +91,7 @@ const RecommendedComponent = (props: Props): JSX.Element => {
                 }}
             />
             <Text style={[textStyles.paragraphStyle, { color: colors.greyishBrown, fontSize: 14, fontStyle: 'italic' }]}>
-                <Trans>Recommended for you</Trans>
+                <Trans>Recommended for me</Trans>
```
rasmus-storjohann-PG: Didn't expect this change

src/components/help/help_component.tsx
```diff
@@ -86,8 +89,10 @@ export const HelpComponent: React.StatelessComponent<Props> = (props: Props): JS
     </Content>
 );
 
-const ContactSettlementWorkerButton: React.StatelessComponent = (): JSX.Element => (
-    <Button style={[applicationStyles.tealButton, applicationStyles.boxShadowBelow, { alignSelf: 'center' }]}>
+const ContactSettlementWorkerButton: React.StatelessComponent<Props> = (props: Props): JSX.Element => (
+    <Button
+        onPress={goToRouteWithParameter(Routes.Services, 'contact-workers-at-your-local-settlement-agency', props.history)}
```
rasmus-storjohann-PG: The id of the topic to open is pretty important and needs to be easier to find that this. Can you put it in a variable, it's just a magic string like it is....

src/stores/questionnaire/reduce_valid_store.ts
```diff
@@ -36,11 +37,7 @@ export const reduceValidStore = (store: ValidQuestionnaireStore, action?: Questi
             return new LoadingQuestionnaireStore(store);
 
         case constants.CLEAR_ALL_USER_DATA:
-            return new ValidQuestionnaireStore({
-                ...store,
-                oldAnswers: {},
-                answers: R.map((answer: Answer): Answer => ({ ...answer, isChosen: false }), store.answers),
-            });
+            return new ValidQuestionnaireStore(buildDefaultStore());
```
rasmus-storjohann-PG: This doesn't seem quite right ... it's better than it was, but not taken far enough. So it should definitely do the same thing in `/src/stores/tasks/reduce_valid_store.ts`, but arguably all stores should do it, so it could be done in `/src/stores/index.ts` ... something to think about, don't want to rush into that right now I don't think.

app.json
bensinjin: Ah, good point

src/components/tasks/task_detail_content_component.tsx
cherrielam: Just putting it here for reference: Should be "you" if not a button, "me" if it is a button

bensinjin: Fixed

bensinjin:
@rasmus-storjohann-PG this should be ready to go, I've added back an image I thought we no longer were using but noticed it missing when viewing through TestFlight.

PR #461, opened 2019-02-08
Feature/issue 453 upgrade Expo

bensinjin:
This was a bit of a slog but I got it eventually. What I found helped was referring to expo's scaffolding command `expo init`. By comparing a fresh expo project with ours I was able to more easily track down what dependencies we really need to get updated in order for things to work. I suggest this approach if we wait this long again but I hope we don't.

Along with the Expo upgrade I've done a bit of clean up on packages that were redundant (ie `jest-expo` brings in `jest` so we shouldn't need to also have `jest` as a dependency). A few little things have changed for example we no longer use `exp` and `react-native-scripts` is no longer a thing we just use `expo`, also I had to add a type that no longer existed in the same place (`LocationData`), I've left a note about this. 

To test this fully remove your node_modules and then run `yarn`.


bensinjin:
Taking this down temporarily - I want to try and fix our peer dependency warnings in this PR.

package.json
```diff
@@ -3,94 +3,76 @@
...
-  "main": "./node_modules/react-native-scripts/build/bin/crna-entry.js",
+  "main": "./node_modules/expo/AppEntry.js",
   "scripts": {
     "postinstall": "bin/postinstall.sh",
     "lint": "tslint src/**/*.ts src/**/*.tsx",
     "build": "bin/buildFixtures.sh && tsc",
-    "test": "yarn run jest --maxWorkers=4",
-    "cbt": "yarn run clean && yarn run build && yarn run test",
-    "bt": "yarn run build && yarn run test",
+    "test": "node ./node_modules/jest/bin/jest.js",
```
rasmus-storjohann-PG: You'll want to keep the `--maxWorkers=4`, it make the tests run faster, surprisingly. Spinning up some huge number of worker processes is time consuming. This has been an intermittent problem on Travis as well, causing timeouts on tests.

package.json
```diff
@@ -3,94 +3,76 @@
...
-      "ts",
-      "tsx",
-      "js",
-      "jsx",
-      "json",
-      "node"
+    "testPathIgnorePatterns": [
+      "/node_modules/",
+      "/lib/",
+      "/helpers/"
```
rasmus-storjohann-PG: We don't have a helpers folder, remove

src/@types/location-data.d.ts
```diff
@@ -0,0 +1,14 @@
...
+    coords: {
+        latitude: number;
+        longitude: number;
+        altitude: number;
+        accuracy: number;
+        heading: number;
+        speed: number;
+    };
+    timestamp: number;
+}
```
rasmus-storjohann-PG: I found I had to do this
```ts
diff --git a/src/sagas/services.ts b/src/sagas/services.ts
index 5d5942c..c0d7836 100644
--- a/src/sagas/services.ts
+++ b/src/sagas/services.ts
@@ -39,7 +39,8 @@ const getLocationIfPermittedAsync = async (): Promise<LocationData | undefined>
         if (permissions.status !== 'granted') {
             return undefined;
         }
-        return await Location.getCurrentPositionAsync({ enableHighAccuracy: false });
+        const lowAccuracy = 1;
+        return await Location.getCurrentPositionAsync({ accuracy: lowAccuracy });
     } catch (error) {
         return undefined;
     }
```
We may have to play with it to determine if 1 is the right value to use, valid values are 0, 1, 2, 3, see https://docs.expo.io/versions/latest/sdk/location/

package.json
```diff
@@ -3,94 +3,76 @@
...
     "extract-strings": "lingui extract",
     "extract-strings-clean": "lingui extract --clean",
     "build-strings": "lingui compile --strict",
     "clean-strings": "rimraf 'locale/_build/**/*.json'",
-    "start": "react-native-scripts start",
-    "eject": "react-native-scripts eject",
-    "android": "react-native-scripts android",
-    "ios": "react-native-scripts ios"
+    "start": "expo start",
+    "eject": "expo eject",
```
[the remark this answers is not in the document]
bensinjin: Fixed

package.json
```diff
@@ -3,94 +3,76 @@
...
-    "moduleFileExtensions": [
-      "ts",
-      "tsx",
-      "js",
-      "jsx",
-      "json",
-      "node"
+    "testPathIgnorePatterns": [
+      "/node_modules/",
+      "/lib/",
```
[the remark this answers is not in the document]
bensinjin: Fixed

package.json
bensinjin: That's a deprecated option. It's suggested you do not set it and use the default.

package.json
bensinjin: We do have helpers in `__tests__` directories for example `stores/__tests/helpers`.

src/@types/location-data.d.ts
bensinjin: Added

rasmus-storjohann-PG:
@bensinjin I'm going to wait a bit before merging this, since the expo changes are breaking the lingui export/import, so I want to get the UX strings in first ... Obviously we also want to make lingui work with the new expo version, soon.

bensinjin:
Np @rasmus-storjohann-PG - I think we need to update lingui with this work. I need to pop over to another project but I think we can also do the lingui update in this PR when we're ready.

package.json
bensinjin: This also means we can likely get our app to work without using our `yarn build --watch` command at some point...

Issue #418, opened 2019-01-24
Add website to Service Listing


cherrielam:
Add website URL to Service Listings screen

## ACs:
* For services that have a URL defined, the service list entry for the service should have a clickable link to the URL with text `[link]`. See also https://github.com/pg-irc/pathways-frontend/issues/429

### Server side:
* Add `Organization.website` and `Organization.email` to the `Service` part of the `ServiceAtLocation` response
* Remove `description` from `Location` part of the `ServiceAtLocation` response, since it usually duplicates the description in the `Service` part of the same response and makes the response almost twice as large as it needs to be.

### Client side:
* Add the `website` and `email` fields to the `Service` DTO returned from the API call
* Show the website and email address in the service listing

cherrielam:
Not sure who wants to pick this up: @rasmus-storjohann-PG @ArtiBPG I'll let you assign and I'll leave this in backlog for now

PR #468, opened 2019-02-13
Issue #418 Add service website and email to service listing

DarwishMenna:


src/components/services/service_list_item_component.tsx
```diff
@@ -58,6 +60,20 @@ const renderPhoneNumbers = (phoneNumbers: ReadonlyArray<PhoneNumber>) => (
     </View>), phoneNumbers))
 );
 
+const renderWebsite = (website: string): JSX.Element => (
```
bensinjin: `renderWebsite` and `renderEmail` aren't really different enough to constitute two functions. You could just have `renderServiceField(label: JSX.Element, content: JSX.Element)` or something.

rasmus-storjohann-PG: I'm fine either way with that.

DarwishMenna: Done

src/components/services/service_list_item_component.tsx
```diff
@@ -58,6 +61,20 @@ const renderPhoneNumbers = (phoneNumbers: ReadonlyArray<PhoneNumber>) => (
...
+    <Text>
+        <Text style={textStyles.paragraphBoldBlackLeft}><Trans>Web:</Trans> </Text>
+        <Link href={website} text={website} style={textStyles.paragraphStyle} />
+    </Text>
+);
+
+const renderEmail = (email: string): JSX.Element => (
+    <Text>
+        <Text style={textStyles.paragraphBoldBlackLeft}><Trans>Email:</Trans> </Text>
+        <Link href={"mailto:" + email} text={email} style={textStyles.paragraphStyle} />
```
rasmus-storjohann-PG: You should get tslint running, it warns about deviations from our coding style. Here it would tell you to use `'` instead of `"`.

PR #474, opened 2019-02-14
Feature/issue 465 import translated strings

rasmus-storjohann-PG:


rasmus-storjohann-PG:
@bensinjin can you have a look?

Issue #420, opened 2019-01-25
Google analytics behavioural data is all zero


rasmus-storjohann-PG:
ACs:
* Understand why we are not seeing event data in google analytics, and fix this
* Enable anonymization of the google data  using [anonymizelp](https://support.google.com/analytics/answer/2763052?hl=en)
* Add robin to the google analytics account
* Confirm that our Sentry account isn't going to expire
* Send language id with each event, if its simple. 
* Anonymize ip addresses
* Report fine grained on which pages hit with the topic id or service id.

PR #483, opened 2019-02-20
Issue #420 Google Analytics

bensinjin:
Live data is coming through using the PG GA ArrivalAdvisor tracking code. Data in the Behaviour section should show up in the next 24 hours.


bensinjin:
I see Sentry was mentioned in this issue too. The account we're using looks to still be active.

PR #490, opened 2019-02-22
Issue #420 Google Analytics

bensinjin:
Note: This only currently tracks user language for screen views. We'll have to address events on a per event basis.

### DIFF
--- a/src/components/main/main_component.tsx
+++ b/src/components/main/main_component.tsx
@@ -11,30 +11,36 @@ import { Location, Action } from 'history';
 import { RouteChangedAction } from '../../stores/router_actions';
 import { Locale } from '../../locale';
 
-interface HeaderMenuState {
-    readonly isHeaderMenuShowing: boolean;
-}
-
-export type MainComponentProps = FooterProps & RouterProps & { readonly locale: Locale };
+export type MainComponentProps = MainProps & FooterProps & RouterProps;
 
 export interface MainComponentActions {
-    readonly routeChanged: (location: Location, locale: Locale) => RouteChangedAction;
+    readonly sendAnalyticsData: (location: Location, locale: Locale) => RouteChangedAction;
+}
+
+interface MainProps {
+    readonly locale: Locale;
 }
 
 type Props = MainComponentProps & MainComponentActions;
 
-export class MainComponent extends React.Component<Props> {
+interface State {
+    readonly isHeaderMenuShowing: boolean;
+}
 
-    readonly state: HeaderMenuState = {
-        isHeaderMenuShowing: false,
-    };
+export class MainComponent extends React.Component<Props, State> {
 
     constructor(props: Props) {
         super(props);
+        this.state = {
+            isHeaderMenuShowing: false,
+        };
         this.openDrawer = this.openDrawer.bind(this);
         this.closeDrawer = this.closeDrawer.bind(this);
-        props.history.listen((location: Location, _: Action) =>
-            props.routeChanged(location, this.props.locale),
+    }
+
+    componentDidMount(): void {
+        this.props.history.listen((location: Location, _: Action) =>
+            this.props.sendAnalyticsData(location, this.props.locale),
         );
     }
 
--- a/src/components/main/main_component.tsx
+++ b/src/components/main/main_component.tsx
@@ -9,15 +9,16 @@ import { HeaderMenuConnectedComponent } from '../header_menu/header_menu_connect
 import { RouterProps } from '../../application/routing';
 import { Location, Action } from 'history';
 import { RouteChangedAction } from '../../stores/router_actions';
+import { Locale } from '../../locale';
 
 interface HeaderMenuState {
     readonly isHeaderMenuShowing: boolean;
 }
 
-export type MainComponentProps = FooterProps & RouterProps;
+export type MainComponentProps = FooterProps & RouterProps & { readonly locale: Locale };
 
 export interface MainComponentActions {
-    readonly routeChanged: (location: Location, action: Action) => RouteChangedAction;
+    readonly routeChanged: (location: Location, locale: Locale) => RouteChangedAction;
 }
 
 type Props = MainComponentProps & MainComponentActions;
@@ -32,7 +33,9 @@ export class MainComponent extends React.Component<Props> {
         super(props);
         this.openDrawer = this.openDrawer.bind(this);
         this.closeDrawer = this.closeDrawer.bind(this);
-        props.history.listen(props.routeChanged);
+        props.history.listen((location: Location, _: Action) =>
+            props.routeChanged(location, this.props.locale),
+        );
     }
 
     render(): JSX.Element {
--- a/src/sagas/analytics/send_analytics_data.ts
+++ b/src/sagas/analytics/send_analytics_data.ts
@@ -1,6 +1,6 @@
 // tslint:disable:no-expression-statement
 
-import { Analytics as ExpoAnalytics, PageHit, Event } from 'expo-analytics';
+import { Analytics as ExpoAnalytics, ScreenHit, Event } from 'expo-analytics';
 import { call, CallEffect, PutEffect, put } from 'redux-saga/effects';
 import { GOOGLE_ANALYTICS_TRACKING_ID, DEBUG_GOOGLE_ANALYTICS } from 'react-native-dotenv';
 import { AnalyticsAsync } from './actions';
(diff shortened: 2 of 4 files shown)


src/sagas/analytics/send_analytics_data.ts
```diff
@@ -22,8 +22,7 @@ export function* sendAnalyticsData(action: WatchedAction): AnalyticsActions {
 
 async function sendAnalyticsDataAsync(action: WatchedAction): Promise<void> {
     const debug = DEBUG_GOOGLE_ANALYTICS === 'true';
-    // tslint:disable-next-line:no-null-keyword
-    const analytics = new ExpoAnalytics(GOOGLE_ANALYTICS_TRACKING_ID, null, { debug });
+    const analytics = new ExpoAnalytics(GOOGLE_ANALYTICS_TRACKING_ID, { aip: 1 }, { debug });
```
rasmus-storjohann-PG: It's not clear without googling what `aip` does. I would create a local variable with value `1` and a name that makes the meaning clear.

rasmus-storjohann-PG:
@DarwishMenna and @tomy-pg please have a look at this PR to watch and participate in our code review process.

src/components/main/main_component.tsx
```diff
@@ -9,15 +9,16 @@ import { HeaderMenuConnectedComponent } from '../header_menu/header_menu_connect
...
 import { Location, Action } from 'history';
 import { RouteChangedAction } from '../../stores/router_actions';
+import { Locale } from '../../locale';
 
 interface HeaderMenuState {
     readonly isHeaderMenuShowing: boolean;
 }
 
-export type MainComponentProps = FooterProps & RouterProps;
+export type MainComponentProps = FooterProps & RouterProps & { readonly locale: Locale };
```
rasmus-storjohann-PG: That's a little hacky I think...

src/components/main/main_component.tsx
```diff
@@ -9,15 +9,16 @@ import { HeaderMenuConnectedComponent } from '../header_menu/header_menu_connect
...
 
 export interface MainComponentActions {
-    readonly routeChanged: (location: Location, action: Action) => RouteChangedAction;
+    readonly routeChanged: (location: Location, locale: Locale) => RouteChangedAction;
```
rasmus-storjohann-PG: Why would a function called `routeChanged` care about the locale? I think a better name for the function will make this easier to follow, probably by naming it according to what it does (analytics) rather than why it's called (route changed)

src/sagas/analytics/send_analytics_data.ts
```diff
@@ -20,33 +20,26 @@ export function* sendAnalyticsData(action: WatchedAction): AnalyticsActions {
...
+// tslint:disable-next-line:no-any
+const setupExpoAnalyticsForAction = (action: WatchedAction): any => {
+    const additionalParameters = action.type === constants.ROUTE_CHANGED ?
+        { aip: 1, ul: action.payload.locale.code } : { aip: 1 };
+    const debug = { debug: DEBUG_GOOGLE_ANALYTICS === 'true' };
+    return new ExpoAnalytics(
+        GOOGLE_ANALYTICS_TRACKING_ID,
+        additionalParameters,
+        debug,
+    );
```
rasmus-storjohann-PG: This can be tidied up a bit: This function is definitely doing more than one thing, and it's dealing with a pretty complex API, so there is a lot for a reader to take in. I would pull out the logic building `additionalParameters` and `debug` into their own functions. The logic for `additionalParameters` could well end up being more than one function. Then you would have something like
```
setupExpoAnalyticsForAction = () => {
   const value = helperFunction();
   const value2 = helperFunction2();
   return new ExpoAnalytics(...);
}
```
which is better, but that would also mix high level code (calling the helper functions) with low level code (the details of how to build a `ExpoAnalytics`. If you pull the `ExpoAnalytics` out into its own function too, then you get a nice high level function that just shows *what* do to but nothing about *how* to do it, separation of concerns, single responsibility, all good stuff.
```
setupExpoAnalyticsForAction = () => {
   const value = helperFunction();
   const value2 = helperFunction2();
   return helperFunction3(value, value2);
}
```
Finally, I also think the name `setupExpoAnalyticsForAction` could be improved. Setup to me implies that the function might have some side effect, but in fact this function does not, it just creates and returns this analytics object, so I would name it something with `create`. You will run into a problem I often see of what to name `helperfunction3`, since it also creates an object of the same type, with as a low level function. But you can figure something out there.

src/sagas/analytics/send_analytics_data.ts
```diff
@@ -20,33 +20,26 @@ export function* sendAnalyticsData(action: WatchedAction): AnalyticsActions {
 }
 
 async function sendAnalyticsDataAsync(action: WatchedAction): Promise<void> {
```
rasmus-storjohann-PG: It's hard to review this change without looking at the resulting code in Visual Code, which is how I'm reviewing it. I think I would structure `sendAnalyticsDataAsync` differently. Currently it mixes high level concerns (which action types should be sent to analytics) with low level detail (creating an `analytics` object and sending `ScreenHit` and other kinds of event to it. What if you changed `sendAnalyticsDataAsync()` so that all it does is decide what actions should be sent to analytics and whether to use `ScreenHit` or `Event`. Everything else, including the fact that you have to create this `analytics` object, should be taken care of by helper functions.

src/sagas/analytics/send_analytics_data.ts
bensinjin: I opted to not create a local variable here as I felt to be consistent we should then create a local for `ul` which TS will only let you do in after a statement that confirms the action type. It was all getting a bit messy so I took another approach. I've decided to create an interface for these to give them a little more meaning and add some type safety when creating these parameters. I've also added a link to the reference docs in the comments which I feel works fine in this instance.

src/components/main/main_component.tsx
bensinjin: Cleaned up.

src/components/main/main_component.tsx
bensinjin: Fixed.

src/sagas/analytics/send_analytics_data.ts
bensinjin: Nice little write up here. The more this stuff is pointed out the easier it gets to notice particularly the Single Level of Abstraction principal which you point out here. I think I've split this up better. Let me know.

src/sagas/analytics/send_analytics_data.ts
bensinjin: `sendAnalyticsDataAsync` now only sends "hits" on objects created with helpers.

src/sagas/analytics/send_analytics_data.ts
```diff
@@ -20,26 +20,67 @@ export function* sendAnalyticsData(action: WatchedAction): AnalyticsActions {
...
-        analytics.hit(new ScreenHit(action.payload.location.pathname));
+        analytics.hit(createScreenHit(action.payload.location.pathname));
     }
     if (action.type === constants.CHOOSE_ANSWER) {
-        analytics.hit(new Event('Questionnaire', 'ChooseAnswer', 'no answer id'));
+        analytics.hit(createAnswerChosenEvent());
     }
     if (action.type === constants.ADD_TO_SAVED_TASKS) {
-        analytics.hit(new Event('Bookmarks', 'SaveTopic', 'no topic id'));
+        analytics.hit(createBookmarkAddedEvent());
```
rasmus-storjohann-PG: What if you moved `const analytics = buildExpoAnalytics(action);` into `createScreenHit()` and `createAnswerChosenEvent()`. It's maybe not completely DRY, but it will make it clearer in those two functions how this API works, and it means that this high level function won't need to know how the API works, i.e. this function will no longer have a local variable `analytics`.

src/sagas/analytics/send_analytics_data.ts
```diff
@@ -20,26 +20,67 @@ export function* sendAnalyticsData(action: WatchedAction): AnalyticsActions {
...
+    readonly ul?: string;
+}
+
+const createGoogleAnalyticsParameters = (action: WatchedAction): GoogleAnalyticsParameters => {
+    if (action.type === constants.ROUTE_CHANGED) {
+        return {
+            aip: 1,
+            ul: action.payload.locale.code,
+        };
+    }
```
rasmus-storjohann-PG: This is better. But this is a very good example of how you can use local variables to make this function clearer, and thereby remove the comment above. `const shouldReportCurrentLanguage = action.type === constants.ROUTE_CHANGED` would explain what is special about that kind of action, same with `const currentLanguage = action.payload.locale.code`, and `const enableIpAnonymization = 1`.

src/sagas/analytics/send_analytics_data.ts
bensinjin: I see your point but by doing what you suggest here we have to do another check in `createScreenHit` for the action type in order for TS to be happy. You'd end up with something like:
```
const createScreenHit = (action): any => {
    // Next line necessary to access payload on action 
    If (action.type === 'screenhit') {
         // Create analytics and call hit()
    }
}
```
I was hoping to reduce the number checks on `action.type` or at least try and keep them close in proximity (my original code did a better job of this). Do you still want me to proceed?

bensinjin: Fixed, see my last commit .. sorry for the large diffs :/

src/sagas/analytics/send_analytics_data.ts
bensinjin: My most recent pass on this work breaks the creation of these into their own helpers with better names.

src/sagas/analytics/send_analytics_data.ts
bensinjin: I'm working on smaller diffs .. I'm starting to realize how reorganizing makes it harder to review and will try and keep those changes in their own commits.

Issue #458, opened 2019-02-06
Simple fix to the recommendation/bookmark confusion + add PG logo


rasmus-storjohann-PG:
Suggestion: Change the icons for either or both these concepts. (see Cherrie's comment below)

Also, Add PeaceGeeks logo to the landing screen. See zeplin: https://zpl.io/adPzkPR

cherrielam:
Use the user-check FontAwesome5 icon. See zeplin: https://zpl.io/beLgK1W (ignore the stars at the bottom)

PR #491, opened 2019-02-22
Issue #458 Update recommended topic logos and add Peacegeeks logo to welcome screen

bensinjin:
Text "Created by" will need a translation.


src/components/recommended_topics/recommended_topics_component.tsx
```diff
@@ -18,6 +18,10 @@ export interface RecommendedTopicsProps {
     readonly recommendedTopics: ReadonlyArray<TaskListItem>;
 }
 
+export const recommendedIconName = 'check-decagram';
+
+export const recommendedIconType = 'MaterialCommunityIcons';
```
rasmus-storjohann-PG: These two constants are very closely associated with each other since the value of the icon name has no meaning without the name of the icon set it is part of, how about
```ts
export const recommendedIcon = {
    name = 'xyz',
    iconSet = 'xyz',
};
```

src/components/recommended_topics/recommended_topics_component.tsx
```diff
@@ -56,17 +60,28 @@ const TaskListHeaderComponent = (props: Props): JSX.Element => (
...
+                </Text>
+                <Icon
+                    style={{
+                        fontSize: 18,
+                        color: colors.lightTeal,
+                    }}
+                    name={recommendedIconName}
+                    type={recommendedIconType}
+                />
+            </View>
```
rasmus-storjohann-PG: I haven't worked enough with JSX code to know for sure, but I'm thinking that if this code was broken up into smaller single responsibility helper components, then these kinds of UX changes would become simpler, and easier to write and review. For one thing, you would see less changes to white space in the code just because the code indentation level changes. I'd like to try to adhere more closely to SRP in JSX code when we start on the next version, just to see if we like it.

src/components/tasks/task_list_item_component.tsx
```diff
@@ -63,5 +64,13 @@ export const TaskListItemComponent: React.StatelessComponent<Props> = (props: Pr
...
-    <Icon style={{ fontSize: 14, color: colors.sunYellow, marginRight: 5 }} name='star' type='FontAwesome' />
+    <Icon
+        style={{
+            fontSize: 18,
+            color: colors.lightTeal,
+            marginRight: 5,
+        }}
+        name={recommendedIconName}
+        type={recommendedIconType}
+    />
```
rasmus-storjohann-PG: This is repeated a few places, seems like there should be a `RecommendedIcon` component, possibly taking the size as a property.

rasmus-storjohann-PG:
Commit message clearly does two things. Not a bit deal.

Splash screen is beautiful with the Peacegeeks logo! Still not sure about the background image, but that is easy to fix.

src/components/recommended_topics/recommended_topics_component.tsx
bensinjin: Now irrelevant with new component.

src/components/recommended_topics/recommended_topics_component.tsx
bensinjin: I've tried to come up with "rules for how we decided when to break up a JSX component" it's tricky because almost every new component you write has a somewhat unique structure (or should if you're at least trying to be DRY). It'd be great if we could say "You're breaking SRP here because of X" but I tend to just try and make sure the functions don't get too long and then break them up in reasonable places. I'm all for trying to develop a strategy around this but have personally found just when you think you have something you find a special case.

bensinjin: Along these lines I think we should maybe try and adopt the `index` file approach for our component directories. This will help us distinguish between things that are useful to the rest of the app and files just housing child components.

src/components/tasks/task_list_item_component.tsx
bensinjin: New component added.

Issue #496, opened 2019-02-25
Move Android release forward


bensinjin:
Version 1.0.6 has been approved for the App Store now it's time to move forward the Android release. Log in to the console to view these links.

ACs (follow this order):

- ~~Bump versionCode in `app.json` to 5 so we can release 1.0.6 on Android~~
- ~~Generate a new Android binary with `expo ba --release-channel production`, we'll use production as we've done for the iOS binary~~
- ~~Upload Android screens from Cherrie to https://play.google.com/apps/publish/?account=7459548275266608645#MarketListingPlace:p=org.peacegeeks.ArrivalAdvisor&appid=4976189706172329622~~
- ~~Upload new binary here https://play.google.com/apps/publish/?account=7459548275266608645#PrepareReleasePlace:p=org.peacegeeks.ArrivalAdvisor&appid=4976189706172329622&releaseTrackId=4699293937028760980&releaseId=4705242303616121478 This is a new closed track I've created for invite only testers for version 1.0.6~~
- ~~Upload Feature Graphic here https://play.google.com/apps/publish/?account=7459548275266608645#MarketListingPlace:p=org.peacegeeks.ArrivalAdvisor&appid=4976189706172329622~~
- ~~Upload Arabic blurb here: https://play.google.com/apps/publish/?account=7459548275266608645#PrepareReleasePlace:p=org.peacegeeks.ArrivalAdvisor&appid=4976189706172329622&releaseTrackId=4700648648132288376&releaseId=4704499860150164016~~
- ~~Once Arabic blurb is in, on the same page, submit for production~~

After we're happy and ready to release to the store we'll simply promote the closed track build to production. For Android (unlike iOS) once you release to production and the app is approved they automatically make it available through the store so you need to make sure you're ready. We likely won't need to make use of the open tracks at this point but can do so if invite only is no longer sufficient.

cherrielam:
Play store screenshots: https://drive.google.com/open?id=1_WXmFw4dtCcwrky3hXkJlKAySBuSRAPc

cherrielam:
Play store banner @bensinjin @rasmus-storjohann-PG https://drive.google.com/open?id=1abmCqt15I2nKoXWO0JXIYFH0cfJbBs23 let me know if this doesn't work

bensinjin:
This has been released - waiting for it to update in the Play Store

PR #497, opened 2019-02-25
Issue #496 Bump Android version

bensinjin:
Please read #496


rasmus-storjohann-PG:
I think bumping this value should be done at the same time as the front-end version is bumped. It's OK for this time, though.

bensinjin:
Yeah I agree, just wasn’t aware of it until yesterday

Issue #473, opened 2019-02-13
Create "Things to do right away" section


cherrielam:
Create a "Things to do right away" header in the home page - 'Recommended for you' section. Items under that header should be list items that are tagged with "explore:rightaway" (This already exists in the content"

The "Things to do right away" does not appear in the Learn page.

This is dependent on the Headers issue #368 . Mocks to come and icon for header is needed

rasmus-storjohann-PG:
There are currently 9 sections in the "Learn all about B.C." page. There needs to be a 10th, called "Things to do right away". It should be the first section on that page. It should be tied to the taxonomy term "explore:rightaway".

DarwishMenna:
What's the description and icon to be used here? @cherrielam

cherrielam:
Icon:
check-square-o - https://expo.github.io/vector-icons/

Colour: #ED604B

Text: <None> --> is this possible?

PR #510, opened 2019-03-01
Issue #473 Add rightaway section in Learn about BC page

DarwishMenna:


DarwishMenna:
@bensinjin

bensinjin:
@DarwishMenna I was able to see your new icon. I'm not sure entirely why your environment is acting up .. Here's mine: 
screen shot 2019-02-28 at 4 27 56 pm

src/fixtures/hard_coded/explore.ts
```diff
@@ -25,8 +38,8 @@ export const buildExploreFixture = (): ExploreStore => (
                     'taxonomyTermId': 'settling_in',
                 }],
             },
-            's2': {
-                id: 's2',
+            's3': {
+                id: 's3',
```
rasmus-storjohann-PG: It's a problem that creating one new entry triggers updates to all these ids. Can you change the ids to be more meaningful, e.g. `'education'`. Just confirm that making that change does not affect the order in which the sections appear in the Learn page in the app, I don't believe it will. We are relying on JS objects retaining the ordering of their elements, which is a little dodgy, a better solution would be to have an ordering property on each element with numerical values defining the sort order ... but let's not do that now.

rasmus-storjohann-PG: One more thing, can you save the next developer from being confused as you were by the two taxonomy.ts files: Add a comment to the top of taxonomy.ts and tasks.ts under hardcoded/ indicating that these files are only included in the build if the tasks.ts file in newcomers_guide is *not* present.

DarwishMenna: Done

PR #531, opened 2019-03-06
Feature/issue 529 update french strings

rasmus-storjohann-PG:


rasmus-storjohann-PG:
@bensinjin for review

Issue #368, opened 2019-01-04
Section headings in topic lists


rasmus-storjohann-PG:
ACs:
* Topic in topic lists are sorted by their sections
* Each section has a heading with section name and icon 
* Needs mocks

cherrielam:
See bottom section of this mock: https://zpl.io/beLgK1W

cherrielam:
Pls discuss who is responsible for this! @rasmus-storjohann-PG @DarwishMenna @bensinjin

PR #542, opened 2019-03-08
Issue #368 topic headings

bensinjin:


rasmus-storjohann-PG:
@tomy-pg I'd like you to get going on reviewing other's code, can you please take a first pass at this PR and flag any problems (code style, unclear names, functions that do too much). I haven't looked at this PR yet, but I will after you're done to see if you missed anything ;-)

tomy-pg:
Looks good @bensinjin . Definitely learned a lot from reviewing this.

cherrielam:
@bensinjin Headings should only be on the home screen, not on related topics.

bensinjin:
> @bensinjin Headings should only be on the home screen, not on related topics.

Yep, I made that change this morning

cherrielam:
@bensinjin cool cool I forgot to comment on the PR as you asked.

rasmus-storjohann-PG:
So the "Things to do right away" section needs to be at the top of the list.

src/components/tasks/build_topic_list_items_with_headings.tsx
```diff
@@ -0,0 +1,32 @@
+import * as R from 'ramda';
+import { TaskListItem } from '../../selectors/tasks/task_list_item';
+import { TopicListHeading } from './topic_list_heading_component';
+
+export type TopicListItemOrHeading = TaskListItem | TopicListHeading;
```
rasmus-storjohann-PG: I'm not a great fan of type names with "or" in them. Better to use to the type name to highlight what the various kinds of items have in common. So here, there is a list of items, each item is either a heading or a topic. So how about
```
export type ListItem = TopicListItem | HeadingListItem;
```

src/components/tasks/build_topic_list_items_with_headings.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+export const buildTopicsListItemsWithHeadings = (topics: ReadonlyArray<TaskListItem>): ReadonlyArray<TopicListItemOrHeading> => (
+    flattenGroupedTopics(groupTopicsByExploreSection(topics))
+);
+
+type GroupedTaskListItems  = { readonly [heading: string]: ReadonlyArray<TaskListItem> };
```
rasmus-storjohann-PG: I think this is much more readable:
```ts
type GroupedTaskListItems  = { 
    readonly [heading: string]: ReadonlyArray<TaskListItem> 
};
```

src/components/tasks/build_topic_list_items_with_headings.tsx
```diff
@@ -0,0 +1,32 @@
...
+                id: heading,
+                heading: heading,
+                icon: groupedTopics[heading][0].exploreSection.icon,
+            },
+            ...groupedTopics[heading],
+        ],
+        [],
+        R.keys(groupedTopics),
+    )
+);
```
rasmus-storjohann-PG: SRP applies here. Long functions should be avoided, and particularly in functional code like this, "long" means 6-7lines or something. This function is doing a flattening with the reduce, as well as some other data manipulation that is not very clear. Breaking that out into a separate function, perhaps used with `R.map()`, will likely make this code easier to follow.

src/components/tasks/build_topic_list_items_with_headings.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+export const buildTopicsListItemsWithHeadings = (topics: ReadonlyArray<TaskListItem>): ReadonlyArray<TopicListItemOrHeading> => (
+    flattenGroupedTopics(groupTopicsByExploreSection(topics))
+);
+
+type GroupedTaskListItems  = { readonly [heading: string]: ReadonlyArray<TaskListItem> };
+
+const groupTopicsByExploreSection = (topics: ReadonlyArray<TaskListItem>): GroupedTaskListItems => (
+    R.groupBy((topic: TaskListItem) => topic.exploreSection.name, topics)
+);
```
rasmus-storjohann-PG: So there are two ways of pulling the grouping out into its own function, either what you've done here, or you can have something like 
```ts
const topicName = (topic: TaskListItem) => topic.exploreSection.name;
```
and then at the call site you would get
```ts
flattenGroupedTopics(R.groupBy(topicName, topics))
```
Not sure which you would prefer.

src/components/tasks/build_topic_list_items_with_headings.tsx
rasmus-storjohann-PG: Actually, now I look at it closer, it's actually very nice. But the way the code is broken up, the big picture of what you're doing is getting a bit lost. So your new accumulator is the old accumulator, plus a heading item, plus the group of topics belonging under that heading. You can take advantage of the fact that the first argument to `R.reduce()` is a function, so you can have local variables in there that you can give names that makes the flow much easier to follow. And you can pull the low level details of how a heading item is actually constructed into a helper function, making this function consistently high level and clearer. @tomy-pg and @DarwishMenna for visibility.

rasmus-storjohann-PG: Do you need the separate `id` and `heading` properties in the heading item?

src/components/tasks/topic_list_heading_component.tsx
```diff
@@ -0,0 +1,30 @@
...
+export interface TopicListHeading {
+    readonly id: string;
+    readonly heading: string;
+    readonly icon: string;
+}
+
+export interface TopicListHeadingProps {
+    readonly heading: string;
+    readonly icon: string;
+}
```
rasmus-storjohann-PG: Do you need both of these types?

src/selectors/tasks/select_saved_tasks.ts
```diff
@@ -1,12 +1,9 @@
...
 import { pickSavedTaskIds } from './pick_saved_task_ids';
 
-export const selectSavedTasks = (appStore: Store): ReadonlyArray<TaskListItem> => {
-    const savedTasksList = pickSavedTaskIds(appStore);
-    return savedTasksList.map((taskId: store.Id) => {
-        return selectTaskAsListItem(appStore, taskId);
-    });
-};
+export const selectSavedTasks = (appStore: Store): ReadonlyArray<TaskListItem> => (
+    R.map(buildSelectorTaskListItem(appStore), pickSavedTaskIds(appStore))
```
rasmus-storjohann-PG: Nice improvement

bensinjin:
@bensinjin TODO: Recreate this PR against develop

rasmus-storjohann-PG:
Closing PR against master, will open a new PR against develop

Issue #516, opened 2019-03-04
Make button accomodate two lines of text


ludivinegh:
**Describe the bug**
The French translation of these two buttons is quite long, and the layout is off. Consider shortening the translations or reducing the font size, to prevent the strings from being so close to the buttons' edges.

**App version**
1.0.7

**To Reproduce**
1. Go to help screen.
2. See error for both buttons.

**Expected behavior**
Shorter buttons or a smaller font would look better

**Screenshots**
http://glbl.me/n83

ACs:
* Update relevant long strings to include `\n` for newline
* Change UX logic to detect the `\n` in the string and make a button that is large enough to acomodate two lines of text

rasmus-storjohann-PG:
So this looks a little better, is this issue considered fixed?

![image](https://user-images.githubusercontent.com/32307781/54044185-30fe3c80-4183-11e9-9014-a64526c2ee35.png)

rasmus-storjohann-PG:
For easier access, this is the screenshot with the broken handling of two lines of text on the button, this is bad:

![image](https://user-images.githubusercontent.com/32307781/54044252-5c812700-4183-11e9-8a2f-0c64bbeb1617.png)

rasmus-storjohann-PG:
Just need to fix one remaining issue with this button

PR #544, opened 2019-03-08
Issue #516 Implement multiline button

bensinjin:
The solution to this problem is to allow text to wrap in a button component. I've created a multi line button component to be used for buttons with more text. The default nativebase button does not support this.

### DIFF
--- a/src/components/help/help_component.tsx
+++ b/src/components/help/help_component.tsx
@@ -9,6 +9,7 @@ import { mapWithIndex } from '../../application/map_with_index';
 import { ClearAllUserDataAction } from '../../stores/questionnaire/actions';
 import { openURL } from '../link/link';
 import { goToRouteWithParameter, Routes } from '../../application/routing';
+import { MultiLineButtonComponent } from '../mutiline_button/multiline_button_component';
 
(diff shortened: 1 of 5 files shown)


src/components/help/help_component.tsx
```diff
@@ -92,13 +93,11 @@ export const HelpComponent: React.StatelessComponent<Props> = (props: Props): JS
...
 const ContactSettlementWorkerButton: React.StatelessComponent<Props> = (props: Props): JSX.Element => (
-    <Button
-        onPress={goToRouteWithParameter(Routes.Services, settlementWorkerTaskID, props.history)}
-        style={[applicationStyles.tealButton, applicationStyles.boxShadowBelow, { alignSelf: 'center' }]}>
+    <MultiLineButtonComponent onPress={goToRouteWithParameter(Routes.Services, settlementWorkerTaskID, props.history)}>
         <Text style={textStyles.button}>
             <Trans>Find a settlement agency near me</Trans>
         </Text>
-    </Button>
+    </MultiLineButtonComponent>
```
rasmus-storjohann-PG: Button looks good, should use it also for the "Find services near me" on the task page (it's two lines in French), and "Delete all data from the app" (ditto). Have a poke through the app to see if there are other ones.

Issue #530, opened 2019-03-06
Make andriod version code track version string


rasmus-storjohann-PG:
Write a unit test that fails unless the version code in `app.json`: `expo/android/versionCode` is equal to `10000*x+100*y+z` given a version string of `x.y.z`. E.g. if the version string is 1.2.3, the version code should be `10203`.

ACs:
* Unit test fails if the above condition is not met.
* The test should be in src/application/__tests__/version.test.ts

PR #549, opened 2019-03-11
Issue #530 Add test to make android version code track version

DarwishMenna:
@rasmus-storjohann-PG


src/application/__tests__/version.test.ts
```diff
@@ -39,4 +39,15 @@ describe('version id in VERSION.txt', () => {
...
         expect(json['version']).toBe(version);
     });
+
+    it('is tracked by android version code', () => {
+        const filename = 'app.json';
+        const content = readFileSync(filename).toString();
+        const json = JSON.parse(content);
+        const androidVersionCode = json['expo']['android']['versionCode'];
+        const versionSplit = version.toString().split('.');
+        const code = 10000*Number(versionSplit[0])+100*Number(versionSplit[1])+Number(versionSplit[2]);
```
rasmus-storjohann-PG: You're using lots of named variables up top to make the code easy to follow, but not for the different version componetns. Variables `major`, `minor` and `patch` would be helpful.

Issue #448, opened 2019-02-05
Make links in disclaimer clickable, all locales


rasmus-storjohann-PG:
See DisclaimerComponent, line 15, but the change has to happen in the corresponding localized strings too.

rasmus-storjohann-PG:
According to @bensinjin the about page is done but not the disclaimer page. The commits fixing the about page should not have been tagged with this story, since this story makes no reference to the about page. The commit message also doesn't make mention of about or disclaimer page.

The work to fix the disclaimer page is still outstanding.

PR #596, opened 2019-04-09
Issue #448 Fix disclaimer links

bensinjin:
These issues sneak by us because the links work when running the app through Expo but don’t when using a binary. The secret here is to make sure your link components are not rendered in `<Trans>` components (just as we do in the About component). I've confirmed this is fixed by building a binary only for use on the simulator which produces the following in Arabic:
image

So you see the links are fixed (open your app to see how they're currently broken) but will need translations.


rasmus-storjohann-PG:
I will merge this PR but leave the story open so we can address the localization issue under this same story.

Issue #594, opened 2019-04-09
Back button with side menu open should close side menu


rasmus-storjohann-PG:
Repro on Android:
- Start the app
- Navigate through a couple of pages
- Open the hamburger menu
- Hit back button
- Observe that the hamburger menu remains open and the navigation backwards through breadcrumbs happens behind it
- Expected is that back button closes the hamburger menu

bensinjin:
@rasmus-storjohann-PG do you mean the Android back button?

bensinjin:
I see what you mean, when the Android back button is pressed, digging...

rasmus-storjohann-PG:
This issue came back, happens on staging now, 1.38.0

rasmus-storjohann-PG:
Fixed

PR #597, opened 2019-04-09
Issue #594 Implement component to handle hardware back events

bensinjin:


rasmus-storjohann-PG:
@bensinjin I added one more commit with a minor tweak, looks OK?

src/components/main/main_component.tsx
```diff
@@ -83,9 +83,10 @@ export class MainComponent extends React.Component<Props, State> {
     }
 
     onHardwareBackButtonPress(): boolean {
-        goBack(this.props.history);
         if (this.state.isHeaderMenuShowing) {
             this.closeDrawer();
+        } else {
+            goBack(this.props.history);
```
bensinjin: Don't think this will work. This basically says "Only go back if the menu is open". We need to always go back when Android's hardware back is pressed and only close the menu when it's open.

bensinjin: This component replaces the "Hack" react router had to make the hardware back button work. They don't work together thus the need to create our own slightly modified version

Issue #323, opened 2018-12-10
Onboarding: Simple 1 page with text


ArtiBPG:
Create a simple onboarding page with text. The onboarding screen should inform the user about:
- what the app does
- what they can expect from the app
- heads up about the questionnaire

- get strings ready for translation ASAP, can do this before we write any code

cherrielam:
![1 1x](https://user-images.githubusercontent.com/21990736/49903706-f2bfde00-fe1c-11e8-967a-327d69df4838.png)
Note: image at the top is a placeholder. Thoughts? Ready to dev?
Back button should go back to the start page (select language), only way to go forward is to "Start personalization" or "Skip personalization"

cherrielam:
If ready to dev, the Zeplin is here: https://app.zeplin.io/project/5bc7bfa42e72260acb03aa86/screen/5c1191cc1c6890243d39d5f7

rasmus-storjohann-PG:
@cherrielam is this mock ready to go or are you making a new one?

rasmus-storjohann-PG:
@cherrielam can you add the text to the translation spreadsheet? This will go in early in V2

PR #611, opened 2019-04-17
Issue #323 Add onboarding store and component

DarwishMenna:


src/selectors/onboarding/pick_show_onboarding.ts
```diff
@@ -0,0 +1,5 @@
+import { Store } from '../../stores';
+
+export const pickShowOnboarding = (appStore: Store): boolean => {
+    return appStore.onboardingInStore.showOnboarding;
+};
```
rasmus-storjohann-PG: The naming of selector functions (select, pick, get, build, etc.) is not always clear cut, but the intention is that select is for top level selector functions that are called from code outside the `selector` subfolder in the source tree, and all the others are helper functions that are private to the `selector` subfolder code. This is not enforced so we need to just maintain that coding standard as consistent as possible. Since this function is called from a connected component outside this folder, it should be called a name with "select", so `selectShowOnboarding`.

src/stores/onboarding/index.ts
```diff
@@ -0,0 +1,30 @@
+import * as constants from '../../application/constants';
+import { OnboardingAction } from './actions';
+
+export interface OnboardingStore {
+    readonly showOnboarding: boolean;
+}
+
+export function buildDefaultStore(): OnboardingStore {
+    return {showOnboarding: false};
+}
```
rasmus-storjohann-PG: Change to `()` form without `return`.

src/components/onboarding/onboarding_component.tsx
```diff
@@ -0,0 +1,186 @@
...
+            resizeMode={'contain'}
+            style={{
+                width: arrivalAdvisorLogoSize,
+                height: arrivalAdvisorLogoSize,
+                marginBottom: 20,
+                marginVertical: 15,
+                alignSelf: 'center',
+            }}
+        />
+        <Text style={textStyles.headlineH2StyleBlackCenter}><Trans>{props.title}</Trans></Text>
```
rasmus-storjohann-PG: Does this `<Trans>` element actually work?

src/sagas/user_data.ts
```diff
@@ -16,6 +16,7 @@ export function* watchUserStateChangesToSaveUserData(): IterableIterator<ForkEff
             constants.REMOVE_FROM_SAVED_TOPICS,
             constants.TOGGLE_IS_TOPIC_COMPLETED,
             constants.CLEAR_ALL_USER_DATA,
+            constants.SET_ONBOARDING,
```
rasmus-storjohann-PG: I've been struggling with this name, it's ambiguous, set it to what? It's also setting it to false, which is more "clear" than "set", so how about CLEAR_SHOW_ONBOARDING_FLAG?

src/components/onboarding/onboarding_component.tsx
rasmus-storjohann-PG: Yes, it does

Issue #618, opened 2019-04-24
Set up Google Analytics events tracking 


cherrielam:
Set up the following events to track performance within the app using Google Analytics. 

We would like to track...

#### 1) Which services are accessed the most frequently

**Services Listing:**
- *#* clicks on website URL of a service (we want to see which service's website URL is clicked the most)
- *#* clicks on phone of a service (we want to see which service's phone # is clicked the most)
- *#* clicks on email of a service (we want to see which service's email is clicked the most)
- *#* clicks on service "Open in Maps" button (we want to see which service is opened in Maps the most)
- *#* We should be able to look up a service, any service and see if the phone, email and/or map links were ever clicked

*Q: is there a way to see all of the above in aggregate? (we want to see which services are clicked (any of URL, email, phone, Open in Maps) the most)* 

- *#* visits to the Services Listing page, specific to a topic 
- *#* clicks on a Services Listing page, specific to a topic
(we want to know how the number of page visits of Services Listing page compares with the number of clicks once the user hits that page - eg. when they get to the Services Listing for "Settlement workers" do they access a service or do they leave? If they leave it would indicate that they didn't find the Services Listing useful)
- *#* We should be able to look up a topic, any topic and see if the service listing page was ever visited

**Topic Detail:**

- *#* clicks on the "Find Services" button, specific to a topic (we want to know which topics have the most clicks on "Find Services", eg. which topics are people most interested in accessing services for?)

#### 2) Which topics people find most useful

**Bookmarks:**

- *#* saved bookmarks, specific to a topic (we want to know which topics are being bookmarked the most)
Note: we should separate from the un-saving of bookmarks (eg. if someone clicks the bookmark icon to save and then clicks again to un-save, the saved bookmark event should not be tracked 2 times. Instead, we should track that the "save" event happened 1 time, and the "unsave" event happened 1 time. If possible it might also be nice in the future to track the time between "save" and "unsave" events to remove the ones that were accidental saves (immediate unsaves)
- *#* We want to know how many topics on average have been bookmarked per user. Since we are not tracking IPs, this could just be an average based on total number of active users.

**Topic Detail:**

- time spent on a topic (we want to know which topics people spend the most time on)

#### 3) Overall usefulness of the app 
(many of these are probably already tracked?)

- *#* saved bookmarks per user  
- time spent on the app
- reopens of the app (returning users)
- *#* of topics read per user (might use # of screens per user as a proxy)
- *#* downloads
- uninstall rate
[...]

PR #619, opened 2019-04-26
Feature/issue 618 more Google Analytics

bensinjin:
The hope here is to strengthen our analytics. I think I've been able to include all the data we require to properly understand how the app is being used based on #618. In summation what I've done is:

- Improved existing questionnaire and bookmark custom events by including the associated question and topic ids
- Established a structure for our custom link and button press events for Google Analytics
- Attached custom events to link clicks and button presses on topics and services (these are the only places I've added events currently)

**How these events look in Google Analytics**
Screen Shot 2019-04-26 at 4 26 33 PM
(For a more feature rich experience wait 24 hours and navigate to Behaviour > Events > Screens)

Along with the data outlined here we also send the current page url for a particular event. All this data is filterable on the Google Analytics side so we should be able to generate the reports we require. The other metrics (how long people stay on a screen, # of users etc ...) outlined in the story should already be available to us through other places in our Google Analytics account. Downloads and uninstalls should be available through our App Store Connect and Google Play accounts.


bensinjin:
Still testing this after the rebase.

bensinjin:
@rasmus-storjohann-PG I've tested this works after the rebase and it's looking good. Please give this a final review.

bensinjin:
@rasmus-storjohann-PG a review on this would be good.

bensinjin:
I'll rebase on master - stand by.

Issue #616, opened 2019-04-23
Related Topics don’t get refreshed after one is selected 


jparlina:
Tested on both v 1.0.13 and v 1.0.14 (staging) for iOS: after user selects one of the 3 related topics listed at the bottom screen, the 3 related topics don’t get refreshed. So the topic selected still shows up at the bottom screen as one of the 3 related topics.

In the screenshots example:
- user selects the topic “Registering your child in a public school”, then the screen shows the article under that topic, 
- user then clicks on one of the related topics listed at the bottom of the screen e.g.  “Independent (private) schools”
- screen now shows the article under “Independent (private) schools” with 3 related topics at the bottom of the screen. However, these related topics are still from the previous topic “Registering your child in a public school”, instead of those related to the current topic shown. 

![E3A5A3DB-FD9F-4F30-8621-B01A48364AC6](https://user-images.githubusercontent.com/48343061/56613055-b3737c00-65ca-11e9-9ebb-f29c94bfc350.jpeg)

![2E9BD670-64DE-456B-99D6-20046EEBBE05](https://user-images.githubusercontent.com/48343061/56613054-b3737c00-65ca-11e9-98a8-653968b736a4.jpeg)

rasmus-storjohann-PG:
Solution in the PR is not a good approach. Better to have a selector that can pull out the relevant topics from the store and pass these to the component, so we don't use component state for this. Unless I'm missing something that motivated the use of state here.

jparlina:
Tested on staging app v 1.0.18 - fixed.
Now related topics get refreshed when one of the related topics is selected (see screen recording).

https://drive.google.com/file/d/1yvmF9ZN-NxtXXN6gZd75smMqSdfHRyz4/view?usp=sharing

PR #622, opened 2019-05-10
Issue #616 refresh related topics

tomy-pg:


src/components/topics/task_list_component.tsx
```diff
@@ -51,6 +51,7 @@ export class TaskListComponent extends React.PureComponent<Props, State> {
     componentDidUpdate(previousProps: Props): void {
         if (previousProps.headerContentIdentifier !== this.props.headerContentIdentifier) {
             this.flatListRef.scrollToOffset({ animated: false, offset: 0 });
+            this.setState(this.getFreshState());
```
rasmus-storjohann-PG: This feels wrong ... let's talk. I think perhaps it's time to think about React hooks!

PR #639, opened 2019-05-22
Issue #356 Force google maps to use the lat/long for the location on …

rasmus-storjohann-PG:
…the map, not the address


rasmus-storjohann-PG:
Need to check if this actually fixes the problem, not so easy to reproduce

PR #655, opened 2019-06-06
Feature/issue 644 arabic service listing

rasmus-storjohann-PG:


src/application/styles.ts
```diff
@@ -167,6 +176,13 @@ export const textStyles = StyleSheet.create({
         letterSpacing,
         ...getNormalFontStylesForOS(),
     },
+    alwaysLeftParagraphStyle: {
+        fontSize: 16,
+        lineHeight: 21,
+        color: colors.black,
+        letterSpacing,
+        ...getAlwaysLeftTextAlign(),
```
rasmus-storjohann-PG: Add `getNormalFontStylesForOS()` call here. Unfortunate code duplication

Issue #662, opened 2019-06-08
Staging V 1.0.19 - Change title to "Terms of Use" on app and website for consistency.  


jparlina:
Missing all translations for "Terms of Use" (highlighted in red at the bottom of screenshot below) 

Add the string to the next round of translations as:
"Terms of Use"

ACs:
* App: Change app string to "Terms of Use" in English, and translate to all languages, change the url to https://arrivaladvisor.ca/terms-of-use/
* App: Translate other new strings as extracted by jslingui
* Website: Change the url to https://arrivaladvisor.ca/terms-of-use/, change the large heading on that page from "Terms & Conditions" to "Terms of Use"

![IMG_0955.JPG](https://images.zenhubusercontent.com/5ceeccc646b329310b95bccd/b9f35990-3a1e-4290-b34d-318ba768efcb)

![C4E553EF-A6C2-489C-A2D3-12F8D35C7CEE](https://user-images.githubusercontent.com/48343061/60302147-f5c97a00-98e7-11e9-85b8-72d43d930687.jpeg)

rasmus-storjohann-PG:
Need to get #366 merged before this can go in. #366 sets up system for managing translations, this PR contains changes that will be managed using that system.

jparlina:
Tested Staging V1.0.24 - 

![DA26178E-4A43-4961-9126-8F786DE77B5F.jpeg](https://images.zenhubusercontent.com/5ceeccc646b329310b95bccd/87056d58-1f55-4bd4-a7bf-75ac25f15267)

![69B7EC21-2859-44BA-BD95-FB134CC6C1DA.jpeg](https://images.zenhubusercontent.com/5ceeccc646b329310b95bccd/b1db01f2-15ea-4ae7-86e7-4fb3950a8205)

![11639E9F-802C-4822-B38C-3F639D2A99E1.jpeg](https://images.zenhubusercontent.com/5ceeccc646b329310b95bccd/5c279315-3d56-49ec-9343-713d8fd28676)

jparlina:
@ArtiBPG The following items are still pending:

- Update the website URL to https://arrivaladvisor.ca/terms-of-use/ 
- Update the bottom menu on the website from "Terms and Conditions" to "Terms of Use"
- Translate "Terms of Use" for all non-English version of the app

cherrielam:
See checklist above - I have now done the first two items. Last item needs to be reviewed by @ArtiBPG

jparlina:
[...]

PR #667, opened 2019-06-11
Issue #662 Fix wording and url for Terms of Use

rasmus-storjohann-PG:


rasmus-storjohann-PG:
- [ ] Change export strings script to name output files `jsx_strings.po`
- [ ] Remove output of CSV files, no longer needed other than as temp files
- [ ] Generate po files
- [ ] Push generated files to the translate repo

PR #670, opened 2019-06-12
Feature/issue 625 move questionnaire strings to po files

tomy-pg:


bin/prepare_deploy.sh
```diff
@@ -320,9 +344,11 @@ validateContentFixture
 getClientDependencies
 createClientEnvironment
 completeManualConfiguration
+./bin/strings.sh --combine-pos-deploy $CLIENT_LOCALE_LOCATION $WORKING_DIRECTORY
+buildStringsForLinguiCatalogs
 buildClientLocally
 testClient
 
 giveExpoCommandsForPublishing
```
rasmus-storjohann-PG: - [ ] Mandated order of po files for concat
- [ ] No longer delete messages.po files
- [ ] Fix commit messages
- [ ] Fail on temporary working directory already existing
- [ ]

bin/prepare_deploy.sh
```diff
@@ -320,9 +344,11 @@ validateContentFixture
 getClientDependencies
 createClientEnvironment
 completeManualConfiguration
+./bin/strings.sh --combine-pos-deploy $CLIENT_LOCALE_LOCATION $WORKING_DIRECTORY
```
rasmus-storjohann-PG: This line should follow the established pattern where it does a CD to the client directory and then runs the command, which can then make assumptions about paths. Note that strings.sh already assumes that it will be executed from the client directory, so you can continue to make that assumption

Issue #663, opened 2019-06-08
Staging V 1.0.19: Service Description - extra indentation before first word and last word gets cutoff before "Read more"


jparlina:
See Service Description on the screenshot below:
- There's an extra space before the first word in the description
- The last word gets cutoff before "Read more" (see red highlight on the right side of the screen and compare with the green highlight after "Read more" was clicked)

![IMG_0957.JPG](https://images.zenhubusercontent.com/5ceeccc646b329310b95bccd/98a4338d-a624-4ea3-a657-82e8762f4e82)

![IMG_0959.JPG](https://images.zenhubusercontent.com/5ceeccc646b329310b95bccd/be4be631-0ea1-436e-bce0-0f03d0481d2c)

bensinjin:
I'm closing this PR as we've decided it runs contrary to designs outlined by Cherrie.

PR #675, opened 2019-06-14
Issue #663 Remove additional padding from exapandable content component

bensinjin:
So I've removed some additional padding we were adding to this component and also the overflow settings for iOS which means the text won't look cut off. On Android we're going to have to live with the cut off text as this is actually a platform specific quirk.

### DIFF
--- a/src/components/expandable_content/expandable_content_component.tsx
+++ b/src/components/expandable_content/expandable_content_component.tsx
@@ -1,6 +1,6 @@
 // tslint:disable:no-class no-expression-statement no-this
 import React from 'react';
-import { Dimensions, LayoutChangeEvent, Animated, I18nManager, Platform } from 'react-native';
+import { Dimensions, LayoutChangeEvent, Animated, I18nManager } from 'react-native';
 import { View, Text, Button, Icon } from 'native-base';
 import { Trans } from '@lingui/react';
 import { colors, textStyles } from '../../application/styles';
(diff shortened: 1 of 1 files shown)


src/components/expandable_content/expandable_content_component.tsx
```diff
@@ -72,19 +72,22 @@ export class ExpandableContentComponent extends React.Component<ExpandableConten
...
         );
     }
 
     private getCollapsedStyle(): object {
-        return {
+        const style = {
             height: this.state.collapseAtHeight,
+        };
+        if (Platform.OS === 'ios') {
+            return style;
```
tomy-pg: without overflow: 'hidden' for ios I am seeing this on Xcode's simulator (using iPhone 8 Plus):

Screen Shot 2019-07-03 at 12 18 24 PM
Screen Shot 2019-07-03 at 12 18 59 PM

bensinjin: Good point @tomy-pg. I looked closer at this and really couldn't come up with anything better than just using `overflow: 'hidden'` for both platforms. I want to be clear that the major issue here is dynamically altering a "non text" component's height after the initial render NOT the fact that we're using the `Markdown` component.

bensinjin:
We're closing this and ensuring any additional styling needed will be included in the PR for https://github.com/pg-irc/pathways-frontend/issues/703

Issue #682, opened 2019-06-20
Use Webview instead of Markdown component


rasmus-storjohann-PG:
Currently we're using a mark down component from `react-native-markdown-renderer` that is hard to customize.

# heading
## heading2

ACs:
* Frontend: Replace the Markdown component with the web view from https://github.com/react-native-community/react-native-webview. Note that the webview shows content generated client side.
* Backend: Change the Newcomers Guide import script to convert mark-down to HTML tags. Note that we're using a very small subset of the markdown features, so it's not that complex. See `/pathways-backend/newcomers_guide/clean_data.py`

ian-lin-PG:
Back end files have been converted to produce HTML instead of markdown. Front end react-native-webview component is incompatible with Expo, currently looking for workarounds.

ian-lin-PG:
The current Expo SDK v32.0.0 is based on React Native 0.57 “September 2018”. However, in October there has been an [architecture change](https://www.reddit.com/r/reactnative/comments/9sim5r/any_thoughts_on_the_react_native_slimming_moving/) where the web-view component has been moved to it's own npm module named react-native-webview. This change introduced a (bug)[https://github.com/react-native-community/react-native-webview/issues/140] for using the new react-native-webview module.

The fix in expo can be found [here](https://github.com/expo/expo/issues/3381) and should have been merged in the [newest release](https://blog.expo.io/expo-sdk-v33-0-0-is-now-available-52d1c99dfe4c). The newest release of expo is also based on the [March 2019 version of react-native](https://facebook.github.io/react-native/blog/2019/03/12/releasing-react-native-059)  which should include the architecture change that happened last October.

ian-lin-PG:
Upgraded expo to version 33.0.0 without errors, new warmings requires react-native link which is currently [unsupported] (https://github.com/expo/expo/issues/56) without detaching to expokit.

ian-lin-PG:
Webview is currently enabled but there are a few outstanding issues:
1. Webview component does not currently scale to the screen size with flex, instead need hard coded dimensions to render
2. Webview component does not scroll vertically if the container outside is able to scroll vertically, might be an android issue.
3. Component above the Webview may be hidden until Webview component is interacted with
4. Webview redirects links WITHIN the component with no way to navigate back

ian-lin-PG:
[...]

PR #693, opened 2019-06-27
Issue #682 Upgrade Expo from 32.0.0 to 33.0.0

ian-lin-PG:


package.json
```diff
@@ -58,13 +58,18 @@
             "/node_modules/",
             "/lib/",
             "/helpers/"
-        ]
+        ],
+        "transformIgnorePatterns": [
+			"node_modules/?!(react-native|native-base)"
+		]
```
rasmus-storjohann-PG: The syntax here is a little confusing, it this equivalent to this simpler syntax?
```
        "transformIgnorePatterns": [
			"node_modules/react-native",
			"node_modules/native-base",
		]
```

ian-lin-PG: According to the documentation [here](https://jestjs.io/docs/en/configuration#transformignorepatterns-array-string) and [here](https://jestjs.io/docs/en/tutorial-react-native#transformignorepatterns-customization) The line takes in a regex pattern to whitelist certain folders or modules. I haven't seen it utilized in multiple lines like you suggested, however, I did notice that removing 'react-native' like so works fine:

`"transformIgnorePatterns": [
            "node_modules/?!(native-base)/"
        ]`

The plausible explanation could be that 'react-native' is already whitelisted from the testPathIgnorePatterns options right before this one. If we choose to whitelist 'native-base' would that be less confusing?

PR #710, opened 2019-07-04
Feature/Issue 687 Upgrade Expo including @types modules

ian-lin-PG:


package.json
```diff
@@ -7,7 +7,6 @@
         "@lingui/babel-preset-react": "2.7.3",
         "@lingui/cli": "2.7.3",
         "@lingui/react": "2.7.3",
-        "@types/expo": "^32.0.0",
```
bensinjin: Why are we dropping these types? Have they been brought into the main project?

bensinjin: (You may have already answered this in our discussions but maybe reiterate here if you don't mind)

ian-lin-PG: The type declaration starting with expo 33.0.0 is automatically included in the standard module,  the release note recommends removing it. 

I can understand the reasoning, since outdated @types/expo may end up overriding the correct type declaration already within the standard expo 33.0.0 module.

bensinjin:
@ian-lin-PG this'll need to be rebased to resolve conflicts. @tomy-pg I'm wondering if you can ensure that the build process for `IPAs` and `APKs` is still working as expected?

tomy-pg:
@bensinjin I can test this to ensure that it is working

tomy-pg:
@bensinjin currently the build process to generate the APK and IPA is not working.

bensinjin:
Closing this as we're addressing it in another PR.

Issue #703, opened 2019-07-04
Mark-down styling issues


rasmus-storjohann-PG:
Issues with the current mark down component:
* Use the same font as the app, should be avenir (this is also wrong for the topic teaser string)
* Have nice headers
* Body text font bigger (text size 16sp from Zepelin)
* Body text font colour dark grey as per Zepelin
* Change the link colour from blue to green (one line change)

Also:
* About page body text font colour should be dark grey as per Zepelin

Reference:
* Zepelin https://zpl.io/aByGdBm

bensinjin:
For Ben: Review these requirements and ensure our Markdown component is sufficient.

cherrielam:
A few remaining UI issues related to the markdown/font styling:
1. The header font used for the category of a topic (eg. "Education" below) should be capitalized. It should be using the same header style is "Related topics"
![IMG_4429](https://user-images.githubusercontent.com/21990736/61565118-f065d800-aa2c-11e9-8751-2266bd603fdb.PNG)

![IMG_4433](https://user-images.githubusercontent.com/21990736/61565326-8bf74880-aa2d-11e9-93cf-c0fe50fc9ebf.PNG)


2. (see image above) The Read More component - "Read more" text needs to be aligned with the Body text above. This may be related to other "Read More" issues... maybe consult @bensinjin to see when this will be tackled.

3. The teaser text is not in Avenir, it is currently using the device default font (on both Apple and Android)
![IMG_4430](https://user-images.githubusercontent.com/21990736/61565214-3753cd80-aa2d-11e9-8c0a-d761f9badea9.PNG)

4. The links are not using Avenir, also using the device default font.
![IMG_4431](https://user-images.githubusercontent.com/21990736/61565251-55213280-aa2d-11e9-9047-6cd14077e5ef.PNG)

ian-lin-PG:
1. This will be more difficult because the explore section names are translated and used throughout the application, it would be hard to keep it capitalized in one place and not in all the other places. Alternatively, the 'RELATED TOPICS' is hard coded and will barely take any time to change to 'Related Topics', just a thought.

2. I applied marginLeft:5 on the 'Read More' and it's more aligned now.

3 and 4 I applied fontFamily: 'AvenirBook' to the components in question.

rasmus-storjohann-PG:
@cherrielam can you QA this story, you're then ones with the sharp eye for subtle font differences.

cherrielam:
[...]

PR #719, opened 2019-07-05
Issue #703 Style markdown component

bensinjin:
Along with the styling outlined in #703 these changes will also address some unwanted padding for the "Read more" component, discussed here: https://github.com/pg-irc/pathways-frontend/issues/663. The issues with cut-off text have been thoroughly discussed and come down to platform specific problems that we don't intend on fixing at this moment.

It may be useful if @cherrielam took a look at this PR as I've had to somewhat improvise on the heading style (although it should seem reasonable).

### DIFF
--- a/src/application/styles.ts
+++ b/src/application/styles.ts
@@ -309,6 +309,7 @@ export const markdownStyles = StyleSheet.create({
         fontFamily: 'AvenirBook',
         fontSize: 16,
         lineHeight: 21,
+        paddingHorizontal: values.backgroundTextPadding,
     },
     heading: {
         lineHeight: 26,
--- a/src/application/styles.ts
+++ b/src/application/styles.ts
@@ -305,12 +305,28 @@ export const applicationStyles = StyleSheet.create({
 export const markdownStyles = StyleSheet.create({
     text: {
         textAlign: 'left',
+        color: colors.greyishBrown,
+        fontFamily: 'AvenirBook',
+        fontSize: 16,
+        lineHeight: 21,
+    },
+    heading: {
+        lineHeight: 26,
+        marginTop: 10,
+        ...getBoldFontStylesForOS(),
+    },
+    heading1: {
+        fontSize: 22,
+      },
+    heading2: {
+        fontSize: 20,
     },
     listUnorderedItemIcon: {
         fontWeight: 'bold',
         fontSize: 35,
         marginLeft: 10,
         marginRight: 10,
+        color: colors.greyishBrown,
         ...Platform.select({
             ios: {
                 lineHeight: 36,
@@ -321,7 +337,7 @@ export const markdownStyles = StyleSheet.create({
         }),
     },
     link: {
-        color: 'blue',
+        color: colors.teal,
         textDecorationLine: 'underline',
     },
 });
--- a/src/application/styles.ts
+++ b/src/application/styles.ts
@@ -312,15 +312,20 @@ export const markdownStyles = StyleSheet.create({
         paddingHorizontal: values.backgroundTextPadding,
     },
     heading: {
-        lineHeight: 26,
         marginTop: 10,
         ...getBoldFontStylesForOS(),
     },
     heading1: {
-        fontSize: 22,
+        lineHeight: 23,
+        fontSize: 18,
       },
     heading2: {
-        fontSize: 20,
+        lineHeight: 22,
+        fontSize: 17,
+    },
+    heading3: {
+        lineHeight: 21,
+        fontSize: 16,
     },
     listUnorderedItemIcon: {
         fontWeight: 'bold',
--- a/src/components/expandable_content/expandable_content_component.tsx
+++ b/src/components/expandable_content/expandable_content_component.tsx
@@ -1,7 +1,7 @@
 // tslint:disable:no-class no-expression-statement no-this
 import React from 'react';
-import { Dimensions, LayoutChangeEvent, Animated, I18nManager } from 'react-native';
-import { View, Text, Button, Icon } from 'native-base';
+import { Dimensions, LayoutChangeEvent, Animated, I18nManager, TouchableOpacity, ViewStyle } from 'react-native';
+import { View, Text, Icon } from 'native-base';
 import { Trans } from '@lingui/react';
 import { colors, textStyles } from '../../application/styles';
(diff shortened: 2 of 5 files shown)


cherrielam:
As per discussion with @bensinjin earlier, if we can add the same kind of padding to the body (markdown component) of topic detail that would be great. Maybe for the heading we could use Heading H3 style or Heading H5 style? as per: https://zpl.io/bLxdOXM (currently is the heading H1 or H2?)

bensinjin:
Thx @cherrielam, to elaborate (personal note to self): We want add the (think it's 10dpi, there should be a RN "class" for this as it's applied throughout the app) padding to the markdown component.

bensinjin:
@cherrielam's comment has been addressed.

bensinjin:
@rasmus-storjohann-PG there are two conflicts here that I'd like to properly understand. I'm sure we could likely rebase this more accurately/quickly if we did it together, let's coordinate something so nothing important is lost.

rasmus-storjohann-PG:
@bensinjin I'll have a shot at this rebase

bensinjin:
> @bensinjin I'll take a shot at this rebase

Sounds good, I can do a quick review after if you need

rasmus-storjohann-PG:
@bensinjin you're right that the easiest way to resolve this is for the two of us to sit down with it. However, that can't happen this week. The changes you're needing to merge in are related to the views as they appear in Arabic. I'll outline the possible regressions that your change may create and you can double check that it all works:
* Task detail page: In Arabic: Keep a close eye on the "Read more"/"Read less" buttons, making sure the arrow ends up on the correct side of the button.
* Service list page: In Arabic, the page should be all in English and look like the page does in English, including the Read more/Read less. My changes were making sure that the arrow appears on the correct side of the text on the label. In the process I also cleaned up the code a bit which made for difficult merge conflicts. That's a risk with cleanup.
So can you just make sure that your PR doesn't break these things and then rebase selecting your changes over mine throughout?

rasmus-storjohann-PG:
This business I needed to do
```
const button = this.flipLeftRightOrientation() ?
            <Button onPress={onPress} transparent iconLeft >
                {buttonIcon}
                {buttonText}
            </Button> :
            <Button onPress={onPress} transparent iconRight >
                {buttonText}
                {buttonIcon}
            </Button>;
```
was all for getting the icon on the correct side of the button.

rasmus-storjohann-PG:
This change
```ts
    const label = <Text style={[textStyles.paragraphBoldBlackLeft, textStyles.alwaysLeftAlign]}>Web: </Text>;
    const link = <AnalyticsLink href={website} currentPath={currentPath} linkContext={linkContext}
        linkType={'Website'} style={textStyles.paragraphStyle}>{website}</AnalyticsLink>;

    return <Text>{label}{link}</Text>;

```
is just about making the code more readable. The long blocks of more than 5 lines of JSX always breaks the single responsibility principle. You don't need to make these changes, but I would be glad to see you shift your JSX coding style towards smaller blocks of markdown.

bensinjin:
@rasmus-storjohann-PG I've completed the rebase and kept in mind your comments, please take a final look.

We sort of have two problems: code style and how to apply SRP to component construction.

Code style is easily addressed with tools like [Prettier](https://prettier.io/), this has worked great on other projects and I'd be happy to get it brought into Arrival Advisor

In regards to SRP when it comes to JSX... I know this has been an ongoing discussion ...

Can we clearly define when a component breaks SRP? On another note, saying a component should only be N number of lines just makes people write longer lines (from what I can tell). I mean technically as you say JSX is always breaking SRP. Also, in React, the major point we would have for splitting up a component is to reuse it elsewhere - I personally have found this to be a much better guide than SRP (within reason, I also don't like huge walls of JSX). I personally think we already have too many "sub" components and that the reason this conversation keeps happening is the fact that SRP doesn't nicely apply to an HTML like structure.

Regardless of my thoughts, I can try and make my component more small and focused.

rasmus-storjohann-PG:
Yeah, @bensinjin  I realize that we have had this discussion before. Here's an example of where there is real value to breaking up the JSX:
```ts
const button = this.flipLeftRightOrientation() ?
            <Button onPress={onPress} transparent iconLeft >
                {buttonIcon}
                {buttonText}
            </Button> :
            <Button onPress={onPress} transparent iconRight >
                {buttonText}
                {buttonIcon}
            </Button>;
```
Here there is some logic to return two variants of a button depending on the situation. This code is written so that the things that are different in the two variants is as clear as I could make it. This was accomplished by removing everything that is NOT different out into other local variables of JSX.

rasmus-storjohann-PG:
In this example
```ts
    const label = <Text style={[textStyles.paragraphBoldBlackLeft, textStyles.alwaysLeftAlign]}>Web: </Text>;
    const link = <AnalyticsLink href={website} currentPath={currentPath} linkContext={linkContext}
        linkType={'Website'} style={textStyles.paragraphStyle}>{website}</AnalyticsLink>;

    return <Text>{label}{link}</Text>;
```
the benefit is to tell the reader that this code makes a text element that contains a label and a link. If it was one big JSX block, then the reader woulld need to parse a lot of JSX, making sure that the white space lines things up correctly so that (s)he parses it correctly, etc, and then interpret each part of the JSX to figure out that it is a label and a link. That is a lot more work.

bensinjin:
Your first point I entirely agree with and make a point of doing myself (let me know when I miss it).

To your second point, this:
```
const label = <Text style={[textStyles.paragraphBoldBlackLeft, textStyles.alwaysLeftAlign]}>Web: </Text>;
const link = <AnalyticsLink href={website} currentPath={currentPath} linkContext={linkContext} linkType={'Website'} style={textStyles.paragraphStyle}>{website}</AnalyticsLink>;
```
I would much sooner write over multiple lines like:
```
... ("label" one is probably fine on one line)
const link = (
    <AnalyticsLink
        href={website}
        currentPath={currentPath}
        linkContext={linkContext}
        linkType={'Website'}
        style={textStyles.paragraphStyle}
    >
            {website}
    </AnalyticsLink>
);
```
And I don't think that should be "not allowed" because it's more than 5 lines. Maybe you could say no more than N nested components per component but 5 lines doesn't seem like the right rule to be making.

rasmus-storjohann-PG:
Did a QA pass on this, looking for regressions in the left/right alignment in Arabic. Looks good.

Issue #740, opened 2019-07-19
Staging V 1.0.25 - Bookmarks disappear when switching EN <> AR


jparlina:
Tested on iOS Staging V1.0.25:
Started app in Arabic, bookmarked certain topics, switched to English:
- the English version opens up in home page instead of staying on the same page that I had opened in the Arabic version
- all my bookmarks disappeared even though I didn’t “Delete all user data”

Same when I started in English, then switched to Arabic.

For other languages, bookmarks did not disappear and you didn’t get rerouted to Home page every time you switch language. 

Note: Answers to questionnaires did not disappear when switching between EN <> AR

ian-lin-PG:
The issues exists when changing from any languages to Arabic and vice versa. Preliminary theory is that the alignment utility, I18nManager, doesn't save the state when relaunching app with the correct language orientation. The issue also doesn't exist in production, which means one of the merged changes specifically related to Arabic caused this change.

Currently combing through merged changes for clues.

jparlina:
Tested Staging V1.1.0, server 1.3.5 on iOS 12.4:

Started at language selection screen - selected Arabic - closed onboarding - bookmarked several topics on Home screen - checked My Bookmarks - changed language to English - PASSED:
- language changed in English and stayed on the same screen instead of opening new language on Home screen
- bookmarks did not disappear when language was changed

Deleted all user data on the app - in English version
Bookmarked topics on Home screen in English version 
Changed language to Arabic - Onboarding screen showed up - is this expected? @tomy-pg 
Bookmarks did not disappear - PASSED 

No more issue with bookmarks so moving this to "Ready for Production" 
Onboarding screen issues are separate.

PR #748, opened 2019-07-26
Issue #740 Fix bookmarked topics not persisting between app reloads

ian-lin-PG:


src/sagas/__tests__/user_data.test.ts
```diff
@@ -0,0 +1,23 @@
...
+
+import { PersistedUserDataBuilder } from '../../stores/__tests__/helpers/user_data_helpers';
+import { aString } from '../../application/__tests__/helpers/random_test_values';
+import { serialize, deserialize } from '../user_data';
+
+describe('persistent user_data tests', () => {
+        
+    test('chosen answer should persist between being serialized and deserialized', () => {
+        const persistentData = new PersistedUserDataBuilder().addChosenAnswer(aString()).buildObject();
+        expect(persistentData).toEqual(deserialize(serialize(persistentData)));
```
rasmus-storjohann-PG: Making tests easy to understand is greatly helped by having very specific assertions that express exactly what you care about. So here, you need to assert on the return value from `deserialize(serialize(persistentData))`, and specifically the chosen answer string within that return value. You'll find that you will actually need to store the `aString()` value in a local variable to do that assertion, and this is the pattern that you will see in a lot of tests.

src/sagas/user_data.ts
```diff
@@ -68,15 +68,15 @@ export async function loadUserDataAsync(): Promise<string> {
     return await AsyncStorage.getItem(USER_DATA_STORAGE_KEY);
 }
 
-const deserialize = (serializedUserData: string): PersistedUserData => (
+export const deserialize = (serializedUserData: string): PersistedUserData => (
```
rasmus-storjohann-PG: Rename functions as appropriate given their now broader scope

src/sagas/__tests__/user_data.test.ts
rasmus-storjohann-PG: Also improve consistency between persistentData vs PersistedUserDataBuilder, persisted.

src/sagas/user_data.ts
```diff
@@ -30,7 +30,7 @@ type SaveActions = IterableIterator<
 export function* saveUserData(): SaveActions {
     try {
         const userData: PersistedUserData = yield select(selectUserDataForLocalPersistence);
-        const serializedUserData = serialize(userData);
+        const serializedUserData = serializePersistedData(userData);
```
rasmus-storjohann-PG: Be consistent with naming, we're using "user data", so don't confusing things with "persisted data".

src/sagas/__tests__/user_data.test.ts
```diff
@@ -0,0 +1,29 @@
...
+import { aString } from '../../application/__tests__/helpers/random_test_values';
+import { serializePersistedData, deserializePersistedData } from '../user_data';
+
+describe('persisted user_data tests', () => {
+        
+    test('chosen answer should persist between being serialized and deserialized', () => {
+        const chosenAnswer = aString();
+        const persistedData = new PersistedUserDataBuilder().addChosenAnswer(chosenAnswer).buildObject();
+        const recreatedData = deserializePersistedData(serializePersistedData(persistedData));
+        expect(chosenAnswer).toEqual(recreatedData.chosenAnswers[0]);
```
rasmus-storjohann-PG: Prefer to put the data from tested code first, and assert that it matches the expected value, not the other way around. So
```
expect(recreatedData.chosenAnswers[0]).toEqual(chosenAnswer);
```

Issue #724, opened 2019-07-10
Review and sign off on onboarding screen and text


ArtiBPG:
Description:
Review the onboarding screen on the staging app: 1.0.23 for content and flow.

 Issues already identified:
- Test by leaving the app and opening it again, changing language EN>Any other language. NO language selection as soon as the user opens the app. see #606. The fix the make the app remember language selection broke the language selection<>onboarding flow.
- Test onboarding in all languages for text expansion and layout, see FR situation: https://screencast.com/t/z5D8zRb64p
- The back and next buttons are not the same as the back and next buttons for the questionnaire.
- Another issue found during QA: https://github.com/pg-irc/arrival-advisor-qa/issues/13

Next steps:
Provide updates mocks and update the issue.

ACs:
Content and UX is acceptable and ready for release.

cherrielam:
Reviewing the onboarding has uncovered a few usability issues:

### UX problems
- dumping the user directly into the questionnaire is not an ideal experience because they don't get a real sense of what the app does and suddenly have to answer a bunch of questions.
- user has no ability to exit the onboarding early
- the back and next buttons feel hard to use, not as intuitive as the swiping pattern of typical onboarding

### Content problems
- content of the onboarding screens is no longer up to date. Images are placeholders.

## Recommendations
- Please see mocks here: https://zpl.io/VD4DB5r titled ONBOARDING/1 to 4. The graphics for the screen can be downloaded directly here.

1. Flow: Onboarding screens need to be **after** the welcome screen (where the user selects the language)
2. Add ability to **SKIP** in the top right corner. It is equivalent to exiting the onboarding flow. Can be replaced with an X if this is preferred? (What are your thoughts on X vs. "Skip"? @ArtiBPG @rasmus-storjohann-PG ) 
3. **Start** Button on the last screen takes you directly to the HOME screen.
4. **ALTERNATIVELY** we can remove the "Skip" button and add the "Start" button on every single screen of the onboarding flow. So the user can "Start" from any onboarding step. (Which is preferred? @ArtiBPG @rasmus-storjohann-PG 
5. Implement swiping ability to the onboarding screens for next and back. Should not need words (reduce need for translation).
6. Images and text have been updated on the screens. Please see the Zeplin link above. Most of the text has already been translated (from the website).

@ArtiBPG @rasmus-storjohann-PG I recommend reviewing & discussing before developing

rasmus-storjohann-PG:
[...]

PR #749, opened 2019-07-26
Issue #724 Update onboarding to reflect designs

bensinjin:
Most of this work should be pretty straight forward. I made a decision to reuse the "X" (aka close button) we used elsewhere in the app as it was mentioned on the issue but no one responded. In my opinion is adds consistency and generally it's easier to reuse a component and avoid additional translations. If this is a deal breaker we can easily change it.

I've done testing across iOS and Android, it's looking good to me! I haven't had the time to write the animations for swiping between screens but it's definitely something we can get in at some point.


src/components/swipeable_content/swipeable_content_component.tsx
```diff
@@ -40,7 +40,12 @@ const buildPanResponder = (currentIndex: number, itemCount: number, setState: Se
     })
 );
 
-const onMoveShouldSetPanResponder = (_event: GestureResponderEvent, _gestureState: PanResponderGestureState): boolean => true;
+const onMoveShouldSetPanResponder = (_event: GestureResponderEvent, gestureState: PanResponderGestureState): boolean => (
+    // For 3d touch devices (newer iPhones) we have problems with tappable items not being tappable when nested in the pan responder.
+    // See https://github.com/facebook/react-native/issues/3082 for more details.
+    // This basically says "your tap must be moving horizontally" in order for us to set the responder.
+    gestureState.dx !== 0 ? true : false
```
rasmus-storjohann-PG: This code is difficult to understand, so the comment is helpful. However, I don't see the need for the ternary operator here, just `gestureState.dx !== 0` is better.

rasmus-storjohann-PG:
Start Button on the last screen takes you directly to the HOME screen

bensinjin:
Fixed!

src/components/swipeable_content/swipeable_content_component.tsx
```diff
@@ -0,0 +1,102 @@
...
+const getNextIndex = (currentIndex: number, itemCount: number): number => (
+    currentIndex + 1 < itemCount ? currentIndex + 1 : currentIndex
+);
+
+const getPreviousIndex = (currentIndex: number): number => (
+    currentIndex - 1 >= 0 ? currentIndex - 1 : 0
+);
+
+const isRightToLeftSwipe = (horizontalMovementValue: number): boolean => (
+    horizontalMovementValue < 0 && Math.abs(horizontalMovementValue) > swipeSensitivity
```
rasmus-storjohann-PG:
```
horizontalMovementValue < -1.0 * swipeSensitivity
```

src/components/swipeable_content/swipeable_content_component.tsx
```diff
@@ -0,0 +1,102 @@
...
+);
+
+const isLeftToRightSwipe = (horizontalMovementValue: number): boolean => (
+    horizontalMovementValue > 0 && Math.abs(horizontalMovementValue) > swipeSensitivity
```
rasmus-storjohann-PG:
```
horizontalMovementValue > swipeSensitivity
```

src/components/swipeable_content/swipeable_content_component.tsx
```diff
@@ -0,0 +1,102 @@
...
+);
+
+const updateIndexIfChanged = (currentIndex: number, newIndex: number, setState: SetState): void => {
+    if (currentIndex !== newIndex) {
+        setState({ currentIndex: newIndex });
+    }
+};
```
rasmus-storjohann-PG: Pretty sure react is smart enough to ignore a setState call that doesn't change the state? Then this function is not needed.

src/components/swipeable_content/swipeable_content_component.tsx
```diff
@@ -0,0 +1,102 @@
...
+    const dotSize = 8;
+    const dotStyle = { width: dotSize, height: dotSize, borderRadius: dotSize / 2, marginHorizontal: 5 };
+    if (props.currentIndex === props.loopIndex) {
+        return <View style={{ ...dotStyle, backgroundColor: colors.topaz }} />;
+    }
+    return <View style={{ ...dotStyle, backgroundColor: colors.lightGrey }} />;
+};
+
+const getNextIndex = (currentIndex: number, itemCount: number): number => (
+    currentIndex + 1 < itemCount ? currentIndex + 1 : currentIndex
```
rasmus-storjohann-PG:
```
R.min(currentIndex + 1, itemCount - 1)
```

src/components/swipeable_content/swipeable_content_component.tsx
```diff
@@ -0,0 +1,102 @@
...
+);
+
+const getPreviousIndex = (currentIndex: number): number => (
+    currentIndex - 1 >= 0 ? currentIndex - 1 : 0
```
rasmus-storjohann-PG:
```
R.max(currentIndex - 1, 0)
```

src/components/swipeable_content/swipeable_content_component.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+interface SwipeableContentComponentProps {
+    readonly contentItems: ReadonlyArray<JSX.Element>;
+}
+
+interface State {
+    readonly currentIndex: number;
+}
+
+type SetState = React.Dispatch<React.SetStateAction<State>>;
```
rasmus-storjohann-PG: You can probably use `Readonly<ReturnType<typeof buildState>>;` here

src/components/swipeable_content/swipeable_content_component.tsx
bensinjin: Fixed

src/components/swipeable_content/swipeable_content_component.tsx
bensinjin: Nice and tidy, fixed.

src/components/swipeable_content/swipeable_content_component.tsx
bensinjin: fixed

src/components/swipeable_content/swipeable_content_component.tsx
bensinjin: yeah you're right, fixed

src/components/swipeable_content/swipeable_content_component.tsx
bensinjin: fixed

src/components/swipeable_content/swipeable_content_component.tsx
bensinjin: fixed

src/components/swipeable_content/swipeable_content_component.tsx
bensinjin: Dunno about this - TS didn't like "the circular reference" - I don't have time at this minute to dig much further.

PR #754, opened 2019-07-30
Feature/issue 636 fix rtl issue

tomy-pg:


src/stores/locale/actions.ts
```diff
@@ -0,0 +1,47 @@
+import * as constants from '../../application/constants';
+import * as helpers from '../helpers/make_action';
+
+export type LoadCurrentLocaleRequestAction = Readonly<ReturnType<typeof loadCurrentLocaleRequest>>;
+export type LoadCurrentLocaleSuccessAction = Readonly<ReturnType<typeof loadCurrentLocaleSuccess>>;
+export type LoadCurrentLocaleErrorAction = Readonly<ReturnType<typeof loadCurrentLocaleFailure>>;
+export type LoadCurrentLocaleResult = LoadCurrentLocaleSuccessAction | LoadCurrentLocaleErrorAction;
```
rasmus-storjohann-PG: Naming: "Current" seems to be redundant in all these names. I also don't think the `LoadCurrentLocaleResult` type is needed.

src/stores/locale/actions.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export type LoadCurrentLocaleAction = LoadCurrentLocaleRequestAction | LoadCurrentLocaleSuccessAction | LoadCurrentLocaleErrorAction;
+
+export type SetLocaleRequestAction = Readonly<ReturnType<typeof setLocaleRequest>>; 
+export type SetLocaleSuccessAction = Readonly<ReturnType<typeof setLocaleSuccess>>; 
+export type SetLocaleErrorAction = Readonly<ReturnType<typeof setLocaleFailure>>; 
+export type SetLocaleResult = SetLocaleSuccessAction | SetLocaleErrorAction
+
+export type SetLocaleAction = SetLocaleRequestAction | SetLocaleSuccessAction | SetLocaleErrorAction;
```
rasmus-storjohann-PG: The opposite of "Load" is "Save", so maybe the names should use "Save" rather than "Set"?

src/stores/locale/actions.ts
```diff
@@ -0,0 +1,47 @@
...
+
+// tslint:disable-next-line:typedef
+export const loadCurrentLocaleRequest = () => {
+    return helpers.makeAction(constants.LOAD_CURRENT_LOCALE_REQUEST);
+};
+
+// tslint:disable-next-line:typedef
+export function loadCurrentLocaleSuccess(message: string, loading: boolean, localeCode: string, isSet: boolean) {
+    return helpers.makeAction(constants.LOAD_CURRENT_LOCALE_SUCCESS, { message, loading, localeCode, isSet });
+}
```
rasmus-storjohann-PG: The message is always the empty string and the loading flag is always false, so these fields are not needed. They may be needed one day in the future, but then they should be added then, not now. Code maintenance is costly, so we only maintain code that does something of value *now*, we don't write code that anticipates the future. Mostly because we are usually wrong about what will happen.

src/stores/locale/actions.ts
```diff
@@ -0,0 +1,47 @@
...
+
+// tslint:disable-next-line:typedef
+export function loadCurrentLocaleFailure(message: string, loading: boolean) {
+    return helpers.makeAction(constants.LOAD_CURRENT_LOCALE_FAILURE, { message, loading });
+}
```
rasmus-storjohann-PG: Remove `loading` property, it's always false.

src/stores/locale/index.ts
```diff
@@ -29,27 +31,79 @@ export const buildDefaultStore = (): LocaleStore => ({
...
-
-export const reducer = (store: LocaleStore = buildDefaultStore(), action?: ReducerActions): LocaleStore => {
+export const reducer = (store: LocaleStore = buildDefaultStore(), action?: LocaleAction): LocaleStore => {
     if (!action) {
         return store;
     }
     switch (action.type) {
         case constants.LOAD_CURRENT_LOCALE_REQUEST:
-            return { ...store, errorMessage: '', loading: true };
+            return updateCurrentLocaleRequest(store);
```
rasmus-storjohann-PG: I like the old way better, same below.

src/stores/locale/actions.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export const setLocaleRequest = (localeCode: string) => {
```
rasmus-storjohann-PG: Your linter is not working, otherwise you would have noticed the missing rule disable comment here.

src/stores/locale/actions.ts
```diff
@@ -0,0 +1,47 @@
+import * as constants from '../../application/constants';
+import * as helpers from '../helpers/make_action';
+
+export type LoadCurrentLocaleRequestAction = Readonly<ReturnType<typeof loadCurrentLocaleRequest>>;
+export type LoadCurrentLocaleSuccessAction = Readonly<ReturnType<typeof loadCurrentLocaleSuccess>>;
+export type LoadCurrentLocaleErrorAction = Readonly<ReturnType<typeof loadCurrentLocaleFailure>>;
```
rasmus-storjohann-PG: Use consistent terminology, "Failure", not "Error".

src/stores/locale/index.ts
```diff
@@ -1,15 +1,17 @@
...
-import * as loadCurrentLocale from './load_current_locale';
-import { SetLocale } from './set_locale';
-import { LoadCurrentLocale } from './load_current_locale';
 import { LocaleInfo } from '../../locale';
+import { 
+    LoadCurrentLocaleRequestAction, loadCurrentLocaleRequest,
+    LoadCurrentLocaleSuccessAction, loadCurrentLocaleSuccess, LoadCurrentLocaleErrorAction, loadCurrentLocaleFailure,
+    SetLocaleRequestAction,setLocaleRequest, SetLocaleSuccessAction, setLocaleSuccess, SetLocaleErrorAction, setLocaleFailure,  
+    SetLocaleAction, LoadCurrentLocaleAction, SetLocaleResult, LoadCurrentLocaleResult, LocaleAction
+} from './actions';
```
rasmus-storjohann-PG: This is messy. I prefer
```ts
import * as actions from './actions';
```

src/stores/locale/index.ts
```diff
@@ -1,15 +1,17 @@
...
-export { SetLocale };
-export { LoadCurrentLocale };
-
-export type ReducerActions = SetLocale.Request | SetLocale.Result |
-    LoadCurrentLocale.Request | LoadCurrentLocale.Result;
+export { 
+    LoadCurrentLocaleRequestAction, loadCurrentLocaleRequest,loadCurrentLocaleSuccess, 
+    LoadCurrentLocaleErrorAction, loadCurrentLocaleFailure, LoadCurrentLocaleResult, LoadCurrentLocaleAction, setLocaleRequest, 
+    SetLocaleAction, setLocaleSuccess, setLocaleFailure, SetLocaleResult, SetLocaleRequestAction, LocaleAction
+};
```
rasmus-storjohann-PG: This is also messy, remove it and have other files import these directly from `/stores/locale/actions.ts`

src/sagas/locale.ts
```diff
@@ -3,32 +3,35 @@ import { takeLatest, call, put, ForkEffect, CallEffect, PutEffect } from 'redux-
 
 import * as constants from '../application/constants';
 import { LocaleInfoManager, saveCurrentLocaleCode, loadCurrentLocaleCode, reload } from '../locale';
-import { SetLocale, LoadCurrentLocale, setLocaleActions, loadCurrentLocaleActions } from '../stores/locale';
+import { LoadCurrentLocaleRequestAction,loadCurrentLocaleSuccess, loadCurrentLocaleFailure, LoadCurrentLocaleResult, SetLocaleRequestAction, SetLocaleResult, setLocaleSuccess, setLocaleFailure, } from '../stores/locale';
```
rasmus-storjohann-PG: Simplify to 
```ts
import * as actions from '../stores/locale/actions';
```
For simplicity you'll want to use the same or similar pattern elsewhere in this PR where you're importing the locale actions.

tomy-pg:
AC:

- [ ] when importing locale actions use import * as actions
- [ ] import actions directly from actions.ts - do not use import/export
- [ ] Remove helper functions from locale reducer 
- [ ] Failure instead of Error - be consistent in the naming 
- [ ] “Current” is redundant - 
- [ ] LoadCurrentLocaleResult not needed 
- [ ] Use term Save instead of Set 
- [ ] Tests - remove messages, loading variables from actions.ts - the messages are alwayys empty and loading flag is always false 
- [ ] Remove loading property - always false 
- [ ] Use linter

src/stores/__tests__/helpers/locale_helpers.ts
```diff
@@ -82,8 +82,8 @@ export class LocaleStoreBuilder {
         return this;
     }
 
-    withIsSet(isSet: boolean): LocaleStoreBuilder {
-        this.isSet = isSet;
+    withIsSet(isSaved: boolean): LocaleStoreBuilder {
```
rasmus-storjohann-PG: Rename function

src/stores/__tests__/locale.test.ts
```diff
@@ -122,33 +123,33 @@ describe('the reducer', () => {
         expect(theStore.fallback).toBe(undefined);
     });
 
-    it('when called with SET_LOCALE_REQUEST should return store with loading flag set', () => {
+    it('when called with SAVE_LOCALE_REQUEST should return store with loading flagAVE', () => {
```
rasmus-storjohann-PG: typo

Issue #757, opened 2019-08-01
Create all caps translations for explore category names


ian-lin-PG:
In the production app, we use the old translation system that allowed strings to be upperCase. In the current weblate implementation upperCase in \<Text\> does not work on text inside the \<Trans\> tag.

### Acceptance Criteria:
- [x] Create upperCase translations of the categories
- [x] Use the upperCase for the task detail pages

### How to QA:
1.  For French and Tagalog - Explore category names are capitalized
2. For Arabic, Korean, Punjabi, Chinese Simplified and Traditional, the Explore category names are shown in its respective locale translation. 

![image.png](https://images.zenhubusercontent.com/5d0be985d8501ed6b9d8c171/fc865639-72ad-47a0-a321-57df44c7fdf6)
![image.png](https://images.zenhubusercontent.com/5d0be985d8501ed6b9d8c171/f452f48e-d4d3-4341-9ee2-880379854183)

tomy-pg:
I will need to do the following:

For English, French and Tagalog:
- [ ] Create capitalized copy of each Explore category string 

For Arabic, Korean, Chinese(ZH, TW) and Punjabi:
- [ ] For capitalized msgid, map them to already existing msgstr

This will all be done via script.

jparlina:
Tested on Staging V1.4.0, server 1.3.9 on iOS:
1) Confirmed in EN, FR, Tagalog: Explore Category is all capitalized under each topic (see sample screenshots)

![7BF52E4D-50EE-4C81-B109-BDDDE5E4F399.jpeg](https://images.zenhubusercontent.com/5ceeccc646b329310b95bccd/cd585874-b4c3-4055-b9ee-b75f17fa4402)

![323FB2C1-9089-4966-B93E-0FCF4BC81D8E.jpeg](https://images.zenhubusercontent.com/5ceeccc646b329310b95bccd/e9d4454a-bff2-485d-9dba-21c22c3cb87a)

![7C063AD9-D226-4BA0-9585-DC60F418ED93.jpeg](https://images.zenhubusercontent.com/5ceeccc646b329310b95bccd/64cbc02a-a22f-4a3a-aa12-5a94f9a008d4)


2) For Punjabi, Korean, Chinese (ZH & TW), Arabic, the Explore category are translated in respective language.

PR #768, opened 2019-08-09
Issue #757 capitalize explore category names

tomy-pg:


src/components/topics/task_detail_content_component.tsx
```diff
@@ -72,7 +72,7 @@ const TaxonomyComponent = (props: Props): JSX.Element => (
         },
     ]}
     >
-        <Trans id={props.topic.exploreSection.name} />
+        <Trans id={props.topic.exploreSection.name.toLocaleUpperCase()} />
```
bensinjin: Does this actually work in an RN environment? Generally you'd pass it a locale right?

tomy-pg: I believe this works in an RN environment. You're correct, I am not using it the way it is intended because you **are** supposed to pass a locale. Changing now. Thanks for reviewing @bensinjin

bensinjin: Np. Maybe you can test it to confirm it works as expected, I'm curious myself.

Issue #759, opened 2019-08-02
Onboarding in Arabic: Invert swipe direction


rasmus-storjohann-PG:
In Arabic, the onboarding flow swipe direction should be inverted.

ACs:
* Swipe right for next screen
* Swipe left for previous screen

To test: 
* Start the app
* Select Arabic
* Navigate to the help screen, scroll to the bottom of the page, hit the button to delete all data for in the app, this is so that the onboarding flow will come up when the app starts
* Exit the app
* Start the app to see the onboarding flow in Arabic
* Confirm the swipe direction to move through the onboarding pages is correct as per the ACs above

jparlina:
Tested on Staging V1.4.0, server V1.3.9 on iOS:
Confirmed the onboarding in Arabic version: 
- Swipe LEFT for next screen
- Swipe RIGHT for previous screen

PR #769, opened 2019-08-09
Issue #759 Fix swipe direction for RTL languages and tidy up

bensinjin:


src/components/swipeable_content/swipeable_content_component.tsx
```diff
@@ -79,18 +88,26 @@ const NavigationDot = (props: { readonly currentIndex: number, readonly loopInde
...
 };
 
-const getNextIndex = (currentIndex: number, itemCount: number): number => (
+const getNextIndexForLeftSwipe = (currentIndex: number, itemCount: number): number => (
+    I18nManager.isRTL ? getDecrementedIndex(currentIndex) : getIncrementedIndex(currentIndex, itemCount)
+);
+
+const getNextIndexForRightSwipe = (currentIndex: number, itemCount: number): number => (
+    I18nManager.isRTL ? getIncrementedIndex(currentIndex, itemCount) : getDecrementedIndex(currentIndex)
+);
```
rasmus-storjohann-PG: This code is misleading: RTL doesn't mean that "next" is "index - 1", as this implies. Next is always +1. Better would be to rename `isLeftSwipe()` -> `isSwipeToNext()` and `isRightSwipe()` -> `isSwipeToPrevious()` and put the RTL logic inside those.

bensinjin: Yes, good point, fixed

src/components/swipeable_content/swipeable_content_component.tsx
```diff
@@ -88,26 +86,32 @@ const NavigationDot = (props: { readonly currentIndex: number, readonly loopInde
...
 
-const isLeftSwipe = (horizontalMovementValue: number): boolean => (
-    horizontalMovementValue < -1.0 * requiredSwipDistanceValue
-);
+const isSwipeToNext = (horizontalMovementValue: number): boolean => {
+    if (I18nManager.isRTL) {
+        return isRightSwipe(horizontalMovementValue);
+    }
+    return isLeftSwipe(horizontalMovementValue);
+};
```
rasmus-storjohann-PG: Nice, this is very clear to follow.

PR #777, opened 2019-08-15
Feature/issue 731 adjust frontend deploy process

tomy-pg:


bin/strings.sh
```diff
@@ -118,37 +118,44 @@ concat_en_source_files() {
...
 
-check_for_fuzzy_attribute() {
+check_for_duplicate_string_pattern() {
     locale=$1
     messages_po_file=$CLIENT_LOCALE_LOCATION/$locale/messages.po
 
-    #fuzzy is an attribute that in most cases indicates that a PO file needs revision
-    grep "fuzzy" $messages_po_file
-    prompt_manual_steps_if_po_file_needs_editing $locale
+    # '#-#-#-#' is the pattern generated when duplicated strings exist
```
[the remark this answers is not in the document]
tomy-pg: @ian-lin-PG  "#-#-#-#" is a pattern that is generated when msgcat finds a duplicate translation string existing in 2 or more of the PO files that are concatenated. In the past this has broken some of our JSX.

bin/strings.sh
```diff
@@ -118,37 +118,44 @@ concat_en_source_files() {
...
+    if [ "$locale" == "ar" ]; then 
+        tail -n +49 $messages_po_file | grep "#-#-#-#"
+        prompt_manual_steps_if_po_file_needs_editing $locale
+
+    else
+        tail -n +46 $messages_po_file | grep "#-#-#-#"
+        prompt_manual_steps_if_po_file_needs_editing $locale
```
rasmus-storjohann-PG: It's not clear that the return value of the grep call determines what `prompt_manual_steps_if_po_file_needs_editing()` does. Move the `if [ "$?" == "0" ]; then ` here to make it clear what's going on.

bin/strings.sh
```diff
@@ -121,18 +121,25 @@ validate_messages_po_files() {
...
+        check_for_duplicate_string_pattern $locale
     done
 }
 
 
-check_for_fuzzy_attribute() {
+check_for_duplicate_string_pattern() {
     locale=$1
-    messages_po_file=$CLIENT_LOCALE_LOCATION/$locale/messages.po
+    messages_po_file=$CLIENT_LOCALE_LOCATION/fr/messages.po
```
rasmus-storjohann-PG: That `fr` looks wrong, I guess it'll to away in a later commit.

bin/strings.sh
```diff
@@ -190,7 +190,7 @@ elif [ "$1" == "--clean" ]; then
 elif [ "$1" == "--combine-pos" ]; then
     combine_po_files
     validate_messages_po_files
-    
+
```
rasmus-storjohann-PG: Feel free to squash this kind of commit with the previous one ... it's just changes that you forgot to make and it's the same "thought" as the previous commit.

bin/strings.sh
```diff
@@ -118,37 +118,44 @@ concat_en_source_files() {
...
-    then
+    if [ "$?" == "0" ]; then 
         echo
-        echo "Warning: the $1 messages.po file requires editing"
+        echo "Warning: locale/$1/messages.po requires editing"
         echo 
         echo "Manual steps:"
         echo
-        echo "review this file and remove duplicate strings that look like the following:"
+        echo " review this file and remove duplicate strings that look like the following:"
```
rasmus-storjohann-PG: Update the message to instruct the user to fix the problem at the source of truth (pot files), locate the duplicate strings and remove them.

bin/strings.sh
tomy-pg: Done. This makes the code much more readable.

bin/strings.sh
tomy-pg: Squashed.

bin/strings.sh
tomy-pg: Done.

bin/strings.sh
```diff
@@ -133,30 +133,31 @@ check_for_duplicate_string_pattern() {
     # '#-#-#-#' is the pattern generated when duplicated strings exist
     if [ "$locale" == "ar" ]; then 
         tail -n +49 $messages_po_file | grep "#-#-#-#"
-        prompt_manual_steps_if_po_file_needs_editing $locale
-
+        if [ "$?" == "0" ]; then 
+        prompt_manual_steps_if_duplicate_strings_exist $locale
```
rasmus-storjohann-PG: * `if_duplicate_strings` in the function name implies that the test is done inside the function, which it's not. 
* Also, I would pass in `$messages_po_file` rather than `$locale`, that way `prompt_manual_steps_if_duplicate_strings_exist()` doesn't need to know about the details of how to get from the locale to the strings file.
* Also, indentation.

PR #787, opened 2019-08-23
Feature/backend issue 423 find search provider

rasmus-storjohann-PG:


rasmus-storjohann-PG:
@ian-lin-PG @tomy-pg can I get some reviews on this PR today?

src/components/search/auto_complete.tsx
```diff
@@ -0,0 +1,44 @@
...
+export interface Props {
+    readonly hits: ReadonlyArray<Suggestion>;
+    readonly currentRefinement: string;
+}
+
+export interface Actions {
+    readonly refine: (value?: string) => void;
+}
+
+export const AutoCompleteComponent = (props: Props & Actions): JSX.Element => (
```
tomy-pg: Why isn't a Partial<T> required here like in the InfiniteHitsComponent?

tomy-pg: In other components in our codebase we are doing Intersection types as follows:

`src/components/topics/task_detail_component`
`type Props = TaskDetailProps & TaskDetailActions`

Should `(props: Props & Actions)` be changed to match this?

rasmus-storjohann-PG: exported types are named like `TaskDetailProps`, non-exported types can have simpler names like `Props` since they have a smaller scope and don't need to make sense in a bunch of different contexts.

rasmus-storjohann-PG: Good question, don't have an answer

PR #793, opened 2019-08-29
Feature/issue 767 add qa language strings

tomy-pg:
Block merging off this until @ArtiBPG has another look at the language strings on Weblate.


rasmus-storjohann-PG:
QA: Checked that the strings are there in all languages. In languages that don't support bold fonts (Chinese, Arabic, Korean), the "contact us" link in the about page doesn't stand out like it does in English, but that's probably not something we can easily fix.

Issue #636, opened 2019-05-22
Text Direction & Alignment Issues when switching from Non-Arabic to Arabic and vice versa


jparlina:
Issues on both versions 1.0.13 and 1.0.17:
- seems to be an issue on iOS but not on Android 
- issue can't be replicated every time (not consistent) when switching between non-Arabic to Arabic, or vice versa 
- happens after a few times switching between Arabic and non-Arabic (use cases: some settlement workers may have to switch between languages a few times to help different clients; some Arabic speakers may switch back and forth - e.g. going to the service location and need to show what the app says to the service provider in English).

You can tell whether the alignment is correct based on the position of the dropdown arrow for language selection on the Start page. Examples:
1. When switching from Arabic to English, if you see the dropdown arrow is in the wrong position (e.g. on the screen shot below the dropdown arrow is on the left side of the word "English")

![IMG_0658](https://user-images.githubusercontent.com/48343061/58200109-d73be780-7c86-11e9-8d21-10d4e3df06f3.JPG)

1a. when you just continue by clicking "Start" button, you'll see the alignment on all pages after that will be incorrect (see screenshots below):

![IMG_0660](https://user-images.githubusercontent.com/48343061/58200147-eb7fe480-7c86-11e9-8358-89a9c1b43d8f.JPG)

![IMG_0662](https://user-images.githubusercontent.com/48343061/58200355-795bcf80-7c87-11e9-8c30-5ff3c14ff47a.JPG)

![IMG_0661](https://user-images.githubusercontent.com/48343061/58200342-73fe8500-7c87-11e9-9b5e-5f067a926405.JPG)

1b. If you see on the Start page the position of the dropdown arrow is incorrect, but instead of just clicking "Start", you click on the dropdown arrow and select the same language again (in this example English, then you'll see the arrow then moves to the correct position and the pages inside will be fine when you click "Start". 

2. Same issues happen sometimes when you switch from non-Arabic to Arabic. If you see the dropdown arrow for the language selection is incorrect (e.g. on the screen shot below the dropdown arrow is on the right side of "Arabic"):

![IMG_0656](https://user-images.githubusercontent.com/48343061/58201256-795ccf00-7c89-11e9-969e-1f9128f062ae.JPG)

2a. Then you just continue by clicking "Start" button, you'll see the alignment on all pages after that will be incorrect (see screenshots below):

[...]

PR #796, opened 2019-08-30
Issue #636 Refactor RTL change logic

ian-lin-PG:


src/sagas/locale.ts
```diff
@@ -10,14 +10,26 @@ export function* watchSaveLocale(): IterableIterator<ForkEffect> {
...
+    yield takeLatest(constants.SAVE_LOCALE_SUCCESS, applyRTLChange);
+}
+
+export function* watchLoadLocaleSuccess(): IterableIterator<ForkEffect> {
+    yield takeLatest(constants.LOAD_CURRENT_LOCALE_SUCCESS, applyRTLChange);
+}
+
+export function* applyRTLChange(action: actions.SaveLocaleSuccessAction | actions.LoadLocaleSuccessAction): Iterator<CallEffect> {
+    const localeCode = action.payload.localeCode;
+    if (yield call(needsTextDirectionChange, localeCode)) {
```
rasmus-storjohann-PG: `if(action.needsTextDirectionChange)` here is the important part of this fix. That should be simple (conceptually, it may touch quite a bit of code) in the `watchSaveLocaleSuccess` case. In the `watchLoadLocaleSuccess` it might be a bit tricky ... need to understand whether `watchLoadLocaleSuccess` is really needed.

src/sagas/index.ts
```diff
@@ -22,7 +22,9 @@ export function buildSaga(): ApplicationSaga {
 export function runSaga(middleware: SagaMiddleware<object>): void {
     middleware.run(watchLoadFonts);
     middleware.run(watchLoadLocale);
+    middleware.run(watchLoadLocaleSuccess);
     middleware.run(watchSaveLocale);
+    middleware.run(watchSaveLocaleSuccess);
```
rasmus-storjohann-PG: Does the order of these lines matter? If so, that would be a bit smelly.

PR #804, opened 2019-09-10
Issue#711 Partially localize the service listing page

ian-lin-PG:


package.json
```diff
@@ -12,6 +12,7 @@
         "@types/query-string": "^6.1.0",
         "@types/ramda": "^0.25.38",
         "@types/react": "^16.8.23",
+        "babel-core": "^7.0.0-bridge.0",
```
tomy-pg: @rasmus-storjohann-PG this package had to be re-added because we were receiving an error while running
`yarn extract-strings`. The error message stated that @babel/core could not be found. I believe babel-core: 7.0.0-bridge.0 is needed because it was made to ease the transition for libraries, in this case lingui, to go from Babel 6 to Babel 7.

rasmus-storjohann-PG: I added a commit to check that extract strings as part of the travis build...

rasmus-storjohann-PG: Yeah, that didn't work, need a .env for Travis, not sure if we want to go there.

src/components/maps_application_popup/maps_application_popup_component.tsx
```diff
@@ -25,9 +25,7 @@ export const MapsApplicationPopupComponent: React.StatelessComponent<MapsApplica
 
         const text = <Text style={textStyles.button} uppercase={false}><Trans>Open in maps</Trans></Text>;
 
-        const flipLeftRightDirection = I18nManager.isRTL;
-
-        const button = flipLeftRightDirection ?
+        const button = I18nManager.isRTL ?
             <Button onPress={onMapsButtonPress(props)} iconRight style={applicationStyles.tealButton} >
                 {text}{icon}
             </Button> :
```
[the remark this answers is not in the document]
ian-lin-PG: The RTL logic was removed in a subsequent commit.

PR #816, opened 2019-09-17
Issue#812 Enforce service name and service description to be LTR in Arabic service listing

ian-lin-PG:


ian-lin-PG:
Adding the picture here for context.
image

Issue #817, opened 2019-09-17
Incorrect Line Break: Korean - Onboarding Page 1


jparlina:
Describe the issue and expected behaviours. If this is a bug, describe the steps to reproduce the problem, what you observe, and what you expected to observe.

### Acceptance Criteria 
- [ ]  Ensure words are not split into two lines  - confirm as per [Weblate ](https://translate.peacegeeks.org/translate/ui-strings/jsx_strings/ko/?checksum=&offset=&q=onboarding&search=substring&source=on&target=on&location=on&type=all&ignored=False&date=&only_user=&exclude_user=)

### Steps for QA: 
1.  Ensure all user data is deleted
2. Change locale to KO
3. Check first slide and ensure that words in the string "Arrival Advisor가 새로운 커뮤니티
 적응을 위한 믿음직한 가이드가 되어드립니다." are cut off. 

![IMG_2424](https://user-images.githubusercontent.com/48343061/65084760-1d693500-d961-11e9-9a52-64b4bcdb2cdf.JPG)

tomy-pg:
These are the changes. I forced a linebreak to make the cut-off consistent across all devices. 
@ArtiBPG @jparlina @cherrielam please view. 
![Screenshot_20190917-170159_Expo](https://user-images.githubusercontent.com/47794846/65088213-5909fc00-d96d-11e9-9477-eee0aa06dbc8.jpg)
![Screenshot_20190917-170203_Expo](https://user-images.githubusercontent.com/47794846/65088215-5909fc00-d96d-11e9-8071-9d4d33a5d209.jpg)
![Screenshot_20190917-170213_Expo](https://user-images.githubusercontent.com/47794846/65088216-5909fc00-d96d-11e9-93ed-2a27a63b8f72.jpg)
![Screenshot_20190917-170217_Expo](https://user-images.githubusercontent.com/47794846/65088217-59a29280-d96d-11e9-9705-7e8e38a478e7.jpg)

jparlina:
@tomy-pg  See our Korean translator's suggestions below.


From: Ran Kang <ran.kanglee@gmail.com>
Date: Fri, Sep 20, 2019 at 10:59 AM
Subject: Re: PeaceGeeks: Language QA & Additional Translations
To: Jean Parlina <jean@peacegeeks.org>
Cc: Arti Bhakta <arti@peacegeeks.org>


Hi Jean, 

[...]

PR #818, opened 2019-09-17
Issue #817 Incorrect Line Break in KO onboarding 

tomy-pg:


tomy-pg:
Do not merge yet. The breaks in the word need to be verified for someone who speaks KO. It is unclear whether or not the break/space between characters are problematic.

PR #855, opened 2019-10-05
added Italian locale support

backciro:


rasmus-storjohann-PG:
Hey @backciro , thanks so much for wanting to help with our project. Can you tell me some more about how you envisage the tool being made available in Italian. One big question for me is that we have a lot of content (articles) that would need to be in Italian for this to make sense. We haven't been planning to make those translations. Do you have any thoughts on how that problem would be solved?

rasmus-storjohann-PG:
Haven't heard back from @backciro , closing the PR.

PR #870, opened 2019-10-15
Issue pg-irc/pathways-backend#291 Keep location data consistent by swapping back lat/lon

ian-lin-PG:


src/stores/services/validation.ts
```diff
@@ -77,8 +77,8 @@ export const serviceFromValidatedJSON = (data: ValidatedServiceAtLocationJSON):
     return {
         id: data.service.id,
         // These values come in the wrong order from the server
```
rasmus-storjohann-PG: remove the comment

ian-lin-PG: Squashed with previous commit.

PR #879, opened 2019-10-18
Feature/issue 455 submodules

rasmus-storjohann-PG:


bensinjin:
Looks good. (This article)[https://docs.travis-ci.com/user/private-dependencies/] will probably help with the Travis problem we're seeing.

rasmus-storjohann-PG:
Thanks @bensinjin , I went away from using a submodule for the content ... don't want to break the repo for users ourside our org, but still want the content to only be available to us, so the manual approach is still preferred, I think. ui-strings handling should be much improved though.

rasmus-storjohann-PG:
It probably makes sense to move some of the CSV files from the content to the backend repo, that way we will update the content repo less often, which will be nice since it's a little cumbersome to make changes there with the versioning etc.

rasmus-storjohann-PG:
@bensinjin @ian-lin-PG and @tomy-pg feel free to have a look at this PR, I think it's ready for review.

PR #880, opened 2019-10-23
Feature/issue 839 search improvements

rasmus-storjohann-PG:


src/components/search/search_input_component.tsx
```diff
@@ -26,44 +28,58 @@ export const SearchInputComponent = (props: Props & Actions): JSX.Element => {
...
+    </I18n>;
 };
 
-interface IconProps {
-    readonly name: string;
+interface SearchTextInputProps {
+    readonly iconName: string;
+    readonly value: string;
+    readonly localizedPlaceholder: string;
+    // tslint:disable-next-line:no-mixed-interface
```
bensinjin: I think we probably should start pruning our `tslint.json` a bit. I find this rule more annoying than useful personally.

rasmus-storjohann-PG: Yeah, we can live without this one, I'll remove it in a later PR

PR #881, opened 2019-10-24
WIP: Feature/issue 840 bookmark services

tomy-pg:


src/selectors/services/types.ts
```diff
@@ -11,8 +11,13 @@ export interface LoadingSelectorTopicServices {
...
+    readonly errorMessageType?: Errors;
     readonly type: 'ServicesForTopic:Error';
 }
 
-export type SelectorTopicServices = ValidSelectorTopicServices | LoadingSelectorTopicServices | ErrorSelectorTopicServices;
+export interface EmptySelectorTopicServices {
+    readonly type: 'ServicesForTopic:Empty';
+}
+
+export type SelectorTopicServices = ValidSelectorTopicServices | LoadingSelectorTopicServices |
```
bensinjin: So I think this type is meant to reflect a response from our API for topic services. It makes sense that the response would create objects in Redux that reflect a "Valid", "Loading" or "Error" state but I don't think you should be using this type to accommodate your needs. All you need is an array of Services that may be empty.

src/stores/services/actions.ts
```diff
@@ -31,4 +43,14 @@ export const buildTopicServicesErrorAction = (
     topicId: TopicId,
     errorMessageType: Errors) => (
         helpers.makeAction(constants.LOAD_SERVICES_FAILURE, { topicId, errorMessageType })
-    );
\ No newline at end of file
+    );
+
+// tslint:disable-next-line:typedef
+export const addToSavedList = (service: Service) => {
```
bensinjin: This name could be better.

src/stores/services/types.ts
```diff
@@ -62,4 +63,5 @@ export interface ServiceMap {
 export interface ServiceStore {
     readonly services: ServiceMap;
     readonly servicesByTopic: ServicesForAllTopics;
+    readonly savedServices: SelectorTopicServices;
```
bensinjin: Why is this in the services store as well as the user data store?

Issue #824, opened 2019-09-19
Tracking Errors as events


ArtiBPG:
We want to track how often service listings do not load and users get the error messages when they click on "find services near me". 
Error message and scenarios are described in [this issue](https://github.com/pg-irc/pathways-frontend/issues/704) 

Question: Could we do this by treating each error message as an event and tracking through Google Analytics?

### Acceptance Criteria 
- [ ]  On a weekly basis we want to track if users are able to load service listing when they click the "find services near me" button. 
- [ ] If the service listings do not load, we want to know why: no wifi or data, server issue, etc as described in [this issue](https://github.com/pg-irc/pathways-frontend/issues/704)

### Steps for QA if GA approach is chosen: 
1.  Load service listings for 10 topics with wifi and data turned off to expose error screens
2. Test number of events on Google Analytics 

### Steps for QA using Sentry:
1. Create a Sentry account for testing, once created, create a Sentry project

2. Set the following values in your `.env` file:
```
SENTRY_AUTH_TOKEN=your token (see below for how to find this)
SENTRY_DSN=your dsn (see below for how to find this)
SENTRY_ENABLE_IN_DEV=true
SENTRY_DEBUG=true
```  
In Sentry find your auth token at: `/settings/account/api/auth-tokens/` and your DSN at: `/settings/{your-project}/projects/{your-project}/keys/`. You can also just search "dsn" for example from the search bar on the `/settings` page.

3. Start expo with -c

4. Test the errors appear in Sentry (this can sometimes take a minute or two). These are the errors triggered from the topic services screen and how to trigger them:
- "No location": test this by turning off your location
- "Check location": test this on Android (no obvious iOS test we know of) by turning off location high accuracy mode
- "Server error": test this by triggering a bad server response, I go to `src/api/index.ts` and change the endpoint at line ~33 so we're hitting something invalid
- "Offline": this needs to be tested on a physical device by turning off your WIFI

bensinjin:
This should be likely be done with Sentry which we already have in place. TODO:

- Test existing Sentry implementation for this use case

rasmus-storjohann-PG:
QA on android (read device):

- [ ] "No location": test this by turning off your location
- [ ] "Check location": test this on Android (no obvious iOS test we know of) by turning off location high accuracy mode
- [ ] "Server error": test this by triggering a bad server response, I go to src/api/index.ts and change the endpoint at line ~33 so we're hitting something invalid
[...]

PR #884, opened 2019-10-30
Issue #824 Track services errors as messages in Sentry

bensinjin:
This PR allows us to be notified when error screens are displayed from the services screen. Sentry does a good job of collecting other various metadata like for instance OS so we don't need to worry about sending that data along in the message. Here's a screenshot of what this looks like in my test Sentry instance:

Screen Shot 2019-10-30 at 9 58 55 AM


App.js
```diff
@@ -1,7 +1,12 @@
 import { Application } from './lib/application';
-import Sentry from 'sentry-expo';
+import * as Sentry from 'sentry-expo';
 
-// DSN for PeaceGeek's pathways project.
-Sentry.config('https://de23d08a2e2b44a49849c045c2a6fc0c@sentry.io/256284').install();
+Sentry.init({
+    // DSN for PeaceGeek's pathways project
+    dsn: 'https://de23d08a2e2b44a49849c045c2a6fc0c@sentry.io/256284',
```
rasmus-storjohann-PG: Should this be checked in? I think all the values in this config, including debug, should be defined in .env

src/components/error_screens/ErrorScreenSwitcherComponent.tsx
```diff
@@ -1,3 +1,4 @@
+// tslint:disable:no-expression-statement
```
rasmus-storjohann-PG: Why is this change needed?

src/components/services/service_list_component.tsx
```diff
@@ -18,6 +19,9 @@ import { ErrorScreenSwitcherComponent } from '../error_screens/ErrorScreenSwitch
 import { Errors } from '../../validation/errors/types';
 import { EmptyListComponent } from '../empty_component/empty_list_component';
 import { LatLong } from '../../validation/search/types';
+import { getSentryMessageForError } from '../../validation/errors/sentry_messages';
+
+const SENTRY_ERROR_CONTEXT = 'ServicesListing';
```
rasmus-storjohann-PG: We have a `constants` file for this.

src/components/services/service_list_component.tsx
```diff
@@ -53,10 +59,12 @@ export const ServiceListComponent = (props: Props): JSX.Element => {
...
+        const errorType = props.topicServicesOrError.errorMessageType;
+        Sentry.captureMessage(getSentryMessageForError(errorType, SENTRY_ERROR_CONTEXT));
         return (
             <ErrorScreenSwitcherComponent
                 refreshScreen={refreshServicesData}
-                errorType={props.topicServicesOrError.errorMessageType}
+                errorType={errorType}
                 header={
                     <ServiceListHeaderComponent title={props.topic.title} />
                 }
```
rasmus-storjohann-PG: There is a lot of code duplication here, you can DRY this up I think.

App.js
bensinjin: Fixed

src/components/error_screens/ErrorScreenSwitcherComponent.tsx
bensinjin: Removed

src/components/services/service_list_component.tsx
bensinjin: Fixed

src/components/services/service_list_component.tsx
bensinjin: Fixed

PR #895, opened 2019-11-14
Feature/issue 813 client sends notification token to server

rasmus-storjohann-PG:


src/api/index.ts
```diff
@@ -73,3 +74,21 @@ async function createAPIResponse(response: Response): Promise<APIResponse> {
...
+
+export async function putPushNotificationToken(token: string, locale: Locale): Promise<APIResponse> {
+    const url = createPushNotificationTokenUrl(baseUrl, token);
+    const response = await fetch(url,
+        {
+            method: 'PUT',
+            headers: [
+                ['Content-Type', 'application/json'],
+            ],
+            body: JSON.stringify({ 'locale': locale.code }),
```
bensinjin: A bit of an odd syntax for an object - can probably drops the single quotes around locale.

src/application/store.ts
```diff
@@ -6,6 +6,7 @@ import { runSaga, ApplicationSaga } from '../sagas';
 import { loadFontsActions } from '../stores/fonts';
 import { loadLocaleRequest } from '../stores/locale/actions';
 import { UserDataPersistence } from '../stores/user_data';
+import { request as requestPushNotificationPost } from '../sagas/post_push_notification_token';
```
bensinjin: Not sure the fact that you're PUTing (not POSTing) is all that important in naming these.

bensinjin:
> As it stands the permission step on iOS fails - we get a status of "undetermined". We'll need to look into this.

So from the docs:

>iOS and Android simulators cannot receive push notifications. To test them out you will need to use a real-life device. Additionally, when calling Permissions.askAsync on the simulator, it will resolve immediately with "undetermined" as the status, regardless of whether you choose to allow or not.

We do still see the "undetermined" when running on a physical iOS device through the Expo app but I'd be tempted to test this further with a staging build.

ian-lin-PG:
> > As it stands the permission step on iOS fails - we get a status of "undetermined". We'll need to look into this.
> 
> So from the docs:
> 
> > iOS and Android simulators cannot receive push notifications. To test them out you will need to use a real-life device. Additionally, when calling Permissions.askAsync on the simulator, it will resolve immediately with "undetermined" as the status, regardless of whether you choose to allow or not.
> 
> We do still see the "undetermined" when running on a physical iOS device through the Expo app but I'd be tempted to test this further with a staging build.

We also have the debug build that's available.

Issue #893, opened 2019-11-13
Improve Search Modal performance


ian-lin-PG:
Right now the search components renders three sub components all at once resulting in unnecessary rendering that contributes to slower responses. Use branching logic as appropriate and render the three sub components as needed.

### Acceptance Criteria 
- [ ]  Search component only renders one of `SearchInputModal` or `LocationInputModal` or the Search screen component

### Steps for QA: 
1.  Open the Search screen
2. Click on one of the search input modal, the modal should open faster than before.

rasmus-storjohann-PG:
We tried a fix to this problem, which improved the speed of the app but broke the functionality. Putting it back in triage for now.

ArtiBPG:
closing as we have an updated design #918

PR #905, opened 2019-11-21
Revert "Issue #893 Put modals and rest of search components into swit…

rasmus-storjohann-PG:
…ch statements"

This reverts commit c5c76f6646a9ecefed3b165226727dc6ac390274.


rasmus-storjohann-PG:
So @ian-lin-PG unfortunately this change broke search. It's pretty fragile, and finding a better solution will take some time.

PR #910, opened 2019-11-23
Feature/issue 908 phone numbers in search results

rasmus-storjohann-PG:


rasmus-storjohann-PG:
@bensinjin @tomy-pg @ian-lin-PG for review.

PR #912, opened 2019-11-26
Feature/issue 862 & 863 UI Tweaks

ian-lin-PG:


src/components/help/help_component.tsx
```diff
@@ -26,17 +26,26 @@ interface HelpContact {
 const fixture: ReadonlyArray<HelpContact> = [
     {
         title: <Trans>Information on emergency services (9-1-1)</Trans>,
-        subTitle: <Trans>Police, fire and medical emergencies</Trans>,
+        subTitle: (
+            <Text note numberOfLines={1} style={{ textAlign: 'left', fontFamily: getNormalFontFamily() }}>
```
bensinjin: Do we want to set numberOfLines here? We probably just want the text to wrap as you cannot click into these items to view all the text. Also, our "default" text style should just be `fontStyles.paragraphStyle` whenever possible for consistency, I think that applies here.

src/components/services/service_list_component.tsx
```diff
@@ -42,6 +42,10 @@ export const ServiceListComponent = (props: Props): JSX.Element => {
     const onlineStatus = useOnlineStatus();
     const refreshServicesData = useRequestDataIfOnlineReturnRefreshDataCallback(props.dispatchServicesRequest);
     const hasNoErrors = onlineStatus !== OnlineStatus.Offline && !isError(props.topicServicesOrError);
+    const emptyMessage = (
+        <Text note numberOfLines={1} style={{ textAlign: 'left', fontFamily: getNormalFontFamily() }}>
```
bensinjin: This is also used for the topic list (ask `task_list_component`) so we should be updating the `<Text>` component in the `<EmptyListComponent />` not adding a text component here.

src/components/help/help_component.tsx
ian-lin-PG: Yeah the `numberOfLines` is definitely redundant considering how short all the strings are that are applied. I will also try and see if the `textAlign: 'left'` is necessary as it seems to be implicit for most text components. I think the main crust of this change was the `note` and `fontFamily`, will also try `fontStyles.paragraphStyle` as comparison.

bensinjin: So what I actually meant was `numberOfLines` will add an ellipses if the text runs too long, which it often does in other languages and really only makes sense if it's say teaser text for a longer article.

src/components/services/service_list_component.tsx
ian-lin-PG: Good catch, I will add the `fontFamily` in the `EmptyListComponent` and remove the `Text` wrapper tag here.

src/components/help/help_component.tsx
ian-lin-PG: So the `numberOfLines` properties was copied over from the teaser text that is being served as the basis for this change. Out of all the locales only 1 translation of tagalog goes into the second line, but it's the same behaviour on production so removing the `numberOfLines`. `textAlignLeft: 'left'` also did not make any noticeable difference. `fontStyles.paragraphStyle` overrides the `note` property which gives text the left grey color with smaller fonts, so will change the final style to just `note style = {{fontFamily = genNormalFontFamily}}`. The `note` property sits outside of the style property, so won't make a new style for these kind of texts.

Issue #841, opened 2019-10-01
Services on the bookmarks screen


rasmus-storjohann-PG:
As a user who has bookmarked a service (#840 )  and a few topics, I want to see my bookmarked topics and services so that I can read or action them at a later time. 

Currently the bookmark screen shows all the topics that are bookmarked. With the ability to bookmark services, the same screen will show both bookmarked topics and services. The services will be listed below the topics in a separate section titled "services". The bookmarked topics will be in a section titled " bookmarks".

ACs:
- [x] Surface bookmarked services on the bookmark page: Below bookmarked topics (if any), display bookmarked services (if any) using the same service detail component as used elsewhere.
- [ ] If the list of bookmarked topics or services is more than 3, there should be a "load more" button to expose the rest of the bookmarked topics or services. 
- [x] Use the existing service list item component and enhance it to have the bookmark icon to be consistent with the bookmark button on the topic list item component.

Mocks: This is a dupe of #891 see that story for mocks.

Steps for QA:

This issue is the UI implementation of Issue #840. As such follow the Steps for QA from that issue as well as:

Ensure bookmarked service list changes from direction based on locale: 
1. Change the language from English to Arabic
2. Navigate to the bookmarks page and click on the services tab: 
- [ ] Ensure the bookmarks page follows the RTL text direction.

tomy-pg:
There was some confusion about the "load more button" Acceptance Criteria as I did not see this as part of the mocks. After discussion with @aron-chen-pg , it was decided that this AC needs to be further researched. As such, I did not include it as part of this PR.

ArtiBPG:
@aron-chen-pg either we have a load more button after X number of results or we have an infinite list- I see only two options. Please pick one so we can move on with this issue.

aron-chen-pg:
Unless it greatly affects our app performance (even with few saved items), we don’t have the need to implement the 'load more' button for saved bookmarks.

tomy-pg:
@ArtiBPG and @aron-chen-pg : I tested this by adding about 100+ services. This did not seem to affect the performance of the app. There is a slight delay when you unbookmark a service but this delay time appears to be consistent irregardless of how many services you have saved.

ArtiBPG:
OK then we will leave it as is, without the load more button.

Issue #840, opened 2019-10-01
Bookmark services


rasmus-storjohann-PG:
As a user who has found a relevant service via the search feature, I want to be able to save the service information so that I can contact or visit the service at a later time.

ACs:
- [x] Build plumbing to persist a service to permanent storage on the client. Follow existing pattern for persistence of chosen answers in `selectUserDataForLocalPersistence()`. Services will be stored in their entirety (not just the ids), so that will be different from the current patterns. Bookmarked services will also be loaded from local memory at app startup, see `UserDataPersistence.loadSuccess()` and how `LOAD_USER_DATA_SUCCESS` is used in several reducers to take data from persistent storage and place it in their respective stores.
- [x] Bookmark button appears on service list item component. Button is filled if the service is currently bookmarked, outline if not, consistent with topic bookmarks. Clicking the button causes the service to be added to or removed from local persistent storage. This is done in a saga that is watching for the action to change the bookmarked state of any service, see `watchUserStateChangesToSaveUserData()`.

Steps for QA:
Services can be bookmarked from topic service list and search service list: 
1. Generate a list of services by topic
2. Click on the bookmark icon
3. Navigate to the bookmarks page and click on the Services tab
- [x] Ensure the service(s) you bookmarked are there

Bookmarked services are available in the app even while the device is offline:
1.  Set the phone to airplane mode
2. Shut down the app 
3. Restart the app
4. Go to the bookmarks page and click on the Services tab
- [x] Ensure that the bookmarked service appears with all its data.

Scope:
* Bookmarking of organizations is OUT OF SCOPE. (Some of the mocks imply that this is a thing.)

ArtiBPG:
@jparlina - tested on Android, please test on iPhone

jparlina:
Tested on Staging V 1.15.0, Server V1.3.21 on iPhone 7s from the HiVE - PASSED

Services can be bookmarked from topic service list and search service list:
1. - Generate a list of services by topic
2. - Click on the bookmark icon
3. - Navigate to the bookmarks page and click on the Services tab
- [x] Ensure the service(s) you bookmarked are there

Bookmarked services are available in the app even while the device is offline:
1. Set the phone to airplane mode
2. Shut down the app
3. Restart the app
4. Go to the bookmarks page and click on the Services tab
- [x] Ensure that the bookmarked service appears with all its data.

PR #917, opened 2019-12-03
Feature/issue 840 bookmark services

tomy-pg:


src/validation/services/persisted_services/__tests__/persisted_services_schema.test.ts
```diff
@@ -0,0 +1,86 @@
...
+        };
+        const anAddress = {
+            type: aString(),
+            id: aNumber(),
+            address: aString(),
+            city: aString(),
+            stateProvince: aString(),
+            postalCode: aString(),
+            country: aString(),
+        };
```
ian-lin-PG: Somehow I think this might already exist. If not create an helper and see if existing test case can make use of this helper. Same goes for `aPhoneNumber` and `aLatLong`.

src/components/search/infinite_hits_component.tsx
```diff
@@ -9,15 +9,26 @@ import { SearchListSeparator } from './separators';
 import { ServiceListItemComponent } from '../services/service_list_item_component';
 import { validateServiceSearchResponse } from '../../validation/search';
 import { toHumanServiceData } from '../../validation/search/to_human_service_data';
+import { Id } from '../../stores/services';
```
ian-lin-PG: Somehow I think `Id` might be too generic, rename to `serviceId`?

src/components/bookmarks/bookmarks_connected_component.tsx
```diff
@@ -0,0 +1,32 @@
...
+export type ServiceDispatchActions = AddServiceToSavedListAction | RemoveServiceFromSavedListAction;
+export type ListActions = TaskListActions & ServiceListItemActions;
+export type TopicDispatchActions = AddToSavedListAction | RemoveFromSavedListAction;
+type DispatchActions = ServiceDispatchActions | TopicDispatchActions;
+
+const mapDispatchToProps = (dispatch: Dispatch<DispatchActions>): ListActions => ({
+    addServiceToSavedList: (service: HumanServiceData): AddServiceToSavedListAction => dispatch(addServiceToSavedListAction(service)),
+    removeServiceFromSavedList: (service: HumanServiceData): RemoveServiceFromSavedListAction => dispatch(removeServiceFromSavedListAction(service)),
+    addToSavedList: (topicId: Id): AddToSavedListAction => dispatch(addToSavedList(topicId)),
+    removeFromSavedList: (topicId: Id): RemoveFromSavedListAction => dispatch(removeFromSavedList(topicId)),
```
ian-lin-PG: If you give more explicit names to service bookmarks, makes sense to do the same for topic bookmarks as well.

tomy-pg: I hear you on this one. The reason why I did not was this change will touch a whole lot of files that are not part of the scope of this PR. I will put a commit up with this refactor and we can decide if we want to keep the commit or if we want to make this refactor as part of a _Tidy Friday_ type issue.

src/components/search/infinite_hits_component.tsx
tomy-pg: There is no other type that utilizes `Id` in this file. I think the instances where we do use an alias like
`Id as ServiceId` is where there are multiple types in a file that use an Id type. 

IE:
```
import { Id as LearnId } from '../stores/explore';
import { Id as TopicId } from '../stores/topics';
```

src/sagas/user_data.ts
```diff
@@ -59,6 +60,10 @@ export function* loadUserData(): LoadActions {
     try {
         const serializedUserData = yield call(loadUserDataAsync);
         const userData = deserializeUserData(serializedUserData);
+        const validatedSavedServices = validateServiceMap(userData.savedServices);
```
rasmus-storjohann-PG: So do you not find it a bit funny that we're just validating part of the `userData` here?

src/components/bookmark_button/bookmark_button_component.tsx
```diff
@@ -2,15 +2,19 @@ import React from 'react';
...
 import { values } from '../../application/styles';
+import { AddServiceToSavedListAction, RemoveServiceFromSavedListAction } from '../../stores/services/actions';
 
 export interface BookmarkButtonProps {
     readonly isBookmarked: boolean;
     readonly textColor: string;
 }
 
+export type AddBookmarkAction = AddToSavedListAction | AddServiceToSavedListAction;
+export type RemoveBookmarkAction = RemoveFromSavedListAction | RemoveServiceFromSavedListAction;
```
rasmus-storjohann-PG: I like the "add bookmark" naming more than "save to list" naming. Rename the bookmark topic actions to remove ambiguity.

src/components/search/infinite_hits_component.tsx
```diff
@@ -9,15 +9,26 @@ import { SearchListSeparator } from './separators';
...
     // tslint:disable-next-line:no-any
     readonly hits: ReadonlyArray<any>;
     readonly hasMore: boolean;
     readonly refine: (searchTerms?: string) => string;
+    readonly savedServicesIds: ReadonlyArray<Id>;
 }
 
+export interface InfiniteHitsActions {
+    readonly addServiceToSavedList: (service: HumanServiceData) => AddServiceToSavedListAction;
+    readonly removeServiceFromSavedList: (service: HumanServiceData) => RemoveServiceFromSavedListAction;
```
rasmus-storjohann-PG: rename to add/remove bookmark

src/components/bookmarks/bookmarks_connected_component.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { selectSavedTopics } from '../../selectors/topics/select_saved_topics';
+import { selectSavedServices } from '../../selectors/services/selected_saved_services';
+import { AddServiceToSavedListAction, RemoveServiceFromSavedListAction, addServiceToSavedListAction, removeServiceFromSavedListAction } from '../../stores/services/actions';
+import { ServiceListItemActions } from '../services/service_list_item_component';
+import { HumanServiceData } from '../../validation/services/types';
+import { getSavedServicesIds } from '../../selectors/services/get_saved_services_ids';
+
+const mapStateToProps = (store: Store): BookmarksProps => ({
+    bookmarkedServices: selectSavedServices(store),
+    savedServicesIds: getSavedServicesIds(store),
```
rasmus-storjohann-PG: Why do you need both of these, they seem to both represent the same information. The `bookmarkedServices: selectSavedServices(store),` line looks correct, the other one looks wrong.

src/components/bookmarks/bookmarks_connected_component.tsx
rasmus-storjohann-PG: You should make this kind of cleanup changes now. Don't put things off for Tidy Friday.

src/stores/services/index.ts
```diff
@@ -102,30 +102,34 @@ const createServiceMap = (services: ReadonlyArray<types.HumanServiceData>): type
...
+    const savedServiceId = savedService.id;
     return {
         ...store,
         services: {
             ...store.services,
-            [serviceId]: {
-                ...service,
-                bookmarked: true,
+            [savedServiceId]: {
+                ...savedService,
```
rasmus-storjohann-PG: Not sure that is cleaner...

src/sagas/__tests__/user_data.test.ts
```diff
@@ -26,4 +27,13 @@ describe('persisted user_data tests', () => {
         const recreatedData = deserializeUserData(serializeUserData(userData));
         expect(recreatedData.completedTopics[0]).toEqual(completedTopic);
     });
+
+    test('saved services should persist between being serialized and deserialized', () => {
+        const savedServiceId = aString();
+        const savedService = new ServiceBuilder().withId(savedServiceId).withBookmarked(true);
+        const savedServiceMap = buildServiceMap([savedService]);
```
rasmus-storjohann-PG: The more I think about this the less I like the current design. There is tight coupling between the builders and `buildServiceMap()`, since the latter make that `build()` call. We should change it to make the build call before calling `buildServiceMap()`, that way we have the option of calling `buildServiceMap()` with arguments that don't come from builders.

This is probably a big change, so not appropriate for this PR.

src/stores/services/actions.ts
```diff
@@ -34,7 +34,7 @@ export const buildServicesRequestAction = (topicId: TopicId, manualUserLocation?
 );
 
 // tslint:disable-next-line:typedef
-export const buildServicesSuccessAction = (topicId: TopicId, services: ReadonlyArray<HumanServiceData>) => (
+export const buildServicesSuccessAction = (topicId: TopicId, services: ServiceListItems) => (
```
rasmus-storjohann-PG: This change actually makes the code less readable, because `ReadonlyArray<HumanServiceData>` carries a lot of information about the type, which `ServiceListItems` doesn't. If it was called `HumanServiceList` or something, it might make the connection more clear, but even then I would have to look up the type to figure out what exactly it is.

I think you can remove this commit from the PR.

src/validation/services/persisted_services/__tests__/persisted_services_schema.test.ts
```diff
@@ -28,25 +28,11 @@ describe('schema for saved service map', () => {
         const anId = aString();
         const aLatitude = aNumber();
         const aLongitude = aNumber();
-        const aLatLong = {
-            lat: aLatitude,
-            lng: aLongitude,
-        };
+        const aLatLong = new LatLongBuilder().withLatitude(aLatitude).withLongitude(aLongitude).build();
```
rasmus-storjohann-PG: Compare what this line would look like if the builder api was 
```ts
const buildLatLong(lat: number, lng: number): LatLong;
```

src/validation/services/persisted_services/__tests__/persisted_services_schema.test.ts
```diff
@@ -28,25 +28,11 @@ describe('schema for saved service map', () => {
...
-            type: aString(),
-            id: aNumber(),
-            address: aString(),
-            city: aString(),
-            stateProvince: aString(),
-            postalCode: aString(),
-            country: aString(),
-        };
+        const aPhoneNumber = new PhoneNumberBuilder().build();
+        const anAddress = new AddressBuilder().build();
```
rasmus-storjohann-PG: nice

src/components/bookmark_button/bookmark_button_component.tsx
```diff
@@ -9,8 +9,8 @@ export interface BookmarkButtonProps {
     readonly textColor: string;
 }
 
-export type AddBookmarkAction = AddToSavedListAction | AddServiceToSavedListAction;
-export type RemoveBookmarkAction = RemoveFromSavedListAction | RemoveServiceFromSavedListAction;
+export type AddBookmarkAction = AddTopicToSavedListAction | AddServiceToSavedListAction;
+export type RemoveBookmarkAction = RemoveTopicFromSavedListAction | RemoveServiceFromSavedListAction;
```
rasmus-storjohann-PG: Use consistent terminology, "bookmark" and "saved list" really mean the same thing, "bookmark" is the clearer name.

src/stores/services/index.ts
```diff
@@ -89,3 +100,55 @@ const createServiceMap = (services: ReadonlyArray<types.HumanServiceData>): type
     };
     return services.reduce(theReducer, {});
 };
+
+const addToSavedServicesList = (store: types.ServiceStore, action: actions.AddServiceToSavedListAction): types.ServiceStore => {
```
bensinjin: So after my most recent PR, where we're saving ALL services, including the ones from Algolia into Redux, this will change, as services you want to bookmark should always already exist in Redux. This means this function could just be called `bookmarkService` and take an id to just index into the store and flip the flag.

src/stores/services/index.ts
```diff
@@ -89,3 +100,55 @@ const createServiceMap = (services: ReadonlyArray<types.HumanServiceData>): type
...
+                ...serviceToRemove,
+            },
+        },
+    };
+};
+
+const clearServicesData = (store: types.ServiceStore): types.ServiceStore => (
+    {
+        ...store,
+        services: {},
```
bensinjin: I'm confused, so when we fire the clear all user data action we're going to remove all the services from the store? I think we'd just want to flip all the bookmarked flags to false no? Also, remember we have `servicesByTopic` under this property too, we shouldn't be deleting that.

src/stores/services/index.ts
```diff
@@ -89,3 +100,55 @@ const createServiceMap = (services: ReadonlyArray<types.HumanServiceData>): type
...
+    }
+);
+
+const loadServicesFromUserData = (store: types.ServiceStore, action: UserDataPersistence.LoadSuccessAction): types. ServiceStore => {
+    const persistedServices: types.ServiceMap = action.payload.savedServices;
+    return {
+        ...store,
+        services: {
+            ...store.services,
+            ...persistedServices,
```
bensinjin: Again not sure this is right, you're not interested in `servicesByTopic` I don't think.

src/stores/services/index.ts
tomy-pg: We've discussed this and it seems from the designs we should be able to add services into memory from the bookmark action.

src/stores/services/index.ts
tomy-pg: From the discussion you, Rasmus and I had, we concluded that clearing when we clear all user data, we essentially want to restore all the states back to default.

src/components/services/service_list_item_component.tsx
```diff
@@ -16,19 +16,33 @@ import { AnalyticsLink, LinkTypes } from '../link/link';
...
+
+type Props = ServiceListItemProps & ServiceListItemActions;
+
 export const ServiceListItemComponent: React.StatelessComponent<Props> =
     (props: Props): JSX.Element => {
         const serviceName = buildServiceName(props.service.organizationName, props.service.name);
         const linkContext = buildAnalyticsLinkContext('Service', serviceName);
         return (
             <View style={{ backgroundColor: colors.white, padding: 10, marginTop: 10 }}>
+                <BookmarkButtonComponent isBookmarked={props.isBookmarked} textColor={colors.teal}
```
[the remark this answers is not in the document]
tomy-pg: As part of the Acceptance Criteria I have received for Issue #840 and #841, the Bookmark Button is expected to be found within the ServiceListItemComponent.  With this change, the bookmark button will be found in the service list. 

However, I see what you mean in that we do not see this in the mocks currently.

src/sagas/__tests__/user_data.test.ts
tomy-pg: In this PR I will assign all variables using ServiceBuilder without `.build()` the suffix _builder_ to make this a bit clearer.

src/components/search/infinite_hits_component.tsx
```diff
@@ -44,7 +55,13 @@ const keyExtractor = (item: SearchServiceData): string => (
...
 );
 
-const renderSearchHit = R.curry((currentPath: string, itemInfo: ListRenderItemInfo<SearchServiceData>): JSX.Element => {
+const renderSearchHit = R.curry((props: Partial<Props>, itemInfo: ListRenderItemInfo<SearchServiceData>): JSX.Element => {
     const item: SearchServiceData = itemInfo.item;
-    return <ServiceListItemComponent service={toHumanServiceData(item)} currentPath={currentPath} />;
+    return <ServiceListItemComponent
+        service={toHumanServiceData(item)}
+        currentPath={props.currentPath}
+        isBookmarked={R.contains(item.service_id, props.savedServicesIds)}
```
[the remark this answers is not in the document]
tomy-pg: toHumanServiceData transforms the DTO coming from Algolia to a HumanServiceData. The DTO coming from Algolia does not know about the property `bookmarked`

src/components/bookmarks/bookmarks_component.tsx
```diff
@@ -0,0 +1,24 @@
...
+    readonly bookmarkedServices: ServiceListItems;
+    readonly savedServicesIds: ReadonlyArray<Id>;
+    readonly bookmarkedTopics: ReadonlyArray<TopicListItem>;
+}
+
+type Props = BookmarksProps & ListActions & RouterProps ;
+
+export const BookmarksComponent: React.StatelessComponent<Props> = (props: Props): JSX.Element => {
+    return (
+        <React.Fragment>
```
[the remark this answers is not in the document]
tomy-pg: If I had used a View here, I would not have been able to see the footer which is why I opted for a fragment knowing that this would eventually change.

src/selectors/services/select_saved_services_ids.ts
```diff
@@ -0,0 +1,7 @@
+import { Store } from '../../stores';
+import { selectSavedServiceMap } from './select_saved_service_map';
+import { Id } from '../../stores/services';
+
+export const selectSavedServicesIds = (appStore: Store): ReadonlyArray<Id> => (
+    Object.keys(selectSavedServiceMap(appStore))
```
rasmus-storjohann-PG: Let's prefer R function over Object ones, since the Ramda library is so much more complete, that way we avoid ever mixing calls to the two libraries.

src/stores/__tests__/helpers/services_helpers.ts
```diff
@@ -78,6 +103,27 @@ export class AddressBuilder {
...
+        this.longitude = longitude;
+        return this;
+    }
+    build(): LatLong {
+        return {
+            lat: this.latitude,
+            lng: this.longitude,
+        };
+    }
+}
```
rasmus-storjohann-PG: Builders are handy when you have many properties, and many tests where you don't care about the values of most of those properties. Here you have 2 properties, which is a pretty small value for "many". Might be fine to just have a function `toLatLong(lat:number, lng: number)` and call it with aNumber() in tests where you don't care about the value.

src/selectors/services/select_saved_services_ids.ts
tomy-pg: The reason I do not use Ramda for this is because of a type error I receive:
```
Type '(string | number)[]' is not assignable to type 'ReadonlyArray<string>'
```
I specifically want this function to return a ReadonlyArray<string>.

src/stores/__tests__/helpers/services_helpers.ts
```diff
@@ -103,26 +103,9 @@ export class AddressBuilder {
...
-        return this;
-    }
-    build(): LatLong {
-        return {
-            lat: this.latitude,
-            lng: this.longitude,
-        };
-    }
-}
+export const toLatLong = (lat: number, lng: number): LatLong => (
```
rasmus-storjohann-PG: dead code

src/components/bookmarks/service_bookmarks_component.tsx
```diff
@@ -28,6 +28,7 @@ export const ServiceBookmarksComponent: React.StatelessComponent<Props> = (props
 export const renderServiceItems = (props: Props): ({ item }: ServiceItemInfo) => JSX.Element => {
     return ({ item }: ServiceItemInfo): JSX.Element => (
         <ServiceListItemComponent
+        history={props.history}
```
rasmus-storjohann-PG: why didn't the compiler catch this?

tomy-pg: the history property is optional. I will make this a required property so this is not missed again.

Issue #915, opened 2019-11-28
Add algolia service search results to Redux store 


rasmus-storjohann-PG:
Currently, search for services related to a given topic causes the returned services to be stored in the redux store. Services returned from Algolia is kept in the component and not saved to the redux store. This stops us from putting a service from Algolia into a service detail screen.

### Acceptance Criteria 
- [x]  Service search results from Algolia are added to the Redux store
- [x] Can open a service detail screen with a service from the algolia search result.
- [x] Add google analytics hooks on how many services are stored, with the view to understand whether services take up too much memory on users' phones, send an event for each time more services are added to the store.

### Testing steps
To get started testing you'll need to first add a test `GOOGLE_ANALYTICS_TRACKING_ID` key to your `.env`, this work sends events to GA so we don't want them going to the production GA account. Now, in GA, open the Reports > Realtime > Events tab so you can monitor events in realtime (see screenshot below). Finally you'll want to adjust the value of `MEMORY_REPORT_SEND_EVERY_SERVICES_COUNT` to be something smaller and easier to test with, say 1 or 2, you can refer to the PR for more implementation details. Confirm the following:

- [x] Search results now click through to their detail screen, services are added to Redux when you tap their title in the search results
- [x] `MemoryReport` Events are sent to GA every `MEMORY_REPORT_SEND_EVERY_SERVICES_COUNT` count, for example, if this was set to 2 we'd see a report every time the number of new services in Redux is divisible (no remainder) by 2, so, 2, 4, 6 etc...
- [x] `MemoryReport` Events are only sent when NEW services are added to Redux, we shouldn't see a report just because the current Redux count is divisible
- [x] Topic service results also trigger `MemoryReport` events in GA

Here's an example of what you should start to see in GA when you start clicking on services:

<img width="1894" alt="Screen Shot 2019-12-04 at 9 44 55 AM" src="https://user-images.githubusercontent.com/3228323/70167108-e1e62900-167a-11ea-9171-e8fcd45793ed.png">

tomy-pg:
Passes QA in Staging Version 1.12

PR #924, opened 2019-12-04
Issue 915 Add search results to Redux and send reports to Google Analytics

bensinjin:
This PR allows us to tap through to service detail screens from the Algolia search results, it also sends a report to GA outlining how many services are in memory under certain conditions. See https://github.com/pg-irc/pathways-frontend/issues/915 for testing instructions. Here are the high level points:

- Services are added to Redux `onPress`, not when they come back from Algolia, this means we're not storing any unnecessary services in Redux
- Reports are sent to GA at intervals of every N new services created in Redux , this is so we're not sending thousands of events to GA but still understand in general what the popular thresholds are


src/application/store.ts
```diff
@@ -45,7 +46,7 @@ const buildStoreWithLocaleData = (): Store => {
 type CreatedStore = Readonly<ReturnType<typeof createStore>>;
 
 export const buildStore = (saga: ApplicationSaga): CreatedStore => {
-    const middleware = applyMiddleware(saga.middleware);
+    const middleware = applyMiddleware(saga.middleware, GoogleAnalytics.MemoryReportMiddleware);
```
rasmus-storjohann-PG: This naming is very confusing, I was thinking this was a google library. By our convention, `MemoryReportMiddleware` is capitalized so it should be a type, but it's actually a function, it's inconsistent with `saga.middleware`. How about `analytics.middleware`?

Clean code has some good advice on using 3rd party names like "google" in the code base, it's a bad idea because if we one day switch to some other analytics solution, we have to make naming changes in our code base for it to not be misleading. Generic terms like "analytics" do not have this problem.

src/middleware/google_analytics.ts
```diff
@@ -0,0 +1,58 @@
...
+        const newServicesCount = countNewServicesToBeCreatedByAction(action, state);
+        const totalServicesCount = countTotalServicesAfterActionFinishes(newServicesCount, state);
+
+        if (shouldSendServicesCountReport(newServicesCount, totalServicesCount, MEMORY_REPORT_SEND_EVERY_SERVICES_COUNT)) {
+            sendServicesCountEvent(totalServicesCount);
+        }
+
+        return next(action);
+    }),
+};
```
rasmus-storjohann-PG: The code in this commit is really quite complicated for what it's doing. Note that the requirement to keep the number of spurious GA events low is not very firm, I don't actually care about getting a few extra events, I just want to know roughly how large the # of services is getting. So it can be as simple as `if number of saved services is a multiple of 100, send an event`. If you use the # of currently saved services + the services to add as an estimate of the total # of services saved, that's fine too.

src/stores/__tests__/services.test.ts
```diff
@@ -159,4 +160,32 @@ describe('services reducer', () => {
...
+            const aNewName = 'ANewName';
+            const aSearchService = new ServiceBuilder().withId(aService.id).withName(aNewName).build();
+            const action: BuildSaveServiceFromSearchAction = {
+                type: constants.SAVE_SERVICE_FROM_SEARCH,
+                payload: {
+                    service: aSearchService,
+                },
+            };
+            const store = reducer(theStore, action);
+            expect(store.services[aSearchService.id].name).toEqual(aNewName);
```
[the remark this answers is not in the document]
bensinjin: Fixed, I'll keep in mind your point about keeping the UX and store changes separate.

src/stores/services/actions.ts
```diff
@@ -29,4 +32,9 @@ export const buildServicesSuccessAction = (topicId: TopicId, services: ReadonlyA
 // tslint:disable-next-line:typedef
 export const buildServicesErrorAction = (topicId: TopicId, errorMessageType: Errors) => (
     helpers.makeAction(constants.LOAD_SERVICES_FAILURE, { topicId, errorMessageType })
+);
+
+// tslint:disable-next-line:typedef
+export const buildSaveServiceFromSearchAction = (service: HumanServiceData) => (
```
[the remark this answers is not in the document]
bensinjin: Fixed and renamed to just `saveService` and `SaveServiceAction` as I think "store" should be the assumption here. At this point I think it makes sense that the Saga that listens for Redux actions be more explicit about which mediums it's saving data to. I agree the two sources for services is not ideal and in this instance we shouldn't add confusion by specifying the source (search) as really this can be used to save a service from anywhere into the store.

src/components/addresses/addresses_component.tsx
```diff
@@ -4,7 +4,7 @@ import { Trans } from '@lingui/react';
 import { Icon } from 'native-base';
```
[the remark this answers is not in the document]
bensinjin: Fixed

src/components/addresses/addresses_component.tsx
```diff
@@ -45,3 +35,18 @@ export const SingleAddressComponent = (props: {readonly address: Address}): JSX.
...
+    const leftContent = <SingleAddressComponent address={address} />;
+    const rightContent = (
+        <Icon
+            name={'location-arrow'}
+            type={'FontAwesome'}
+            style={{ color: colors.teal, fontSize: values.smallIconSize, paddingRight: 10 }}
+        />
+    );
+    const onPress = props.onPressForAddress ?
+        props.onPressForAddress(address) : (): undefined => undefined;
```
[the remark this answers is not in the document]
bensinjin: Ah of course, I've fixed this up.

src/components/addresses/addresses_component.tsx
```diff
@@ -45,3 +35,18 @@ export const SingleAddressComponent = (props: {readonly address: Address}): JSX.
         </Text>
     </View>
 );
+
+const buildCardButtonProps = (address: Address, props: Props): CardButtonProps => {
```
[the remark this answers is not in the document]
bensinjin: Fixed

src/middleware/google_analytics.ts
bensinjin: To be honest I saw this as an opportunity to lay the groundwork for sending Redux analytics to GA. With this in mind I wanted this to be somewhat "high quality" and accurate. 

Here's my thoughts on your points: 

- Without the requirement for the action payload to contain at least one new service, when for example loading 100 services from a topic, you would send the event to GA once, then, you'd send it every time you pressed into one of those topics as the number of services in Redux is still a multiple of 100. I wanted the events to at least accurately reflect hitting that threshold once and not send unnecessary duplicate events.
-  We're not always adding services one at a time (this works for topic services too), also, we're not always guaranteed the service(s) we're "adding" aren't already in Redux, we could do this check before we write to Redux but I'm not convinced that's really preferable

src/components/search/infinite_hits_component.tsx
```diff
@@ -44,7 +51,22 @@ const keyExtractor = (item: SearchServiceData): string => (
...
 );
 
-const renderSearchHit = R.curry((currentPath: string, itemInfo: ListRenderItemInfo<SearchServiceData>): JSX.Element => {
-    const item: SearchServiceData = itemInfo.item;
-    return <ServiceListItemComponent service={toHumanServiceData(item)} currentPath={currentPath} />;
+const renderSearchHit = R.curry((
+        currentPath: string,
+        history: History,
+        saveServiceFromSearch: (service: HumanServiceData) => BuildSaveServiceFromSearchAction,
+        itemInfo: ListRenderItemInfo<SearchServiceData>,
```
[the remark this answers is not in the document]
bensinjin: Fixed

src/middleware/__test__/google_analytics.test.ts
```diff
@@ -0,0 +1,95 @@
...
+import { ServiceBuilder, buildNormalizedServices } from '../../stores/__tests__/helpers/services_helpers';
+import { buildDefaultStore } from '../../stores';
+import { aNumber } from '../../helpers/random_test_values';
+
+describe('Google Analytics memory report middleware helpers', () => {
+    const aServiceBuilder = new ServiceBuilder();
+    const store = {
+        ...buildDefaultStore(),
+        services: buildNormalizedServices([aServiceBuilder], []),
+    };
```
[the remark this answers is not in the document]
bensinjin: Fixed

src/application/store.ts
bensinjin: Fixed

src/middleware/google_analytics.ts
bensinjin: I've rewritten this to directly use the service reducer which should make the code easier to follow.

bensinjin:
Everything addressed @rasmus-storjohann-PG

src/application/store.ts
```diff
@@ -1,7 +1,7 @@
 import { createStore, applyMiddleware, compose } from 'redux';
 
 import { reducer, Store, buildDefaultStore } from '../stores';
-import { GoogleAnalytics } from '../middleware/google_analytics';
+import { analytics } from '../middleware/analytics';
```
rasmus-storjohann-PG:
```
import * as analytics from '../middleware/analytics';
```

src/middleware/analytics.ts
```diff
@@ -0,0 +1,62 @@
...
+import { sendServicesCountEvent } from '../sagas/analytics/events';
+import { SAVE_SERVICE, LOAD_SERVICES_SUCCESS, MEMORY_REPORT_SEND_EVERY_SERVICES_COUNT } from '../application/constants';
+import { reducer as servicesReducer, ServiceStore } from '../stores/services';
+import { ServicesAction } from '../stores/services/actions';
+
+export const analytics = {
+    middleware: R.curry((
+        store: MiddlewareAPI<Dispatch, Store>,
+        next: Dispatch,
+        action: AnyAction) => {
```
rasmus-storjohann-PG:
```ts
export middleware: R.curry((store: MiddlewareAPI<Dispatch, Store>, next: Dispatch, action: AnyAction) => {
```

src/middleware/analytics.ts
```diff
@@ -0,0 +1,62 @@
...
+        const state = store.getState();
+
+        if (actionUpdatesServices(action)) {
```
rasmus-storjohann-PG: this becomes a case statement

src/middleware/analytics.ts
```diff
@@ -0,0 +1,62 @@
...
+            sendServicesCountReport(state, <ServicesAction> action);
```
rasmus-storjohann-PG: Lose the cast here

src/middleware/analytics.ts
```diff
@@ -0,0 +1,62 @@
...
+        }
+
+        return next(action);
+    }),
+};
+
+export const serviceWasAdded = (currentServiceState: ServiceStore, nextServiceState: ServiceStore):
```
rasmus-storjohann-PG: oldstore, newstore

src/middleware/__tests__/analytics.tests.ts
```diff
@@ -0,0 +1,72 @@
...
+import { buildDefaultStore } from '../../stores/services';
+import { aNumber } from '../../helpers/random_test_values';
+
+describe('Analytics memory report middleware helpers', () => {
+
+    describe('serviceWasAdded()', () => {
+
+        it('returns true when next service store state has a new service', () => {
+            const aServiceBuilder = new ServiceBuilder();
+            const currentState = buildDefaultStore();
```
rasmus-storjohann-PG: `buildNormalizedServices([], []);`

src/application/store.ts
bensinjin: Fixed

src/middleware/analytics.ts
bensinjin: Fixed

src/middleware/analytics.ts
bensinjin: Landed on using and if

src/middleware/analytics.ts
bensinjin: Fixed

src/middleware/analytics.ts
bensinjin: Fixed

bensinjin:
@rasmus-storjohann-PG this is ready

src/stores/__tests__/services.test.ts
```diff
@@ -176,16 +176,20 @@ describe('services reducer', () => {
...
                 payload: {
-                    service: aSearchService,
+                    service: newService,
                 },
             };
-            const store = reducer(theStore, action);
-            expect(store.services[aSearchService.id].name).toEqual(aNewName);
+
+            const newStore = reducer(oldStore, action);
+            expect(newStore.services[serviceId].name).toEqual(newService.name);
```
rasmus-storjohann-PG: `sendServicesCountReport` refactor: Using R functions consistently is cleaner than mixing Object and .length; Change helper function `computeNumberOfServicesAdded()` to return the # of services so the logic can be put inline. Removed tests for isDivisable, which is really just testing how `%` works. I also fixed up the test "overwrites an existing service with the same id" because it wasn't testing what it said it was testing.

rasmus-storjohann-PG: If we get services from the topic search, more than one service may be added to the store, so we could jump from e.g. 98 services to 102, which should trigger an GA event but currently wont. Need to fix this in a later PR, I want to get this one in. I still have to do QA on it, so don't merge it yet.

bensinjin: Good point. Could we just say: "If new services are added and the total count / send every count > 1"?

bensinjin: Good point ... do we need to keep track of the last count we sent the event at? So the algorithm to send again would be something like: send if: (count after action - last sent at count) / send per count >= 1 and if that passes update last sent at count to count after action?

some examples:
```
// count after action: 102, last sent at count: 0, send per count: 100
(102 - 0) / 100 >= 1 ? // 1.02 passes, update last sent at count to 102

// count after action: 202, last sent at count: 102, send per count: 100
(202 - 102) / 100 >= 1 ? // 1 passes, update last set at count to 202

// count after action: 203, last sent at count: 202, send per count: 100
(203 - 202) / 100 >= 1 ? // .01 fails, do nothing
```

rasmus-storjohann-PG: So in `sendServicesCountReport()` or nearby, you will have the old and new services count C1 and C2, and you have the constant M = 100. if (Math.floor(C1/M) < Math.floor(C2/M)) then send the report.

bensinjin: Yes, that's much nicer. Should I go ahead and implement that?

bensinjin: @rasmus-storjohann-PG Fixed!

PR #934, opened 2019-12-07
QA-Tool Issue include services_at_location_id to human service data

tomy-pg:
`services_at_location_id` has been added as an optional property. The reason it has been made option and not required is because the current Algolia data does not provide the `service_at_location_id`.


src/validation/services/__tests__/helpers/services_schema_helpers.ts
```diff
@@ -400,9 +401,15 @@ export class LocationJSONBuilder {
 }
 
 export class ServiceAtLocationJSONBuilder {
+    id: number | null = aNumber();
```
rasmus-storjohann-PG: Simplify: This should just be a number, since you have the different build* functions below to give you the flexibility you need there.

Issue #937, opened 2019-12-12
Client: Upgrade all libraries to the latest version


rasmus-storjohann-PG:
There's a problem with Typescript with errors "this value may be undefined".

tomy-pg:
Closing this as I do not see the error.

PR #941, opened 2019-12-13
#937 Upgrade to Expo SDK 35 (and fix TS problems)

bensinjin:
This PR doesn't upgrade all our packages (it was a long day) but does get Expo & its dependencies up to date as well as fixing the TS problems mentioned in the story. It also outlines some new update instructions that I'm sure will evolve over time. Next time we do updates (I can do another round next week if we want), we should look at running the `yarn outdated` command and tackling a couple of the packages that are looking older according to that report.

Note:

One of the commits here sets the Saga generator type to "any" in tests if anyone cares to take a crack at properly typing this feel free as I gave up, although I'm not sure that's a hill worth dying on ;)


src/sagas/post_push_notification_token.ts
```diff
@@ -32,10 +32,10 @@ export function* watchRequestPostPushNotificationToken(): IterableIterator<ForkE
 }
 
 function* requestPostPushNotificationToken(_: PushNotificationPostRequestAction): Result {
-    const existingStatus = yield call(Permissions.getAsync, Permissions.NOTIFICATIONS);
+    const existingStatus: Permissions.PermissionResponse = yield call(Permissions.getAsync, Permissions.NOTIFICATIONS);
     // tslint:disable-next-line: no-let
-    let finalStatus = existingStatus;
-    if (existingStatus !== 'granted') {
+    let finalStatus: Permissions.PermissionResponse = existingStatus;
+    if (existingStatus.status !== 'granted') {
```
bensinjin: This actually also fixes a small bug that caused us to always ask for permission on iOS even when permission was already granted.

PR #942, opened 2019-12-13
Feature/issue 932 notification open on page

rasmus-storjohann-PG:


src/components/main/main_component.tsx
```diff
@@ -55,6 +59,7 @@ export class MainComponent extends React.Component<Props, State> {
         this.props.history.listen((location: Location, _: Action) =>
             this.props.sendAnalyticsData(location, this.props.locale),
         );
+        Notifications.addListener(notificationListener(this.props.history));
```
bensinjin: So does listener navigate to a screen when a notification is pressed or when it's received?

rasmus-storjohann-PG: Yes

PR #945, opened 2019-12-17
Feature/issue 903 create analytic to track expandable component

ian-lin-PG:


src/components/expandable_content/expandable_content_component.tsx
```diff
@@ -10,10 +10,13 @@ import {
     defaultExpandableContentState, isDefaultState,
 } from './expandable_content_states';
 import { values } from '../../application/styles';
+import { ExpandDetailAction, ReduceDetailAction } from '../../stores/topics';
 
 export interface ExpandableContentProps {
     readonly content: JSX.Element;
     readonly contentId?: string;
+    readonly expandDetail?: () => ExpandDetailAction;
```
bensinjin: I think this should probably be a connected component where you always fire the associated Redux actions when we expand or collapse however, if there's a case for not firing this action, or if your desire is not couple this to the store, your function should more closely resemble an even listener ie "onExpandContent" (with type: `() => void`) and not a Redux action creator.

src/components/expandable_content/expandable_content_component.tsx
```diff
@@ -112,6 +115,12 @@ export class ExpandableContentComponent extends React.Component<ExpandableConten
             ...this.state,
             expandableState: toggleExpandedState(this.state.expandableState),
         });
+        if (this.state.expandableState === 1 && this.props.expandDetail) {
```
bensinjin: You can use the enum for example: `ExpandableContentStates.isCollapsed`

src/application/constants.ts
```diff
@@ -63,4 +63,6 @@ export const POST_PUSH_NOTIFICATION_TOKEN_REQUEST = 'POST_PUSH_NOTIFICATION_TOKE
 export const POST_PUSH_NOTIFICATION_TOKEN_SUCCESS = 'POST_PUSH_NOTIFICATION_TOKEN_SUCCESS';
 export const POST_PUSH_NOTIFICATION_TOKEN_FAILURE = 'POST_PUSH_NOTIFICATION_TOKEN_FAILURE';
 
-export const MEMORY_REPORT_SEND_EVERY_SERVICES_COUNT = 100;
\ No newline at end of file
+export const MEMORY_REPORT_SEND_EVERY_SERVICES_COUNT = 100;
+export const EXPAND_DETAIL = 'EXPAND_DETAIL';
+export const REDUCE_DETAIL = 'REDUCE_DETAIL';
```
tomy-pg: Changing this to `COLLAPSE_DETAIL` would make this more consistent with the code base.

src/stores/topics/actions.ts
```diff
@@ -27,3 +33,13 @@ export const addToSavedList = (topicId: Id) => {
...
+
+// tslint:disable-next-line:typedef
+export const expandTopicDetail = (topicId: Id) => (
+    helpers.makeAction(constants.EXPAND_DETAIL, { topicId })
+);
+
+// tslint:disable-next-line:typedef
+export const reduceTopicDetail = (topicId: Id) => (
+    helpers.makeAction(constants.REDUCE_DETAIL, { topicId })
+);
```
tomy-pg: Changing this to `collapseTopicDetail` would make things more consistent.

src/components/expandable_content/expandable_content_component.tsx
ian-lin-PG: Currently this action only fires for the `Read more` and `Read less` on the topic page. More work will be required so that the organization detail, service detail and service listing detail all to be logged. 
So far all of google analytics events are watched by saga, so I decided to follow that pattern. So the suggestion is to have an event listener that trigger the GA events directly instead of going through saga?

ian-lin-PG: After deliberating with rasmus, changed the function to be `onExpand` with void return.

PR #948, opened 2019-12-19
Feature/issue 742 exclude our devices from analytics

rasmus-storjohann-PG:


rasmus-storjohann-PG:
@ian-lin-PG @tomy-pg @bensinjin PR is up for you to look at. Quite a bit of renames, and some Redux changes.

src/components/search/api/use_fetch_lat_long_from_location.ts
```diff
@@ -63,3 +68,16 @@ const getTextIfValidOrThrow = (response: Response): Promise<string> => {
 const handleError = R.curry((setLatLong: (latLong: LatLong) => void, _: string): void => {
     setLatLong(undefined);
 });
+
+export const useDisableAnalyticsOnEasterEgg = (location: string, disableAnalytics: (disable: boolean) => DisableAnalyticsAction): void => {
```
bensinjin: This probably isn't a hook you'll use more than once so probably can just put it in the file it's used in.

bensinjin: Dunno if this is really reusable so maybe should just live in the file it's used in.

rasmus-storjohann-PG: Fixed

Issue #861, opened 2019-10-08
Services load slowly on iOS


cherrielam:
The loading time for services listing is currently quite slow on iOS, and this is a performance issue. It needs to be faster on iOS

### Difficulty, impact:
|--(technical difficulty = ??)--| -- high user importance (SHOULD have) -- |

### Acceptance Criteria:

- [ ]  Subsequent loads of services are quicker

ArtiBPG:
Does it seem slow on Android?

bensinjin:
To improve performance we suggest the following:

ACs:
- Add `expiresAt` property to `servicesForTopics` object in the store
- When topic services are fetched from the server we update `expiresAt` to the current time plus one day
- The topic services component does not hit the server if there are unexpired topic services available

tomy-pg:
@bensinjin can we get steps for QA on this issue?

jparlina:
Tested on Staging V1.16.0 using iOS 7s at the HiVE: 
It takes about 9-12 seconds to load services. Discussed with @cherrielam, we both think it's still slow. Is it technically possible to cut the load time in half for better user experience? 

@bensinjin Is there a benchmark for acceptable loading time?

bensinjin:
So the original work actually is in place to speed up subsequent request, not the initial request, I did some digging however and think we can use another API to improve the initial fetch which is actually slow due to the time it takes to find your location. See: https://github.com/pg-irc/pathways-frontend/pull/1003

Testing instructions:

- [ ] Initial services fetch is faster
- [ ] An appropriate level of "distance" accuracy is achieved

jparlina:
Tested on Staging V1.22.0, server V1.3.24 on iPhone 7s at the HiVE - PASSED

Testing instructions:
- [x]   Initial services fetch is faster
- [x]   An appropriate level of "distance" accuracy is achieved

PR #950, opened 2019-12-23
Issue #861 Fix slow services

bensinjin:
This PR adds an expiry to topic services and ensures we only fetch services from the server once every 24 hours for a given topic.


src/components/services/service_list_component.tsx
```diff
@@ -42,83 +39,77 @@ export interface ServicesUpdater {
...
+
+const refreshServices = (props: Props): void => {
+    if (servicesFetchRequired(props.topicServicesOrError)) {
+        props.dispatchServicesRequest();
+    }
+};
+
+const servicesFetchRequired = (topicServicesOrError: SelectorTopicServices): boolean => (
+    !(topicServicesOrError.type === constants.TOPIC_SERVICES_VALID && topicServicesOrError.isStaleData === false)
+);
```
rasmus-storjohann-PG: Nice short functions. Can be made more readable by the rule `!(A && B) === !A || !B`. Generally avoid `=== false` and `=== true`

src/components/services/service_list_component.tsx
```diff
@@ -42,83 +39,77 @@ export interface ServicesUpdater {
...
         return topicServicesOrError.errorMessageType;
     }
     return Errors.Exception;
 };
 
-const isLoadingServices = (topicServicesOrError: SelectorTopicServices): boolean => (
-    topicServicesOrError.type === constants.TOPIC_SERVICES_LOADING
+const getServices = (topicServicesOrError: SelectorTopicServices): ReadonlyArray<HumanServiceData> => (
+    topicServicesOrError.type === constants.TOPIC_SERVICES_VALID && topicServicesOrError.services
 );
```
rasmus-storjohann-PG: I think renaming this function to `getServicesIfValid()` will make us think of the invalid case at call sites and squash bugs.

src/selectors/services/is_error_selector_topic_services.ts
```diff
@@ -0,0 +1,7 @@
+import { SelectorTopicServices, ErrorSelectorTopicServices } from './types';
+import * as constants from '../../application/constants';
+
+export const isErrorSelectorTopicServices = (topicServicesOrError: SelectorTopicServices):
+    topicServicesOrError is ErrorSelectorTopicServices => (
+        (<ErrorSelectorTopicServices>topicServicesOrError).type === constants.TOPIC_SERVICES_ERROR
+    );
```
rasmus-storjohann-PG: Have a look at the code with this function just called `isError()`. Does it cause ambiguity at the call sites, or would you say that the context makes it clear what the function is doing? If making this rename, do the same to the isValid case.

src/selectors/services/to_valid_selector_topic_services.ts
```diff
@@ -6,4 +6,5 @@ export const toValidSelectorTopicServices = (topicServices: ValidServicesForTopi
     ValidSelectorTopicServices => ({
         services: topicServices.serviceIds.map((serviceId: ServiceId) => services[serviceId]),
         type: constants.TOPIC_SERVICES_VALID,
+        isStaleData: topicServices.expiresAt < Date.now(),
```
rasmus-storjohann-PG: Avoid introducing multiple terms for the same thing, so `isExpired`.

src/validation/services/__tests__/to_human_service_data.test.ts
```diff
@@ -65,64 +65,36 @@ describe('Adapting server service object to client service object', () => {
...
-                    country: serverServiceObject.location.addresses[0].address.country,
-                }],
-                website: serverServiceObject.service.organization_url,
-                email: serverServiceObject.service.organization_email,
-                organizationName: serverServiceObject.service.organization_name,
-                bookmarked: true,
-            };
-            expect(clientServiceObject[0]).toEqual(expectedClientServiceObject);
-        });
+    it('builds expected structure with "bookmarked" set to true when bookmark exists', () => {
```
rasmus-storjohann-PG: There are two tests here, and they're testing for three things: build the expected structure, sets bookmarked to false, and sets bookmarked to true. As they stand, both test checks for the expected structure, which is why there is duplication. Clearly there should be three tests, only one of them should contain a bunch of JSON.

src/validation/services/index.ts
```diff
@@ -53,47 +53,42 @@ export interface ValidatedServiceAtLocationJSON {
...
+    ReadonlyArray<types.HumanServiceData> => (
+        R.map(serviceFromData(bookmarkedServiceIds), data)
+    );
 
-export const serviceFromValidatedJSON = (data: ValidatedServiceAtLocationJSON): types.HumanServiceData => {
-    const phoneNumbers = R.map((phoneNumber: ValidatedPhoneNumberJSON): types.PhoneNumber => ({
-        type: phoneNumber.phone_number_type,
-        phone_number: phoneNumber.phone_number,
-    }), data.location.phone_numbers);
+const serviceFromData = R.curry((bookmarkedServiceIds: ReadonlyArray<Id>, data: ValidatedServiceAtLocationJSON):
```
rasmus-storjohann-PG: This function would benefit from being broken up into several helper functions.

src/stores/__tests__/services.test.ts
```diff
@@ -139,6 +139,38 @@ describe('services reducer', () => {
...
+                    topicId,
+                    services: [serviceBuilder.build()],
+                },
+            };
+            const storeAfterAction = reducer(storeBeforeAction, theAction);
+            const servicesByTopic = storeAfterAction.servicesByTopic[topicId];
+
+            if (isValidServicesForTopic(servicesByTopic)) {
+                const twentyFourHoursAfterExpiry = expiresAt + (24 * 60 * 60 * 1000);
+                expect(servicesByTopic.expiresAt).toBeGreaterThanOrEqual(twentyFourHoursAfterExpiry);
```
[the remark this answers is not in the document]
bensinjin: Turns out we can actually test for the exact value here which I've updated this test to do.

src/components/services/service_list_component.tsx
bensinjin: Sounds good - I've cleaned this up to add clarity and removed the `=== false`.

src/components/services/service_list_component.tsx
bensinjin: Fixed

src/stores/__tests__/services.test.ts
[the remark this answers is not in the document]
bensinjin: I stand corrected, this works locally but we get some latency when CI runs it - I've added an upper limit.

src/selectors/services/is_error_selector_topic_services.ts
bensinjin: This was renamed to better reflect that we're referring to the selector type as is `isValidSelectorTopicServices`. There exists another version of these that accepts the store's  types so I felt it was preferable to be specific.

src/selectors/services/to_valid_selector_topic_services.ts
bensinjin: fixed

src/validation/services/__tests__/to_human_service_data.test.ts
bensinjin: fixed

src/validation/services/index.ts
bensinjin: fixed

bensinjin:
@rasmus-storjohann-PG I think this is ready

rasmus-storjohann-PG:
Looking at it now, it dropped off my radar a bit

PR #951, opened 2019-12-27
Feature/issue 841 services on bookmarks screen

tomy-pg:
Notes: 

Please see comments on `src/components/organization_detail_component.tsx` left by @bensinjin in regards to the usage of Native-Base's Tabs. Native-base Tabs are buggy and I was unable work around existing RTL issues. I have instead opted to use `react-native-tab-view`. 

Although I have prepared the service list item component to click through to the organization detail component, this part has not yet been completed. 

I extracted the new and edited strings from the bookmark component and removed no longer used jsx strings. Here is the PR on the ui-strings repo: https://github.com/pg-irc/ui-strings/pull/5.


src/components/services/service_list_item_component.tsx
```diff
@@ -32,15 +31,20 @@ export const ServiceListItemComponent: React.StatelessComponent<Props> =
...
-                <BookmarkButtonComponent isBookmarked={props.isBookmarked}
-                textColor={colors.teal}
-                bookmark={addBookmark}
-                unbookmark={removeBookmark}
-                />
-                {renderName(serviceName, props.onPress)}
-                {renderDescription(props.service.description)}
-                {renderAddresses(filterPhysicalAddresses(props.service.addresses))}
+            <View style={{ backgroundColor: colors.white, marginTop: 10, paddingVertical: 15, flex: 1, flexDirection: 'row'}}>
+                {/* native-base Button/Icon component has default padding */}
```
rasmus-storjohann-PG: Not sure what this comment means

src/components/bookmarks/service_bookmarks_component.tsx
```diff
@@ -2,11 +2,12 @@ import React from 'react';
 import { FlatList } from 'react-native';
 import { HumanServiceData } from '../../validation/services/types';
 import { EmptyListComponent } from '../empty_component/empty_list_component';
-import { ServiceListHeaderComponent, ServiceItemInfo } from '../services/service_list_component';
```
rasmus-storjohann-PG: So the type itself should also be removed

src/components/bookmarks/bookmarks_component.tsx
```diff
@@ -46,40 +47,64 @@ export interface TabSwitcherProps {
...
-    const serviceTabHeading = _('Services');
-    return (
+    const topicTabTitle = _('Topics');
+    const serviceTabTitle = _('Services');
+    const initialRoutes: TabRoutes = [
+        // key must be 'first', 'second', 'third' etc
+        { key: 'first', title: topicTabTitle },
+        { key: 'second', title: serviceTabTitle },
+    ];
+    const [index, setIndex]: [number, (n: number) => void] = React.useState(0);
```
rasmus-storjohann-PG: would it make sense to use 'first' and 'second' for index too?

src/components/bookmarks/bookmarks_component.tsx
```diff
@@ -47,65 +47,60 @@ export interface TabSwitcherProps {
...
-        );
+        </View>
+    );
 
     const renderScene = SceneMap({
-        first: FirstRoute,
-        second: SecondRoute,
+        topics: TopicsRoute,
+        services: ServicesRoute,
     });
```
rasmus-storjohann-PG: I'm struggling a bit with this terminology. I know that this library uses these terms, and they will appear in the API we call, but we are not obligated to carry it trough to our types. So I think `Component` is better than `Route`, and `renderScene()` could be simplified to just `render()`

src/components/bookmarks/bookmarks_tab_switcher.tsx
```diff
@@ -0,0 +1,90 @@
+import React, { Dispatch, SetStateAction } from 'react';
```
rasmus-storjohann-PG: The filename is redundant, `src/components/bookmarks/bookmarks_tab_switcher.tsx` can be simplified to `src/components/bookmarks/tab_switcher.tsx` without creating ambiguity.

src/components/bookmarks/bookmarks_tab_switcher.tsx
```diff
@@ -0,0 +1,90 @@
...
+import { Trans } from '@lingui/react';
+import { colors, textStyles } from '../../application/styles';
+import { BookmarksProps } from './bookmarks_component';
+import { View, Text } from 'native-base';
+import { ListActions } from './bookmarks_connected_component';
+import { RouterProps } from '../../application/routing';
+
+export interface BookmarksTabSwitcherProps {
+    readonly i18n: ReactI18n;
+}
```
rasmus-storjohann-PG: Again see if we already have this type somewhere.

src/components/bookmarks/bookmarks_tab_switcher.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+export const BookmarksTabSwitcher = (props: Props): JSX.Element => {
+    const _ = props.i18n._.bind(props.i18n);
+    const tabRoutes: TabRoutes = [
+        { key: 'topics', title: _('Topics') },
+        { key: 'services', title: _('Services') },
+    ];
+
+    const [index, setIndex]: [number, (n: number) => void] = React.useState(0);
+    const [routes]: [TabRoutes, Dispatch<SetStateAction<TabRoutes>>] = React.useState(tabRoutes);
```
rasmus-storjohann-PG: Surprised to see the routes hook, what's the reasoning? I would have thought we could just use `tabRoutes`. Also, we usually import from React so that we can just call `useState()`, not `React.useState()`

src/components/bookmarks/bookmarks_tab_switcher.tsx
```diff
@@ -0,0 +1,90 @@
...
+        style={{backgroundColor: colors.white}}
+        initialLayout={{ width: Dimensions.get('window').width }}
+        sceneContainerStyle={{backgroundColor: colors.lightGrey}}
+        />
+    );
+};
+
+export interface NavigationStateRoute {
+    readonly navigationState: NavigationState<Route>;
+}
```
rasmus-storjohann-PG: You could also perhaps just add `readonly navigationState: NavigationState<Route>;` as a property to the `SceneRendererProps` type?

src/components/bookmarks/bookmarks_tab_switcher.tsx
tomy-pg: Removed the routes hook to just use the tabRoutes. This was not needed.

src/components/services/service_list_item_component.tsx
tomy-pg: This was just a reminder for myself. Removed in a later PR.

src/components/bookmarks/bookmarks_component.tsx
tomy-pg: `navigationState` expects `index` to be of type number.

src/components/bookmarks/bookmarks_tab_switcher.tsx
tomy-pg: `renderTabBar` expects to receive a prop of type 
`SceneRendererProps & {
        navigationState: NavigationState<T>;
    }`. 

I was just trying to help myself understand what this was by separating NavigationState into a separate interface and then using a Union type to combine them.

PR #953, opened 2019-12-27
Feature issue 938 search input fields

ian-lin-PG:


src/components/search/search_input_component.tsx
```diff
@@ -36,36 +35,38 @@ export const SearchInputComponent = (props: Props & Actions): JSX.Element => {
...
-                            debug(`SearchInputComponent location text changed to '${d}'`);
-                            setLocation(d);
-                        }}
-                        value={location}
-                        onEndEditing={(): void => props.setLocation(location)}
-                        placeholder={buildTranslatedPlaceholder(i18nRenderProp.i18n, 'Near My location')} // TODO translate
-                        placeholderTextColor={colors.white}
-                    />
-                </View>
-                <InputFormSeparator />
```
ian-lin-PG: Delete this component if no where else uses it.

src/components/search/search_input_component.tsx
```diff
@@ -48,7 +56,7 @@ export const SearchInputComponent = (props: Props & Actions): JSX.Element => {
         debug(`SearchInput Component useEffect with '${props.currentRefinement}'`);
         props.refine(props.currentRefinement);
     }, [props.latLong]);
-
+    const [locationInput, setLocationInput]: [boolean, (b: boolean) => void] = useState();
```
ian-lin-PG: Make sure states are always initialized.

src/components/search/search_input_component.tsx
ian-lin-PG: Component deleted as no other part of the code base uses it.

src/components/search/search_input_component.tsx
ian-lin-PG: Renamed the state to be `hideMyLocationButton` and initialized to be `true`.

rasmus-storjohann-PG:
One issue found:
* Open search page
* Enter some search terms and get the results
* Click on a service to open the service detail
* Click the back button
* Expected: The search terms are still there
* Observed: The search screen input fields have been reset to empty

PR #955, opened 2020-01-02
Issue #841 add missing bookmark actions to search components

tomy-pg:
On iOS devices, I was receiving errors when reloading the app from LTR to RTL and vice versa. 

As part of the newest SDKs Expo has released, they require modular imports when using their packages.


.env.example
```diff
@@ -8,3 +8,5 @@ GOOGLE_ANALYTICS_TRACKING_ID='UA-00000000-0'
 DEBUG_GOOGLE_ANALYTICS=false
 ALGOLIA_SEARCH_API_KEY='api key'
 ALGOLIA_SERVICES_INDEX='dev_services'
+DISABLE_ANALYTICS_STRING='true'
```
bensinjin: Maybe @rasmus-storjohann-PG should weigh in on what sensible defaults for these should be I don't think they're meant to reflect boolean values.

Issue #865, opened 2019-10-08
Sort Recommended first in Related Topics


cherrielam:
In Related Topics, we would like to see the Recommended topics sorted above the non-recommended. (Currently it appears to be sorting alphabetically without regarding for recommended or not).

![IMG_5422.jpg](https://images.zenhubusercontent.com/5823860ee40e5714b1700133/b03c6788-9346-454d-9ae0-8834cd08714b)

### Rationale:
Recommended topics should have higher importance to the user by nature of being "recommended". The fact that they are sorted higher also helps to educate the user that the Recommended star means it is more important to me

### Difficulty, impact:
This is not hard

### Acceptance Criteria:
- [x] Related Topics are sorted such that the Recommended ones appear at the top.

### Steps for QA:
1. Select a topic from the home screen or the explore screen.
2. Scroll down to where it says 'Related Topics'
- [ ] Ensure that recommended topics are listed first.

jparlina:
Tested on Staging V1.17.0, Server V1.3.23 on iPhone 7s (iOS 13.3) and Android 6.0.1:
- [x] Related Topics are sorted such that the Recommended ones appear at the top.

**iOS:**
![iOS_SortRecommendedFirst_RelatedTopics](https://user-images.githubusercontent.com/48343061/73213948-0a802180-4106-11ea-9d20-f956863b66f9.png)

**Android:**
![Android_SortRecommendedFirst_RelatedTopics](https://user-images.githubusercontent.com/48343061/73213974-153ab680-4106-11ea-85b1-bfdc9e2378de.png)

PR #959, opened 2020-01-03
feature/Issue #865 sort recommended topics

tomy-pg:


src/selectors/topics/sort_topic_list.ts
```diff
@@ -1,11 +1,11 @@
-import { Topic } from './topic';
```
rasmus-storjohann-PG: So can this file be deleted now?

src/selectors/topics/types.ts
```diff
@@ -0,0 +1,24 @@
...
+    readonly id: string;
+    readonly title: string;
+    readonly description: string;
+    readonly taxonomyTerms: ReadonlyArray<TaxonomyTermReference>;
+    readonly exploreSection: ExploreSection;
+    readonly isRecommended: boolean;
+    readonly relatedTopics: ReadonlyArray<TopicListItem>;
+}
+
+export interface TopicListItem {
```
rasmus-storjohann-PG: Rename to `RelatedTopic`.

src/selectors/topics/types.ts
```diff
@@ -0,0 +1,24 @@
...
+    readonly id: string;
+    readonly title: string;
+    readonly description: string;
+    readonly isRecommended: boolean;
+    readonly exploreSection: ExploreSection;
+}
+
+export type ExploreTopic = Topic | TopicListItem;
+
+export type ExploreTopicList = ReadonlyArray<Topic> | ReadonlyArray<TopicListItem>;
```
rasmus-storjohann-PG: Create new interface with the properties that the selector code needs, and change the selector code to use that. Remove these two Explore types

src/selectors/topics/select_topic_for_current_explore_section.ts
```diff
@@ -14,8 +14,9 @@ import { pickTopics } from './pick_topics';
 import { sortTopicList } from './sort_topic_list';
 import { getAllTaxonomyIdsFromAnswers } from '../questionnaire/get_all_taxonomy_ids_from_questionnaire';
 import { pickAnswers } from '../questionnaire/pick_answers';
+import { ExploreTopicList} from './types';
 
-export const selectTopicForCurrentExploreSection = (appStore: Store, routerProps: RouterProps): ReadonlyArray<Topic> => {
+export const selectTopicForCurrentExploreSection = (appStore: Store, routerProps: RouterProps): ExploreTopicList=> {
```
rasmus-storjohann-PG: Keep the return type the way it was.

src/selectors/topics/types.ts
tomy-pg: Removed the two Explore types. 

However, creating an interface with just the properties the sort function needs did not work as expected. 
I created an interface like the following:

```
interface SelectorTopic {
     readonly id: string;
     readonly isRecommended: boolean;
}

```
and included it as the expected type for the sortTopicList function to return. 

When I tried to use sortTopicList while passing either a `Topic` or `TopicListItem`, I received errors like this: 

```
Type ‘readonly SelectorTopic[]’ is not assignable to type ‘readonly Topic[]’.
 Type ‘SelectorTopic’ is missing the following properties from type ‘Topic’: title, description, taxonomyTerms, exploreSection, relatedTopics
```

To get this error to go away, I would have to change many of the selectors to return `SelectorTopic`. We would be less precise with which type we are expecting in this approach and it does not seem like a great trade off. Instead I chose to create a separate sorting function for `TopicListItems` as well.

src/selectors/topics/types.ts
tomy-pg: After taking another look at this, renaming this to RelatedTopic would make the code more difficult to understand. `TopicListItem` is the the type that we use for a recommended topic, explore topic and a bookmarked topic when rendering a topic as part of our UI.

rasmus-storjohann-PG: OK

src/selectors/topics/types.ts
rasmus-storjohann-PG: Yeah, Topic would be assignable to SelectorTopic, but not the other way around. I pick up on the fact that you would need to go both ways.

PR #970, opened 2020-01-09
Feature/issue 956 search navigation bug

ian-lin-PG:


src/components/search/search_input_component.tsx
```diff
@@ -90,7 +100,10 @@ export const SearchInputComponent = (props: Props & Actions): JSX.Element => {
                             debug(`SearchInputComponent location text changed to '${d}'`);
                             setLocationInputField(d);
```
bensinjin: I think this can be removed, we don't need this local state if it's in Redux now. Maybe check nothing like this is happening with the search value as well.

src/components/search/search_input_component.tsx
```diff
@@ -40,7 +44,12 @@ const renderMyLocationButton = (hideMyLocationButton: boolean, setLocation: Func
     return (
         <TouchableOpacity
             style={applicationStyles.whiteButton}
-            onPress={(): void => setLocation('Near My Location')}>
+            onPress={(): void => {
+                // tslint:disable-next-line: no-expression-statement
+                setLocation('Near My Location');
+                // tslint:disable-next-line: no-expression-statement
+                saveLocation('Near My Location');
```
bensinjin: Do we need set and save location?

ian-lin-PG: One `saveLocation` to the redux store, and the other `setLocation` for algolia. I agree the two can probably be streamlined together such that they are one of the same.

rasmus-storjohann-PG:
@ian-lin-PG can you please rebase this PR on develop so I can take it for a spin?

Issue #972, opened 2020-01-10
Bring in Lingui TS types


bensinjin:
Lingui has TS type defs: https://lingui.js.org/guides/typescript.html, however, in our current state if we bring them in we get complaints in places about our custom types conflicting. This should be tidied up.

### Acceptance Criteria 
- [ ]  Lingui TS types installed and custom types removed, places where breakage occurs are fixed

### Steps for QA: 
1.

Issue #892, opened 2019-11-13
Bookmark notification


cherrielam:
The current bookmarks notification is barely usable. On iOS, the notification is down at the bottom and very hard to be seen. On iPhone X + models, the bookmark notification is blocked by the home bar. On Android, the notifications won't even show up. 

We propose a UI change for the bookmark notification so that it is more noticeable and informative for the users. The bookmark is now a dark grey [snackbar](https://reactnativeexample.com/react-native-smart-tip-including-toast-modal-snackbar/) that pops up _vertically above_ the bottom nav. (Reference: https://github.com/PeaceGeeksSociety/pathways-ux-ui/issues/77)

### Mocks
- Zeplin: [See individual screens](https://app.zeplin.io/project/5bc7bfa42e72260acb03aa86/dashboard?seid=5dcb5f0e101ee1bf67e7c315)
- Sketch Cloud: Clickable prototypes - for flow: 
  - [Topic saved/removed](https://sketch.cloud/s/LMGO5/a/zVGyjl/play) 
  - [Service saved/removed](https://sketch.cloud/s/LMGO5/a/JxDAGM/play)

### Acceptance Criteria 
- [ ]  

### Steps for QA: 
1.

jparlina:
Tested on Staging V1.17.0, Server V1.3.23 on:

**1. iPhone 7s (iOS 13.3)**: 
Confirmed with @cherrielam that the wording stays as "Bookmark added/removed" instead of "Topic saved/removed" and "Service saved/removed" since "Bookmark added/removed" has previously been translated to all languages. 
**ISSUES** (per discussion with @cherrielam and @aron-chen-pg):
- Bookmark is too high (see screenshots) - should be just above the navigation bar just as shown on the sketches. 
- Bookmark stayed too long on the screen (see recorded video [here](https://drive.google.com/file/d/1JV6Z1uFrkBe6yrqEwnXSlHV375nH_Yq5/view?usp=sharing)) 

![IMG_3937](https://user-images.githubusercontent.com/48343061/73209846-ed475500-40fd-11ea-8995-ce0c155bef6e.PNG)


![IMG_3938](https://user-images.githubusercontent.com/48343061/73209847-ed475500-40fd-11ea-9aa2-7e253c723218.PNG)


![IMG_3939](https://user-images.githubusercontent.com/48343061/73209848-ed475500-40fd-11ea-99d0-38c27f3af88b.PNG)


[...]

PR #974, opened 2020-01-10
Issue #892 Better notifications

bensinjin:
This work spun off two new issues:
- https://github.com/pg-irc/pathways-frontend/issues/975
- https://github.com/pg-irc/pathways-frontend/issues/972

These can be completed by anyone.


.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-- "8.3"
+- "10"
```
bensinjin: After pulling the most develop branch I was getting CI problems due to a dependency requiring node >= 10. This fixes those problems and I don't see it being an issue, node 8.3 is quite old.

src/@types/I18n.d.ts
```diff
@@ -1,6 +1,7 @@
 // tslint:disable-next-line:no-class
 declare class I18n {
     t(message: TemplateStringsArray): string;
+    _(message: TemplateStringsArray): string;
```
bensinjin: When we properly bring in Lingui types this will all go away. I have a ticket open for this.

src/components/bookmark_button/bookmark_button_component.tsx
```diff
@@ -35,6 +44,27 @@ const BookmarkIcon = (props: Props): JSX.Element => (
...
+    const toastProps = {
+        style: applicationStyles.toast,
+        textStyle: textStyles.toast,
+    };
+    if (props.isBookmarked) {
+        return (): void => {
+            props.unbookmark();
+            Toast.show({
+                ...toastProps,
+                text: i18n._(t`Bookmark removed`),
```
bensinjin: The new "lingui/macros" dependency allows up to use the `t` macro here. Our string extraction command is able to pick up these translatable strings so we don't have to put them into `<Trans>` components anymore.

src/components/main/main_component.tsx
```diff
@@ -64,39 +63,40 @@ export class MainComponent extends React.Component<Props, State> {
...
-                    <HardwareBackButtonHandlerComponent onHardwareBackButtonPress={this.onHardwareBackButtonPress} />
-                    <AppModalsComponent
-                        isAboutVisible={this.state.isAboutModalOpen}
-                        isDisclaimerVisible={this.state.isDisclaimerModalOpen}
-                        closeAboutModal={this.closeAboutModal}
-                        closeDisclaimerModal={this.closeDisclaimerModal}
-                    />
-                </Container>
-            </Drawer>
+            <Root>
```
bensinjin: For the new notifications toast we need this `<Root>` component. It uses it to position the toast.

src/components/bookmark_button/bookmark_button_component.tsx
```diff
@@ -19,11 +23,19 @@ export interface BookmarkButtonActions {
...
-    <Button onPress={getButtonOnPress(props)} transparent icon>
-        <BookmarkIcon {...props} />
-    </Button>
-);
+export const BookmarkButtonComponent = (props: Props): JSX.Element => {
+    const onButtonPress = props.isBookmarked ? onBookmarkedItemPressed(props.unbookmark) : onUnbookmarkedItemPressed(props.bookmark);
+    return (
+        <I18n>
+            {
+                ({ i18n }: any): JSX.Element =>
```
tomy-pg: I have created a type for the entire i18nRenderProp you could use here temporarily. Although this will be cleaned up when we bring in the new lingui types in the issue you have created. #972

bensinjin: > I think this PR looks good! However, after looking at the mocks, it seems that when we bookmark a topic or a service the message in the notification should either be: `Topic Saved` or `Service Saved` as opposed to `Bookmark Saved`.

Yeah I noticed that, I wanted to double check that was accurate so I'll do that, thanks!

bensinjin:
> I think this PR looks good! However, after looking at the mocks, it seems that when we bookmark a topic or a service the message in the notification should either be: `Topic Saved` or `Service Saved` as opposed to `Bookmark Saved`.

Yeah I noticed that, I wanted to double check that was accurate so I'll do that, thanks!

src/components/bookmark_button/bookmark_button_component.tsx
bensinjin: I actually figured out how to properly type this. See https://github.com/pg-irc/pathways-frontend/pull/974/commits/c759c11aade3ad3be05a2a010fc3570d992cebd2. Thanks for pointing me to it.

PR #986, opened 2020-01-20
Feature/issue 938 seach input ui tweak

ian-lin-PG:


src/components/search/search_input_component.tsx
```diff
@@ -46,6 +44,12 @@ const renderMyLocationButton = (hideMyLocationButton: boolean, setLocation: Func
     );
 };
 
+const getOpacity = (input: string): object => (
+    {
+        opacity: (input === '') ? .6 : 1,
+    }
+);
+
```
tomy-pg: For readability, this function should be placed below the SearchInputComponent. You can also simplify this a bit if you put it inside the SearchInputComponent and do the following:

`const opacity =  !props.currentRefinement ?  .6 : 1 `

ian-lin-PG: Set the opacity inline now

Issue #979, opened 2020-01-16
Add feedback checkboxes and input fields to service detail page


rasmus-storjohann-PG:
### Acceptance Criteria 
- [x] When navigating to Service detail page from the "Suggest an update" modal, the screen has a check box "Suggest an update" for each field ([mock](https://zpl.io/bLngBNG)). 
- [x] When navigating to the same screen through other routes, the checkboxes are not there.
- [x] Checking a check box adds an multiline text input field for the user to make suggestions ([mock with keyboard](https://zpl.io/adwL9kp) and [mock without](https://zpl.io/2E96yex)). 
- [x] Typing in text is in black ([mock](https://zpl.io/brpgmEO)) but after field blurs text is teal ([mock](https://zpl.io/anQ019J)).
- [x] Unchecking the check box hides the input field but does not discard the data. (So if you check the button, type in something, uncheck the button, check the button, what you typed in is still there)

Testing:
- [ ] Tapping the "Suggest an update" button at the top of the screen enables and allows for creation of "suggested updates"

bensinjin:
I'm thinking we'll have a dedicated route for suggesting edits to the service.

aron-chen-pg:
Just reattaching my comment from PR regarding the updated mocks:

aron-chen-pg:
Here is the updated mock for feedback module on Zeplin: https://zpl.io/VQ7nm95

tomy-pg:
Passes QA. Not part of staging app as it this lives on a feature branch.

PR #994, opened 2020-01-24
Issue #979 Service suggestions (feedback module)

bensinjin:
Please test and show this to Aron. I have spoken with him and do expect some styling tweaks we've also talked about the deviations from his original design and reasons why for example on iOS the multiline fields look different (platform related). @aron-chen-pg Please put design updates and questions on this PR and I'll get to them asap.

Note: @tomy-pg is in charge of building the "Suggest an edit" button and can remove the temporary one for testing.

### DIFF
--- a/src/components/services/service_detail_component.tsx
+++ b/src/components/services/service_detail_component.tsx
@@ -1,9 +1,10 @@
-import React from 'react';
+import React, { useState, Dispatch, SetStateAction } from 'react';
+import * as R from 'ramda';
 import { History } from 'history';
 import { Trans } from '@lingui/react';
 import { View, Text, TouchableOpacity } from 'react-native';
 import { Content } from 'native-base';
-import { values, textStyles } from '../../application/styles';
+import { values, textStyles, colors } from '../../application/styles';
 import { DescriptorComponent } from '../content_layout/descriptor_component';
 import { TitleComponent } from '../content_layout/title_component';
 import { MarkdownBodyComponent } from '../content_layout/markdown_body_component';
@@ -21,6 +22,17 @@ import { buildServiceName } from '../services//build_service_name';
 import { openInMapsApplication } from '../maps_application_popup/open_in_maps_application';
 import { getLocationTitleFromAddresses } from '../services/get_location_title_from_addresses';
 import { EmailComponent } from '../email/email_component';
+import { SuggestUpdateComponent } from '../suggest_update/suggest_update_component';
+
+type SuggestedUpdates = Partial<{
+    readonly nameSuggestion: string;
+    readonly organizationSuggestion: string;
+    readonly descriptionSuggestion: string;
+    readonly addressSuggestion: string;
+    readonly phoneSuggestion: string;
+    readonly websiteSuggestion: string;
+    readonly emailSuggestion: string;
+}>;
 
 export interface ServiceDetailProps {
     readonly history: History;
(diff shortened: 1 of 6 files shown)


aron-chen-pg:
Based on the conversation I had with Ben, I've come up with an alternative design for Feedback Module. Essentially, all the information is sectioned into different box containers and “suggest an update” + the checkbox is position below instead of above the data. That way we don’t need to provide custom labels (except the ones that don’t have one, i.e. service name and service description). We can also avoid the confusion of having only one suggestion for multiple phone numbers since they will all be grouped in the same box.

src/components/suggest_update/suggest_update_component.tsx
```diff
@@ -0,0 +1,115 @@
...
+}
+
+const ToggleInputComponent = (props: ToggleInputComponentProps): JSX.Element => {
+    const iconName = props.isEditing ? 'check-square' : 'square-o';
+    const color = props.isEditing ? colors.teal : colors.darkGreyWithAlpha;
+    const textStyle = [textStyles.headline6, { color }];
+
+    return (
+        <View style={{ flexDirection: 'row', alignItems: 'center', justifyContent: 'space-between' }} >
+            <Text style={textStyle}>
```
tomy-pg: From the mocks, I think the style for the label is supposed to be different from the "Suggest an update" text.

src/components/suggest_update/suggest_update_component.tsx
```diff
@@ -0,0 +1,115 @@
...
+                {props.label}
+            </Text>
+            <TouchableOpacity onPress={props.onPress}>
+                <View style={{ flexDirection: 'row', alignItems: 'center' }}>
+                    <Text style={textStyle}>
```
tomy-pg: This should also not be bolded when it is not in focus.

src/components/suggest_update/suggest_update_component.tsx
```diff
@@ -76,22 +75,37 @@ interface EditingInputComponentProps {
...
-            style={{
-                borderColor: colors.teal,
-                borderWidth: 1,
-                borderRadius: 5,
-                padding: 5,
-            }}
-        />
+        <I18n>
+        {
+            (({ i18n }: { readonly i18n: I18n }): JSX.Element =>
```
rasmus-storjohann-PG: We have a type for this already somewhere

src/components/services/service_detail_component.tsx
```diff
@@ -103,8 +103,8 @@ const ServiceOrganization = (props: ServiceOrganizationProps): JSX.Element => (
         <DividerComponent />
         <SuggestUpdateComponent
             isEnabled={props.isSuggestingUpdates}
-            onChangeText={props.setSuggestedUpdateForField('nameSuggestion')}
-            textValue={props.suggestedUpdates.nameSuggestion || ''}
+            onChangeText={props.setSuggestedUpdateForField('organizationSuggestion')}
+            textValue={props.suggestedUpdates.organizationSuggestion || ''}
```
rasmus-storjohann-PG: Not a big problem, but this kind of mop-up commit I would usually squash with another commit to make for a more coherent PR

src/components/services/service_detail_component.tsx
```diff
@@ -30,35 +42,90 @@ export interface ServiceDetailProps {
 type Props = ServiceDetailProps & RouterProps;
 
 export const ServiceDetailComponent = (props: Props): JSX.Element => {
+    const [isSuggestingUpdates, setIsSuggestingUpdates]: readonly [boolean, Dispatch<SetStateAction<boolean>>] = useState(false);
+    const [suggestedUpdates, setSuggestedUpdates]: readonly [SuggestedUpdates, Dispatch<SetStateAction<SuggestedUpdates>>] = useState({});
+    const setSuggestedUpdateForField = R.curry((field: string, value: string): void => (
+        setSuggestedUpdates({...suggestedUpdates, [field]: value})
+    ));
```
rasmus-storjohann-PG: You can make this typesafe by passing in a call back to each suggest update component. Haven't tried it... but it seems like it should be possible.

src/components/services/service_detail_component.tsx
```diff
@@ -1,9 +1,10 @@
-import React from 'react';
+import React, { useState, Dispatch, SetStateAction } from 'react';
```
rasmus-storjohann-PG: This file is getting very long, could be good to pull out the helper components into their own files.

rasmus-storjohann-PG:
Found a UX issue in the PR: The issue is that for fields close to the bottom of the service detail page, if I edit them, then they are stuck so close to the bottom of the screen that the keyboard covers up what I'm trying to type

src/components/suggest_update/suggest_update_component.tsx
bensinjin: The designs are currently in flux so I'll hold off on these two comments.

src/components/services/service_detail_component.tsx
```diff
@@ -30,35 +42,89 @@ export interface ServiceDetailProps {
...
             <BannerImageComponent imageSource={undefined} />
+            <SuggestAnUpdateButton onPress={onSuggestAnUpdatePress} />
             <DescriptorComponent descriptor={<Trans>SERVICE</Trans>} />
-            <TitleComponent title={props.service.name} />
-            <ServiceOrganization history={props.history} name={props.service.organizationName} />
-            <MarkdownBodyComponent body={props.service.description} shouldBeExpandable={true} />
+            <SuggestUpdateComponent
+                isEnabled={isSuggestingUpdates}
+                onChangeText={setSuggestedUpdateForField('nameSuggestion')}
+                textValue={suggestedUpdates.nameSuggestion || ''}
```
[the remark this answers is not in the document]

src/components/suggest_update/suggest_update_component.tsx
bensinjin: This type is actually stronger/more accurate. We have an open issue around how we need to clean up our custom Lingui types.

src/components/services/service_detail_component.tsx
bensinjin: I was able to use `keyof` here to ensure you can only pass keys that exist on the `SuggestedUpdates` type.

bensinjin:
Once we have finalized design from @cherrielam I'll be opening a new PR against the `feedback_longlived_feature_branch`

src/components/services/service_detail_component.tsx
bensinjin: I agree and this file is set up quite nicely to be split up - should we do this now? I wonder if it may constitute a refactor that we may want to avoid at this point? You make the call.

bensinjin:
> Found a UX issue in the PR: The issue is that for fields close to the bottom of the service detail page, if I edit them, then they are stuck so close to the bottom of the screen that the keyboard covers up what I'm trying to type

Fixed

bensinjin:
As it stands Cherrie has not signed off on Aron's designs so we're waiting on that, everything else here has been addressed.

src/components/services/service_detail_component.tsx
rasmus-storjohann-PG: Very true, let's not do this now.

src/components/services/service_detail_component.tsx
rasmus-storjohann-PG: Nice

rasmus-storjohann-PG:
Closing this PR, making a new against the correct branch

PR #1001, opened 2020-02-05
Feature/issue 724 disable google analytics

tomy-pg:


src/components/addresses/addresses_component.tsx
```diff
@@ -1,3 +1,4 @@
+ // tslint:disable: no-expression-statement
```
rasmus-storjohann-PG: It's just the one line that needs this, put the exception there

src/components/addresses/addresses_component.tsx
```diff
@@ -22,34 +28,43 @@ export const AddressesComponent = (props: Props): JSX.Element => (
...
-            type={'FontAwesome'}
-            style={{ color: colors.teal, fontSize: values.smallIconSize, paddingRight: 10 }}
-        />
-    );
+    const onPress = (): Promise<void> => {
+        if (props.latLong) {
+            const linkType = 'Button';
+            const linkValue = 'Open in maps';
+            props.analyticsLinkPressed(props.currentPathForAnalytics, props.linkContextForAnalytics, linkType, linkValue);
+            return openInMapsApplication(props.locationTitle, props.latLong.lat, props.latLong.lng);
```
rasmus-storjohann-PG: The type system can't help us against the bug that happens if we mix up lat and long here. It's preferable to pass in an object with the lat/long properties on it, that way there are 2 places where mix ups can happen (when we set the values and when we read them), whereas with this approach there are N places, each time they are passed through a function. Here it's easy to make this a lot safer by doing
```ts
return openInMapsApplication(props.locationTitle, props.latLong);
```

src/components/phone_numbers/phone_numbers_component.tsx
```diff
@@ -20,40 +21,30 @@ interface Props {
...
+const buildPhoneNumber = R.curry((props: Props, phoneNumber: PhoneNumber, index: number): JSX.Element => {
     const onPress = (): void => {
-        const linkValue = 'tel: ' + props.phoneNumber.phone_number;
+        const linkValue = 'tel: ' + phoneNumber.phone_number;
         props.analyticsLinkPressed(props.currentPathForAnalytics, props.linkContextForAnalytics, LinkTypes.phone, linkValue);
         openURL(linkValue);
     };
     return (
-        <View key={props.index}>
+        <View key={index}>
```
rasmus-storjohann-PG: Here you can also use the phone number as index, I think, and remove the index

src/components/phone_numbers/phone_numbers_component.tsx
```diff
@@ -70,4 +64,12 @@ const renderPhoneNumber = (phoneNumber: PhoneNumber): JSX.Element => {
...
\ No newline at end of file
+};
+
+const renderIcon = (): JSX.Element => (
+    <Icon
+        name={'phone'}
+        type={'FontAwesome'}
+        style={{ color: colors.teal, fontSize: values.smallIconSize, paddingRight: 10 }}
+    />
+);
```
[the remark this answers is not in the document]
tomy-pg: Yes I will clean this up. There are four components in the service detail page that use pretty much the same icon component.

PR #1008, opened 2020-02-10
Feature/issue 913 location permission button

ian-lin-PG:


src/components/search/search_input_component.tsx
```diff
@@ -49,7 +60,7 @@ const renderMyLocationButton = (showMyLocationButton: boolean, saveLocation: Fun
                 Permissions.askAsync(Permissions.LOCATION).then((response: object) => {
```
tomy-pg: You can use `PermissionResponse `as opposed to `object` for the `response` type

src/components/search/search_input_component.tsx
```diff
@@ -49,7 +60,7 @@ const renderMyLocationButton = (showMyLocationButton: boolean, saveLocation: Fun
...
                     if (Object(response)['granted']) {
```
tomy-pg: You can also use the Permissions.PermissionsStatus enum as well for this logic. It will look something like
```
if (response.status === Permissions.PermissionStatus.GRANTED) {
...
```

src/components/search/search_input_component.tsx
```diff
@@ -49,7 +60,7 @@ const renderMyLocationButton = (showMyLocationButton: boolean, saveLocation: Fun
...
                         saveLocation('Near My Location');
-                    }
+                    } else toLocationSetting();
```
tomy-pg: I think moving the logic in the onPress to its own function to be called here instead would be cleaner.

src/components/search/search_input_component.tsx
ian-lin-PG: Addressed and used in a switch case.

src/components/search/search_input_component.tsx
ian-lin-PG: Done

src/components/search/search_input_component.tsx
```diff
@@ -55,24 +47,49 @@ const renderMyLocationButton = (showMyLocationButton: boolean, saveLocation: Fun
...
+    const status = await getPermission();
+    switch (status) {
+        case Permissions.PermissionStatus.GRANTED: saveLocation('Near My Location');
+            break;
+        case Permissions.PermissionStatus.DENIED: openAppSettings();
+            break;
+        case Permissions.PermissionStatus.UNDETERMINED: askPermission(saveLocation);
+            break;
+        default:
+            return;
```
rasmus-storjohann-PG: Line breaks after `:` is nicer

src/components/search/search_input_component.tsx
```diff
@@ -55,24 +47,49 @@ const renderMyLocationButton = (showMyLocationButton: boolean, saveLocation: Fun
...
+        IntentLauncher.startActivityAsync(
+            IntentLauncher.ACTION_APPLICATION_SETTINGS,
+        );
+    } else openURL('app-settings:');
+};
+
+const askPermission = async (saveLocation: Function): Promise<void> => {
+    Permissions.askAsync(Permissions.LOCATION).then((permissionResponse: Permissions.PermissionResponse) => {
+        if (permissionResponse.status === Permissions.PermissionStatus.GRANTED) {
+            saveLocation('Near My Location');
```
rasmus-storjohann-PG: String should be defined in constants file, don't want a typo to create a bug

src/components/search/search_input_component.tsx
```diff
@@ -25,6 +28,14 @@ export interface Actions {
     readonly setLocation: (s: string) => void;
 }
 
+const toLocationSetting = (): void => {
```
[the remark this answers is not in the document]
ian-lin-PG: The function was renamed to `openAppSettings()` in a later commit. Both android and iOS redirects to app settings where user can update their permissions.

src/components/search/search_input_component.tsx
```diff
@@ -79,7 +83,9 @@ const openAppSettings = (): void => {
         IntentLauncher.startActivityAsync(
             IntentLauncher.ACTION_APPLICATION_SETTINGS,
         );
-    } else { openURL('app-settings:'); }
+    } else {
+        openURL('app-settings:');
+    }
```
rasmus-storjohann-PG: +1

Issue #1004, opened 2020-02-06
Suggestion Page for Feedback Module


aron-chen-pg:
Some terminology to be defined:

- **Service detail**: Content regarding a particular service. For instance, service name, service provider, service description, address, phone, etc...
- **Service Detail Page**: The UI that is used to show the service detail regarding a particular service.
- **Suggestion Page**: The UI that is used to make suggestions to the service details regarding a particular service.

The Suggestion Page for Feedback Module provides users with the ability to make suggestions to the individual parts of the service detail (i.e. service name, description, address, etc..). Unlike the Service Detail Page, the suggestion page should not have a clear visual hierarchy (i.e. Service name owns the heaviest visual weight). All service detail shown within the Suggestion Page should have the same visual treatment so that the users can easily skim through all the content without being distracted. 

Here is the updated mock: https://zpl.io/VDwj8kw

### Acceptance Criteria 
- [ ]  Checking the individual checkbox will expand and reveal the input field for the user to type in their suggestion for individual parts. Once the editing is done (the keyboard is closed), the user input will appear in teal colour.
- [ ]  Individual labels (both customized or existing labels) and description should have the same appearances (same font style, same font size)
- [ ]  Clear divider (grey, extends full width) to section off different parts of the service detail. 

### Steps for QA: 
1.

bensinjin:
Notes from talk from Aron:

- The light grey divider now replaces the existing tan one, this is a universal change and should appear full width everywhere
- The labels in the designs are only in edit mode
- We really need to be able to render these fields as consistently as possible which means dropping styling (ie H1s) and making it look as close to the designs as possible
- The entire "Suggest an update" with checkbox section should be clickable

bensinjin:
Both issues outlined here have their own tickets now: #1016 & #1017

PR #1010, opened 2020-02-11
Issue #1004 Improve suggest edits mode

bensinjin:
After extensive discussion with Aron and Cherrie we've come to the conclusion we need greater control over how we render existing service fields when we're in edit mode. This PR addresses the following:

- In "edit" mode field labels are consistently styled and fields missing labels are given labels
- In "edit" mode field values are consistently styled, previously certain fields (ie. the title) would still render larger, we do not want to draw attention to any specific field so we need them to all look the same
- Dividers (horizontal rules) are updated app wide to be full width and grey

These changes should match designs outlined here: 
https://app.zeplin.io/project/5bc7bfa42e72260acb03aa86/screen/5e0fe741ac79931c869ce48c

Please review this ticket: https://github.com/pg-irc/pathways-frontend/issues/1004 for additional context.


src/components/phone_numbers/phone_numbers_component.tsx
```diff
@@ -20,25 +21,27 @@ interface Props {
...
     const onPress = (): void => {
         const linkValue = 'tel: ' + phoneNumber.phone_number;
         props.analyticsLinkPressed(props.currentPathForAnalytics, props.linkContextForAnalytics, LinkTypes.phone, linkValue);
         openURL(linkValue);
     };
+    const shouldAddDivider = index !== 0;
+
     return (
         <View key={phoneNumber.phone_number}>
+            {shouldAddDivider && <DividerComponent />}
```
rasmus-storjohann-PG: This logic is repeated quite a bit, pass in the index to the DividerComponent and have it do nothing when index is zero. Component obviously needs a name to reflect this behaviour.

src/components/services/service_detail_component.tsx
```diff
@@ -59,36 +60,39 @@ export const ServiceDetailComponent = (props: Props): JSX.Element => {
...
             <SuggestUpdateComponent
-                isEnabled={isSuggestingUpdates}
-                onChangeText={setSuggestedUpdateForField('nameSuggestion')}
-                textValue={suggestedUpdates.nameSuggestion}
-                label={<Trans>Name</Trans>}
-                style={{ paddingHorizontal: values.backgroundTextPadding }}
+                onChangeSuggestionText={setSuggestedUpdateForField('nameSuggestion')}
+                suggestionText={suggestedUpdates.nameSuggestion}
+                fieldLabel={<Trans>Name</Trans>}
+                suggestingEnabled={isSuggestingUpdates}
```
rasmus-storjohann-PG: Avoid using different terms for the same thing.

src/components/services/service_detail_component.tsx
```diff
@@ -59,36 +60,39 @@ export const ServiceDetailComponent = (props: Props): JSX.Element => {
...
-                isSuggestingUpdates={isSuggestingUpdates}
-                setSuggestedUpdateForField={setSuggestedUpdateForField}
-                suggestedUpdates={suggestedUpdates}
+            <DividerComponent />
+            <SuggestUpdateComponent
+                onChangeSuggestionText={setSuggestedUpdateForField('organizationSuggestion')}
+                suggestionText={suggestedUpdates.organizationSuggestion}
+                fieldLabel={<Trans>Organization</Trans>}
+                suggestingEnabled={isSuggestingUpdates}
+                suggestingEnabledComponent={<SuggestingEnabledComponent value={props.service.organizationName} />}
```
rasmus-storjohann-PG: So the design now is that all the different fields will look the same when in suggestion mode, so you should always be able to just pass in the value here and have `SuggestUpdateComponent` use the `SuggestingEnabledComponent` internally. This will simplify code in 6 places.

src/components/suggest_update/suggest_update_component.tsx
```diff
@@ -1,39 +1,42 @@
...
-    readonly style?: ViewStyle;
+    readonly suggestingEnabled: boolean;
+    readonly suggestionText: string;
+    readonly onChangeSuggestionText: (text: string) => void;
+    readonly fieldLabel: JSX.Element;
+    readonly suggestingEnabledComponent: JSX.Element;
+    readonly suggestingDisabledComponent: JSX.Element;
 }
 
 export const SuggestUpdateComponent = (props: Props): JSX.Element => {
```
rasmus-storjohann-PG: This component will have two quite different behaviours depending on a bool. This is a good use case for the strategy or policy pattern. All the parts of the behaviour that depends on the value of the bool are delegated to the policy object, and the policy object will be one of two different classes depending on the bool

src/components/services/service_detail_component.tsx

src/components/services/service_detail_component.tsx

src/components/phone_numbers/phone_numbers_component.tsx
bensinjin: This is a special case that we likely only require for phone numbers but also include for addresses. I don't think we will ever get back more than one physical address, although it's technically possible. I'm not sure changing the Divider for these cases is necessary.

src/components/suggest_update/suggest_update_component.tsx
bensinjin: Could you suggest an implementation? I did read up on the strategy pattern but I'm not sure how implementing it here would clean this up, we'd still have to pass the flag and the two built components to likely a function to determine our output although

src/components/suggest_update/suggest_update_component.tsx
```diff
@@ -1,84 +1,114 @@
...
-                {props.label}
+interface ToggleButtonComponentProps {
+    readonly onPress: () => void;
+    readonly isEditing: boolean;
+}
+
+const ToggleTextInputComponent = (props: ToggleButtonComponentProps): JSX.Element => (
+    <TouchableOpacity onPress={props.onPress}>
+        <View style={{ flexDirection: 'row', alignItems: 'center', justifyContent: 'space-between' }}>
+            <Text style={getEditingTextStyle(props.isEditing)}>
```
rasmus-storjohann-PG: `            <Text style={props.suggestionMode.getEditingTextStyle()}>`

src/components/suggest_update/suggest_update_component.tsx
```diff
@@ -1,84 +1,114 @@
...
-    const color = props.isEditing ? colors.teal : colors.darkGreyWithAlpha;
-    const textStyle = [textStyles.headline6, { color }];
+const ToggleInputComponent = (props: ToggleInputComponentProps): JSX.Element => (
+    <View
+        style={{
+            borderColor: getEditingColor(props.isEditing),
+            borderWidth: 1,
+            borderRadius: 10,
+            padding: 10,
+        }}
```
rasmus-storjohann-PG: `style={props.editingMode.getTextStyle()}`

src/components/suggest_update/suggest_update_component.tsx
```diff
@@ -1,84 +1,114 @@
...
-                    <Icon
-                        type={'FontAwesome'}
-                        name={iconName}
-                        style={{ padding: 5, color }}
-                    />
-                </View>
-            </TouchableOpacity>
+            <Icon
+                type={'FontAwesome'}
+                name={getEditingIcon(props.isEditing)}
```
rasmus-storjohann-PG: `name={props.editingMode.getEditingIcon()}`

src/components/suggest_update/suggest_update_component.tsx
```diff
@@ -91,25 +121,38 @@ const InputComponent = (props: InputComponentProps): JSX.Element => {
...
+        [ textStyles.paragraphStyle, { color: getEditingColor(isEditing) }]
+);
+
+const getEditingColor = (isEditing: boolean): string => (
+    isEditing ? colors.teal : colors.grey
+);
+
+const getEditingIcon = (isEditing: boolean): string => (
+    isEditing ? 'check-square' : 'square-o'
+);
```
rasmus-storjohann-PG: So these three functions would become methods on the two editingMode objects, one to be used when editing, and the other when not editing

bensinjin: So there are two states here: "Enabled" - via the button on the service detail and "Editing" - via the checkbox after enabled is pressed. Your first comment to me seemed to address the fact that one boolean passed to the `SuggestUpdate` component drastically affected the output, which maybe could be done better. Your comments here talk to styles that are applied when we're in "Editing" mode after we're already "Enabled". I've tried to tidy up these functions to produce one of two objects based on this boolean - this object drives the styling. This feel closer to what you're looking for but feel free to get in there yourself and I can review your approach.

Issue #1017, opened 2020-02-14
Notifications stay up too long on iPhone


bensinjin:
Created from https://github.com/pg-irc/pathways-frontend/issues/892#issuecomment-579996788 notification time open should be less for iOS

### Acceptance Criteria 
- [ ]  

### Steps for QA: 
1.

tomy-pg:
Staging version 1.21.0: 
Pass. Notification duration is 2 seconds.

PR #1018, opened 2020-02-14
Issue #1017 Adjust notification time

bensinjin:
So I tested this side by side on Android and iOS simulators and I think 2 seconds is a good amount of time for both platforms. However, I'm told that maybe on some physical Android devices the notification popup is too short. Please check and let me know!


rasmus-storjohann-PG:
@tomy-pg and @bensinjin please keep in mind that github issues are forever, pull requests are ephemeral, so any substantive discussion of the issue, such as the time duration, should always go on the issue so that we keep that record. Rule of thumb, discussions of what to change -> issue, discussion of how to change it -> PR.

PR #1021, opened 2020-02-21
Feature/issue 957 shrink input fields

ian-lin-PG:


src/components/search/search_input_component.tsx
```diff
@@ -167,7 +167,11 @@ export const SearchInputComponent = (props: Props & Actions): JSX.Element => {
                     <Text> near </Text>
                     <Text style={{ fontWeight: 'bold' }}>{locationInputField}</Text>
                 </View>
-                <ClearInputButton visible={true} onPress={clearSearch} />
+                <ClearInputButton visible={true} onPress={(): void => {
+                    props.saveSearchTerm('');
+                    props.setLocation('');
+                    props.setCollapseInput(false);
```
tomy-pg: Breaking this part out into its own onPress function, similar to the previously used `clearSearch` function, would make this easier to read.

src/components/search/search_input_component.tsx
```diff
@@ -154,56 +157,80 @@ export const SearchInputComponent = (props: Props & Actions): JSX.Element => {
...
-                        placeholder={buildTranslatedPlaceholder(i18nRenderProp.i18n, 'Enter city, address, or postal code')} // TODO translate
-                        placeholderTextColor={colors.greyishBrown}
-                        selectionColor={colors.black}
-                    />
-                    <ClearInputButton visible={locationInputField !== ''} onPress={clearLocation} />
-                </TouchableOpacity>
-                <View style={{ flexDirection: 'row-reverse', display: 'flex', justifyContent: 'space-between', margin: 4 }}>
-                    {renderSearchButton(showSearchButton, searchInputField, locationInputField, props.saveSearchTerm, props.setLocation)}
-                    {renderMyLocationButton(showMyLocationButton, setLocationInputField)}
+    const smallInput = (
```
tomy-pg: Small input / Big input are too ambiguous I think. Changing this to something like`collapsedInput` / `uncollapsedInput` could help make this clearer.

src/components/search/search_component.tsx
```diff
@@ -49,6 +50,7 @@ export const SearchComponent = (props: Props): JSX.Element => {
     const [location, setLocation]: readonly [string, (s: string) => void] = useState(props.searchLocation);
     const [latLong, setLatLong]: readonly [LatLong, (latLong: LatLong) => void] = useState(undefined);
     const [isLatLongLoading, setIsLatLongLoading]: readonly [boolean, SetIsLatLongLoading] = useState(false);
+    const [collapseInput, setCollapseInput]: readonly [boolean, SetIsInputCollapsed] = useState(false);
```
tomy-pg: You are using the name `SetIsInputCollapsed` for the type of setCollapseInput. For consistency, you can make these names the same. `SetInputCollapsed` feels to me like the better name as it is easier to read. You can rename the state to 
`const [isInputCollapsed, setIsInputCollapsed]`

tomy-pg: You are using the name `SetIsInputCollapsed` for the type of setCollapseInput. For consistency, you can make these names the same. `SetInputCollapsed` feels to me like the better name as it is easier to read. You can rename the state to 
`const [isInputCollapsed, setIsInputCollapsed]`

src/components/search/search_input_component.tsx
```diff
@@ -227,13 +224,30 @@ export const SearchInputComponent = (props: Props & Actions): JSX.Element => {
...
     if (props.collapseInput) {
-        return <View>{collapsedInput(props, searchInputField, locationInputField, getTranslatedSnapshot)}</View>;
+        return <View>{collapsedInput(props, searchInputState)}</View>;
     }
-    return <View>{expandedInput(props, searchInputField, setSearchInputField, locationInputField,
-        setLocationInputField, buildTranslatedString)}</View>;
+    return <View>{expandedInput(props, searchInputState)}</View>;
 };
 
+interface SearchInputState {
```
tomy-pg: I think this should be re-ordered so that it is closer to where it is first being used. A reader will not know what SearchInputState is until they scroll all the way down the file.

src/components/search/search_input_component.tsx
```diff
@@ -140,98 +220,18 @@ export const SearchInputComponent = (props: Props & Actions): JSX.Element => {
...
-                        {renderMyLocationButton(setLocationInputField)}
-                    </View>
-                </View >
-            )}
-
-        </I18n>
-    );
     if (props.collapseInput) {
-        return <View>{smallInput}</View>;
+        return <View>{collapsedInput(props, searchInputField, locationInputField, getTranslatedSnapshot)}</View>;
```
tomy-pg: Instead of the one line return, what are your thoughts on doing this instead? I think it is easier to read the JSX this way. 
```
return (
     <View>
          {collapsedInput(props, searchInputField, locationInputField, getTranslatedSnapshot)}
     </View>
)
```

src/components/search/search_input_component.tsx
```diff
@@ -115,21 +115,101 @@ const renderSearchButton = (searchTerm: string, location: string,
...
+                </View>
+            )}
+
+        </I18n>
+    );
+
+const expandedInput = (props: Props & Actions, searchInputField: string, setSearchInputField: Function,
+    locationInputField: string, setLocationInputField: Function, buildTranslatedString: Function): JSX.Element => {
+    const searchInputRef = React.useRef(undefined);
+    const searchLocationRef = React.useRef(undefined);
```
tomy-pg: You can import `useRef` on its own here if you import it like you do with `useState`

src/components/search/search_input_component.tsx
```diff
@@ -115,21 +115,101 @@ const renderSearchButton = (searchTerm: string, location: string,
...
+        )}
+
+    </I18n>;
+};
+
 export type SetShowMyLocationButton = Dispatch<SetStateAction<boolean>>;
 
 export type SetShowSearchButton = Dispatch<SetStateAction<boolean>>;
 
 export const SearchInputComponent = (props: Props & Actions): JSX.Element => {
```
tomy-pg: I noticed in this component you have the `SearchInputComponent` ordered below other components that are used inside of it. In our other files, we do the opposite, as we place the main component on top of the file, and the child components below.

src/components/search/search_input_component.tsx
```diff
@@ -59,13 +59,17 @@ export const SearchInputComponent = (props: SearchProps & SearchActions): JSX.El
...
+        );
     }
-    return <View>
-        {expandedInput(searchProps)}
-    </View>;
+    return (
+        <View>
+            {expandedInput(searchProps)}
+        </View>
+    );
```
rasmus-storjohann-PG: Either of these forms are fine to me

Issue #1012, opened 2020-02-12
Service list refreshes every time 'load more services' is clicked


aron-chen-pg:
Every time when the 'load more services' button is clicked, the whole service listing refreshes. Users would lose their scrolling position and have to scroll down again in order to explore more services. The experience gets progressively worse since the list will only grow longer every time the button is clicked, thus consuming more of users' time to scroll. In addition, since all the services are listed in text format without any visual cue, it became incredibly difficult to rediscover the service the users were last on.

The experience is recorded here in the link below:
https://drive.google.com/file/d/1-14TLSFv10B_BO08w3BcDi4nZEa-9GEv/view?usp=sharing

To address the bug, the ideal experience would be to retain the scrolling position upon clicking the 'load more services' button. That way users can simply continue their search without having to spend time to scroll down the list again.

### Acceptance Criteria
- [ ] After clicking onto the 'load more services' button, the scrolling position is retained.

ian-lin-PG:
This issue is related to issue #892. Wrapping the Flatlist and loadmore button in a scrollview sometimes mitigate the issue but is ill advised as it causes warning.

tomy-pg:
This will be an Acceptance Criteria as part if Issue #1058

PR #1033, opened 2020-03-02
Issue #1012 clicking load more services keeps scroll position

tomy-pg:
When Load more services was clicked, a complete re-render of the infiniteHitsComponent was triggered which would start the search service list from the top. In this PR, I overlaid the LoadingServiceList over top of the service list to prevent this from happening.

### DIFF
--- a/src/components/search/infinite_hits_component.tsx
+++ b/src/components/search/infinite_hits_component.tsx
@@ -82,33 +82,30 @@ export const InfiniteHitsComponent = (props: Partial<InfiniteHitsAndStateResults
         return renderErrorComponent(searchErrorProps);
     }
 
-    const renderLoadingScreen = (): JSX.Element => {
-        if (!isLoading(props.searching, props.isLatLongLoading)) {
-            return <EmptyComponent />;
-        }
-        return (
-            <View style={{ height: '100%', width: '100%' }}>
-                <LoadingServiceListComponent />
-            </View>
-        );
-    };
-
-    const ServiceList = (
-        <FlatList
-            style={{ backgroundColor: colors.white }}
-            refreshing={false}
-            data={searchResults}
-            keyExtractor={keyExtractor}
-            renderItem={renderSearchHit(props)}
-            ItemSeparatorComponent={SearchListSeparator}
-            ListHeaderComponent={renderHeader(showPartialLocalizationMessage, hidePartialLocalizationMessage)}
-            ListFooterComponent={loadMoreButton} />
-    );
-
     return (
         <View style={{ flexDirection: 'column', backgroundColor: colors.lightGrey, flex: 1 }}>
-            {renderLoadingScreen()}
-            {ServiceList}
+            {renderLoadingScreen(props.searching, props.isLatLongLoading)}
+            <FlatList
+                style={{ backgroundColor: colors.white }}
+                refreshing={false}
+                data={searchResults}
+                keyExtractor={keyExtractor}
+                renderItem={renderSearchHit(props)}
+                ItemSeparatorComponent={SearchListSeparator}
+                ListHeaderComponent={renderHeader(showPartialLocalizationMessage, hidePartialLocalizationMessage)}
+                ListFooterComponent={loadMoreButton}
+            />
+        </View>
+    );
+};
+
+const renderLoadingScreen = (searching: boolean, isLatLongLoading: boolean): JSX.Element => {
+    if (!isLoading(searching, isLatLongLoading)) {
+        return <EmptyComponent />;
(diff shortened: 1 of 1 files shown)


src/components/search/infinite_hits_component.tsx
```diff
@@ -167,16 +169,17 @@ const isLoading = (searching: boolean, isLatLongLoading: boolean): boolean => (
     searching || isLatLongLoading
 );
 
-const isSearchErrorType = (onlineStatus: OnlineStatus, searchResults: ReadonlyArray<SearchServiceData>, latLong: LatLong): boolean => (
-    isOffline(onlineStatus) || hasNoResultsFromSearchTermQuery(searchResults) || hasNoResultsFromLocationQuery(latLong)
+const isSearchErrorType =
+(onlineStatus: OnlineStatus, searchResults: ReadonlyArray<SearchServiceData>, latLong: LatLong, searching: boolean): boolean => (
+    isOffline(onlineStatus) || hasNoResultsFromSearchTermQuery(searchResults, searching) || hasNoResultsFromLocationQuery(latLong)
```
rasmus-storjohann-PG: Better to use `searching` as part of the logic here, since if it's true, isError should be false, no need to consult `hasNoResultsFromSearchTermQuery()` to figure that out. Boolean arguments to functions are a code smell, since it tends to be a sign that functions do more than one thing, as is the case with `hasNoResultsFromSearchTermQuery()` here.

src/components/search/infinite_hits_component.tsx
```diff
@@ -215,15 +219,14 @@ const ErrorComponent = (props: { readonly errorType: Errors, readonly refreshScr
     />
 );
 
-const determineErrorType = (onlineStatus: OnlineStatus, searchResults: ReadonlyArray<SearchServiceData>, latLong: LatLong): Errors => {
+const determineErrorType =
+(onlineStatus: OnlineStatus, searchResults: ReadonlyArray<SearchServiceData>, latLong: LatLong, searching: boolean): Errors => {
```
rasmus-storjohann-PG: So you should be able to remove `searching` from the signature on this function and things should simplify nicely.

src/components/search/infinite_hits_component.tsx
```diff
@@ -63,40 +63,51 @@ export const InfiniteHitsComponent = (props: Partial<InfiniteHitsAndStateResults
...
+    if (isSearchErrorType(onlineStatus, searchResults, latLong, searching)) {
         const searchErrorProps = {
             onlineStatus,
             searchResults,
             latLong,
             showPartialLocalizationMessage,
             hidePartialLocalizationMessage,
             refreshSearchServices,
+            searching,
         };
```
rasmus-storjohann-PG: This list of properties is getting very long. You could just pass in `props`, and you can leave out a number of these. I'm also thinking that there is some redundancies between `isSearchErrorType`, `renderErrorComponent` and `determineErrorType()`. You might be able to refactor to a new `renderErrorComponent()` that returns a component if there is an error and undefined on no error, so that you can test in `InfiniteHitsComponent` whether there is an error or not, something like
```ts
    const errorComponent = renderErrorComponent(...);
    if (errorComponent) {
        return errorComponent;
    }
```
Then you'll have just one function that both figures out what kind of error you have, and what error component to show. Give it a try...

tomy-pg: I gave this a try and I cut down the redundancies. However, creating a single function that returns either a component or undefined competes with how the loading screen is rendered. I kept the `isSearchErrorType` because of this.

src/components/search/infinite_hits_component.tsx
```diff
@@ -154,8 +149,31 @@ const isLoading = (searching: boolean, isLatLongLoading: boolean): boolean => (
     searching || isLatLongLoading
 );
 
-const isSearchErrorType = (onlineStatus: OnlineStatus, searchResults: ReadonlyArray<SearchServiceData>, latLong: LatLong): boolean => (
-    isOffline(onlineStatus) || hasNoResultsFromSearchTermQuery(searchResults) || hasNoResultsFromLocationQuery(latLong)
+const isSearchErrorType =
+(onlineStatus: OnlineStatus, searchResults: ReadonlyArray<SearchServiceData>, latLong: LatLong, searching: boolean): boolean => (
+    isOffline(onlineStatus) || hasNoResultsFromSearchTermQuery(searchResults) && !searching || hasNoResultsFromLocationQuery(latLong)
```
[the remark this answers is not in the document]
tomy-pg: This way is much clearer. Thank you!

Issue #1030, opened 2020-02-28
Refactor search to not use Algolia instantsearch


rasmus-storjohann-PG:
InstantSearch from algolia gives us grief with no benefit, there is not enough screen area devoted to the search results to be able to tell that the results change while typing, and the location entry field also doesn't make sense with instant search. We also have #1016 which we have not been able to solve using instant search.

Solution is to use a simpler approach where we hit the Algolia REST API directly rather than using the algolia client side library for this. We can then reintroduce the bookmark on/off functionality to the service listings, effectively reverting the change that was made under #1026.

Issue to re-introduce bookmarking can be found here: #1046 

### Acceptance Criteria 
~~Services can again be bookmarked from service search result list, without resetting the list position as happened under #1026.~~
- [x] Search functions are before without performance hit. (extra re-renders)

### Steps for QA: 
1.

ian-lin-PG:
@rasmus-storjohann-PG @tomy-pg I believe this issue can be closed, but can the acceptance criteria be changed. It may cause confusion in the future.

tomy-pg:
@ian-lin-PG the Acceptance criteria seems a bit off here. I will adjust the acceptance criteria for QA.

PR #1045, opened 2020-03-12
Feature/issue #1030 refactor search

tomy-pg:


src/components/search/api/fetch_lat_long_from_location.ts
```diff
@@ -0,0 +1,60 @@
+// tslint:disable:no-expression-statement
+import { OnlineStatus } from '../../../hooks/use_online_status';
+import { LatLong } from '../../../validation/latlong/types';
+import { toGeoCoderLatLong } from '../../../validation/latlong';
+import BuildUrl from 'build-url';
+// import * as R from 'ramda';
```
niko-pg: Left over comment?

src/components/search/search_component.tsx
```diff
@@ -37,57 +36,58 @@ export interface SearchComponentActions {
...
+        let geocoderLatLong = props.searchLatLong;
+        if (props.searchLocation !== location) {
+            geocoderLatLong = await fetchLatLongFromLocation(location, onlineStatus).catch((): undefined => undefined);
+            props.saveSearchLatLong(geocoderLatLong);
         }
-    }, [location]);
-
-    useFetchLatLongFromLocation(location, setLatLong, setIsLatLongLoading);
-    useDisableAnalyticsOnEasterEgg(location, props.disableAnalytics);
+        const searchResults = await fetchSearchResultsFromQuery(searchTerm, geocoderLatLong).catch((): undefined => undefined);
```
niko-pg: Since you're using await, maybe better to wrap with try finally? (no catch block because it seems like you dont do anything for the exceptions.

try {
  const searchResults = await fetchSearchResultsFromQuery(searchTerm, geocoderLatLong);
  props.saveSearchResults(searchResults);
} finally {
   setIsLoading(false);
}

src/components/search/search_component.tsx
```diff
@@ -37,57 +36,58 @@ export interface SearchComponentActions {
...
+                        saveSearchTerm={props.saveSearchTerm}
+                        saveSearchLocation={props.saveSearchLocation}
+                        isSearchInputCollapsed={props.isSearchInputCollapsed}
+                        setIsSearchInputCollapsed={props.setIsSearchInputCollapsed}
+                        onSearchPress={onSearchPress}
+                    />
+                    <SearchResultsComponent {...searchResultsProps} />
+                </View>
+            );
+        }}
```
niko-pg: Any reason for wrapping this as an expression? Maybe a lot of cleaner if we just wrap the whole thing with <I18n> component?

src/components/search/search_input_component.tsx
```diff
@@ -103,11 +107,11 @@ const collapsedInput = (props: Props): JSX.Element => {
                     <TouchableOpacity style={applicationStyles.searchContainer}
                         onPress={(): void => props.setIsSearchInputCollapsed(false)}>
                         <InputIcon name='search' />
-                        <Text numberOfLines={1} style={{ flex: 1 }}>
-                            {combineSearchInputs(props.searchInputField, props.locationInputField, i18nRenderProp)}
+                        <Text numberOfLines={1} style={[textStyles.paragraphStyle, { flex: 1 }]}>
+                            {combineSearchInputs(props.termInput, props.locationInput, i18nRenderProp)}
```
rasmus-storjohann-PG: The name `combineSearchInputs()` confused me, I would call it `buildBriefSeachString()` or something

src/components/search/search_input_component.tsx
```diff
@@ -141,38 +145,39 @@ const expandedInput = (props: Props): JSX.Element => {
                     <TouchableOpacity style={applicationStyles.searchContainer}>
                         <InputIcon name='search' />
                         <TextInput
-                            ref={searchInputRef}
-                            style={[applicationStyles.searchInput, { opacity: getOpacity(props.searchInputField) }]}
-                            onChangeText={(d: string): void => {
-                                debug(`SearchInputComponent search text changed to '${d}'`);
-                                props.setSearchInputField(d);
+                            ref={props.termInputRef}
```
rasmus-storjohann-PG: So now you bubble refs up to the parent component ... this is getting a little bit funky

src/components/search/search_input_component.tsx
```diff
@@ -141,38 +145,39 @@ const expandedInput = (props: Props): JSX.Element => {
...
-                                debug(`SearchInputComponent location text changed to '${d}'`);
-                                props.setLocationInputField(d);
+                            ref={props.locationInputRef}
+                            style={[applicationStyles.searchInput, { opacity: getOpacity(props.locationInput) }]}
+                            onChangeText={(text: string): void => {
+                                debug(`SearchInputComponent location text changed to '${text}'`);
+                                props.setLocationInput(text);
                             }}
-                            value={props.locationInputField}
+                            onFocus= {(): void => props.setShowMyLocationButton(true)}
```
rasmus-storjohann-PG: Consider instead using names like `onFocus= {props.showMyLocationButton}` and `hideMyLocationButton()`

src/components/search/search_input_component.tsx
```diff
@@ -201,16 +206,17 @@ const renderSearchButton = (props: Props): JSX.Element => {
         <TouchableOpacity
             style={[applicationStyles.searchButton, { backgroundColor: colors.lightTeal }]}
             onPress={(): void => {
-                props.setIsSearchInputCollapsed(true);
-                props.saveSearchTerm(props.searchInputField);
-                props.setLocation(props.locationInputField);
+                props.onSearchPress(props.termInput, props.locationInput);
```
rasmus-storjohann-PG: I much prefer this one-liner to the three lines you had before. You could probably change this to `props.onSearchPress();` and then define that function to access the other two values in a closure. We can talk more about that if it's not clear.

src/components/search/search_input_component.tsx
```diff
@@ -228,32 +234,36 @@ const renderMyLocationButton = (saveLocation: Function): JSX.Element => {
...
         default:
             break;
     }
 };
 
 const getPermission = (): Promise<Permissions.PermissionStatus> => {
-    return Permissions.getAsync(Permissions.LOCATION).then((permissionResponse: Permissions.PermissionResponse) => {
+    return Permissions.getAsync(Permissions.LOCATION).then((permissionResponse: Permissions.PermissionResponse): Permissions.PermissionStatus => {
         return permissionResponse.status;
     });
```
rasmus-storjohann-PG: Long lines, let's get into the habit of line breaks like this:
```ts
return Permissions.getAsync(Permissions.LOCATION).
    then((permissionResponse: Permissions.PermissionResponse): Permissions.PermissionStatus => {
        return permissionResponse.status;
    });
```
to mimic regular imperative code

src/components/search/search_results_component.tsx
```diff
@@ -0,0 +1,166 @@
...
+import { OnlineStatus } from '../../hooks/use_online_status';
+import { ErrorScreenSwitcherComponent } from '../error_screens/ErrorScreenSwitcherComponent';
+import { Errors } from '../../validation/errors/types';
+import { EmptyComponent } from '../empty_component/empty_component';
+import { LoadingServiceListComponent } from '../loading_screen/loading_service_list_component';
+import { LatLong } from '../../validation/latlong/types';
+
+export interface SearchResultsProps {
+    // tslint:disable-next-line:no-any
+    readonly searchResults: ReadonlyArray<any>;
```
rasmus-storjohann-PG: If your validation function has the correct return type, this doesn't need to be an `any`.

src/components/search/search_results_component.tsx
```diff
@@ -0,0 +1,166 @@
...
+    readonly searchLatLong: LatLong;
+    readonly isLoading: boolean;
+    readonly onlineStatus: OnlineStatus;
+}
+
+export interface SearchResultsActions {
+    readonly bookmarkService: (service: HumanServiceData) => BookmarkServiceAction;
+    readonly unbookmarkService: (service: HumanServiceData) => UnbookmarkServiceAction;
+    readonly hidePartialLocalizationMessage: () => HidePartialLocalizationMessageAction;
+    readonly onSearchPress: (searchTerm: string, location: string) => Promise<void>;
```
rasmus-storjohann-PG: Can this function be renamed so that it's clearer why it exists. I can see that it's called when I hit search, but why does this need to bubble out of the component?

src/components/search/search_results_component.tsx
```diff
@@ -0,0 +1,166 @@
...
+export const SearchResultsComponent = (props: Props): JSX.Element => {
+    useTraceUpdate('SearchResultsComponent', props);
+
+    if (searchTermIsEmpty(props)) {
+        return renderEmptyComponent(props);
+    }
+
+    if (isSearchErrorType(props)) {
+        return renderErrorComponent(props);
+    }
```
rasmus-storjohann-PG: I like how you broke out these two helper functions. Even nicer to do the same with the happy path so that you here do just `return renderComponentWithResults(props);`, then this function is all high level code and the helpers contain all the low level stuff.

src/components/search/search_results_component.tsx
```diff
@@ -0,0 +1,166 @@
...
+const isOffline = (onlineStatus: OnlineStatus): boolean => (
+    onlineStatus === OnlineStatus.Offline
+);
+
+const hasNoResultsFromSearchTermQuery = (searchResults: ReadonlyArray<SearchServiceData>): boolean => (
+    searchResults.length === 0
+);
+
+const hasNoResultsFromLocationQuery = (latLong: LatLong): boolean => (
+    latLong && latLong.lat === 0 && latLong.lng === 0
```
rasmus-storjohann-PG: if latLong can be null, the type should signify that with `const hasNoResultsFromLocationQuery = (latLong?: LatLong): boolean => (`. It's a bit of funny syntax, but it gets the point accross

src/components/search/search_component.tsx
rasmus-storjohann-PG: I like it

src/components/search/search_component.tsx
```diff
@@ -37,57 +36,58 @@ export interface SearchComponentActions {
     readonly unbookmarkService: (service: HumanServiceData) => UnbookmarkServiceAction;
     readonly saveSearchTerm: (searchTerm: string) => SaveSearchTermAction;
     readonly saveSearchLocation: (searchLocation: string) => SaveSearchLocationAction;
+    readonly saveSearchLatLong: (searchLatLong: LatLong) => SaveSearchLatLongAction;
```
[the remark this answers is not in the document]
ian-lin-PG: I would recommend keeping them separate as we get `geocoderLatLong` asynchronously which saves at a different time than just saving the location.

src/components/search/search_component.tsx
```diff
@@ -37,57 +36,58 @@ export interface SearchComponentActions {
...
+        props.saveSearchTerm(searchTerm);
         props.saveSearchLocation(location);
-        if (location === '') {
-            setLatLong(undefined);
+        setIsLoading(true);
+        // tslint:disable-next-line: no-let
+        let geocoderLatLong = props.searchLatLong;
+        if (props.searchLocation !== location) {
+            geocoderLatLong = await fetchLatLongFromLocation(location, onlineStatus).catch((): undefined => undefined);
+            props.saveSearchLatLong(geocoderLatLong);
```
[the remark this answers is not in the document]
ian-lin-PG: I wouldn't save it in the if case since we don't always have to call the geocoder API. I can put it after the if case such that the `saveSearchTerm` and `saveSearchLocation` are called every time and still close to `saveSearchLatLong`.

src/components/search/search_input_component.tsx
ian-lin-PG: The two `useRefs` used to be in `ExpandedInput` which is part of an if else case. As we discussed  hooks cannot be within any sort of conditional statements. That's why they are being passed in instead of living in a lower level component.

src/components/search/search_input_component.tsx
ian-lin-PG: This search function will be reused for the load more button in the `SearchResultsComponent` which will take the parameters from the store instead of the input fields. Keeping it higher level and reusable in both scenarios.

src/components/search/search_component.tsx
ian-lin-PG: We were using `i18nRenderProp` that belong to the `<I18n>` tag, it's not being used here so we are deleting it now.

src/components/search/search_results_component.tsx
ian-lin-PG: We call this function every time we send an algolia search request which can happen in other components besides the search button. I can rename it `onSearchRequest` if it's any better.

src/components/search/search_input_component.tsx
rasmus-storjohann-PG: Right, makes sense... it's getting complex, need to keep an eye on that and keep our minds open to possible ways to simplify.

src/components/search/search_component.tsx
```diff
@@ -72,25 +71,20 @@ export const SearchComponent = (props: Props): JSX.Element => {
...
-        </I18n>
+        <View style={{ backgroundColor: colors.pale, flex: 1 }}>
+            <SearchInputComponent
+                searchTerm={props.searchTerm}
+                searchLocation={props.searchLocation}
+                saveSearchTerm={props.saveSearchTerm}
+                saveSearchLocation={props.saveSearchLocation}
+                collapseSearchInput={props.collapseSearchInput}
+                setCollapseSearchInput={props.setCollapseSearchInput}
+                onSearchRequest={onSearchRequest}
```
rasmus-storjohann-PG: All the names are consistent, good!

src/components/search/search_input_component.tsx
```diff
@@ -69,35 +69,35 @@ const buildTranslatedString = (i18n: ReactI18n, placeholder: string): string =>
...
     if (searchInput === '') {
         return '';
     }
-    if (locationInput === '') {
+    if (searchLocationInput === '') {
         return searchInput;
     }
-    if (locationInput === MY_LOCATION) {
+    if (searchLocationInput === MY_LOCATION) {
         return searchInput + ' ' + buildTranslatedString(i18nRenderProp.i18n, 'near ' + MY_LOCATION);
```
rasmus-storjohann-PG: Better to translate "near" and "near my location" as two separate strings.

PR #1063, opened 2020-03-18
Feature/issue 981 other feedback screens

niko-pg:


src/components/feedback/feedback_modal_container.tsx
```diff
@@ -0,0 +1,113 @@
...
+				onOtherOptionPress={showOther}
+				onRemoveServicePress={showRemoveService}
+			/>
+			<FeedbackSuggestionComponent
+				isVisible={isVisible}
+				headerLabel={content.header}
+				inputLabel={content.label}
+				placeholder={content.placeholder}
+				onClose={close}
+			/>
```
rasmus-storjohann-PG: I'm struggling with these names `FeedbackOptionsModalComponent` and `FeedbackSuggestionComponent`, which is what exactly?

rasmus-storjohann-PG:
QA: "Suggest an Update" string and "X" button don't line up really well.

src/components/feedback/feedback_modal_container.tsx
niko-pg: `FeedbackOptionsModalComponent` is the popup that opens where you can choose the options `Change name or other details`, `Remove this service` or `Other`

`FeedbackSuggestionComponent` is the screen I did for `Remove for service` and `Other`. I'm okay with changing the name. What do you think of `FeedbackOtherRemoveServiceModal`? I guess that's beneficial from the developer's standpoint?

src/components/feedback/feedback_modal_container.tsx
```diff
@@ -0,0 +1,113 @@
...
+			toggle();
+			setContent(SUGGESTION_CONTENT.OTHER);
+		},
+		[toggle, setContent],
+	);
+
+	const close = useCallback<() => void> (
+		() => toggle(),
+		[toggle],
+	);
```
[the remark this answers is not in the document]
niko-pg: Great point. I'm going to actually move `useToggleState` into the common hooks directory because I believe that this could be reused. I will be making improvements to the api to account for your comments

src/components/feedback/feedback_modal_container.tsx
```diff
@@ -0,0 +1,113 @@
+import React, { useCallback, useState } from 'react';
+
+import { FeedbackOptionsModalComponent } from './feedback_options_modal_component';
+import { FeedbackSuggestionComponent } from './feedback_suggestion_component';
+
+interface SuggestionContent {
+ 	header: string;
+	label: string;
+	placeholder: string;
+};
```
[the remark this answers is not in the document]
niko-pg: Thanks for the reminder. Just noticed my vscode doesn't have `tslint` automatically configured. It's only configured for `eslint`,  will be turning it on now.

src/components/feedback/feedback_modal_container.tsx
```diff
@@ -0,0 +1,113 @@
...
+	},
+};
+
+const useToggleState = (initialValue: boolean): readonly [boolean, () => void] => {
+	const [state, toggleState] = useState<boolean>(initialValue);
+
+	const toggle = useCallback(
+		() => toggleState(prevState => !prevState),
+		[toggleState],
+	);
```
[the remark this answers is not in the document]
niko-pg: Done

src/components/feedback/feedback_modal_container.tsx
```diff
@@ -0,0 +1,109 @@
...
+const useFeedbackSuggestionContent = (): UseFeedbackContent => {
+    const {
+        state: isVisible,
+        on: showModal,
+        off: hideModal,
+    }: ToggleStateApi = useToggleState();
+    const [content, setContent]: readonly[SuggestionContent, Dispatch<SetStateAction<SuggestionContent>>]
+        = useState<SuggestionContent>(SUGGESTION_CONTENT.OTHER);
+
+    const showRemoveService = useCallback<() => void>(
```
bensinjin: Are you using `useCallback` for these functions primarily because you're passing them into child components and you want to limit re-renders? What's your metric for when you see the need for `useCallback`?

src/components/feedback/feedback_other_remove_service_component.tsx
```diff
@@ -0,0 +1,128 @@
...
+                ({ i18n }: I18nProps): JSX.Element => (
+                    <Content padder>
+                        <Item placeholderLabel={true} stackedLabel>
+                            <Label style={styles.inputLabel}>
+                                <Trans id ={props.inputLabel} />
+                            </Label>
+                            <Input
+                                multiline
+                                numberOfLines={5}
+                                placeholder={i18n._(t`${props.placeholder}`)}
```
bensinjin: I don't think this placeholder will be translatable, you need to hardcode in the string for lingui to pick it up.

src/components/feedback/feedback_modal_container.tsx
```diff
@@ -0,0 +1,109 @@
...
+        placeholder: 'Comment or suggest edits',
+    },
+    REMOVE_SERVICE: {
+        header: 'Remove Service',
+        label: 'What is your reason for removal?',
+        placeholder: 'e.g. Service is permanently closed',
+    },
+};
+
+const useFeedbackSuggestionContent = (): UseFeedbackContent => {
```
bensinjin: What's your reasoning for a custom hook here? My impression was we create them when we require the same piece of reusable functionality in more than one component, totally just curious on what you think.

src/components/feedback/feedback_modal_container.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+interface UseFeedbackContent {
+    readonly content: SuggestionContent;
+    readonly isVisible: boolean;
+    readonly showRemoveService: () => void;
+    readonly showOther: () => void;
+    readonly close: () => void;
+}
+
+type FeedbackOptionsModalProps = {
```
bensinjin: Notice the jump between types and interfaces, reason?

src/components/feedback/feedback_other_remove_service_component.tsx
```diff
@@ -0,0 +1,128 @@
...
+        </Footer>
+    );
+};
+
+type FeedbackSuggestionProps = HeaderComponentProps & ContentComponentProps & ModalProps;
+
+export const FeedbackOtherRemoveServiceModal = (props: FeedbackSuggestionProps): JSX.Element => {
+    return (
+        <Modal isVisible={props.isVisible} style={{ margin: 0 }}>
+            <Container>
```
bensinjin: I think the native base docs suggest one `<Container>` per screen, all our screens already have a parent container so this may be problematic. Also, I've personally moved slowly further and further away from native base components as they often "black box" functionality and can act buggy, I don't think you need to feel the need to use them if you don't need/want to.

src/components/feedback/styles/index.ts
```diff
@@ -0,0 +1,85 @@
...
+		color: colors.white,
+	},
+
+	submitTextDisabled: {
+		fontSize: 16,
+		color: colors.white,
+	},
+});
+
+export default styles;
```
bensinjin: This is nice! I'd actually like to start adopting the `index` approach for our components too but we haven't gotten there yet...

src/components/services/service_detail_component.tsx
```diff
@@ -51,14 +51,18 @@ export interface ServiceDetailActions {
 type Props = ServiceDetailProps & ServiceDetailActions & RouterProps;
 
 export const ServiceDetailComponent = (props: Props): JSX.Element => {
-    const [feedbackEnabled, setFeedbackEnabled]: readonly [boolean, Dispatch<SetStateAction<boolean>>] = useState(false);
-    const [feedback, setFeedback]: readonly [Feedback, Dispatch<SetStateAction<Feedback>>] =
-        useState(getDefaultFeedbackValues());
+    const [feedback, setFeedback]: readonly[Feedback, Dispatch<SetStateAction<Feedback>>]
+        = useState<Feedback>(getDefaultFeedbackValues());
```
bensinjin: Are you using Prettier by any chance?

src/hooks/use_toggle_state.ts
```diff
@@ -0,0 +1,34 @@
+import { Dispatch, SetStateAction, useCallback, useState } from 'react';
+
+export interface ToggleStateApi {
+    readonly state: boolean;
+    readonly on: () => void;
+    readonly off: () => void;
+    readonly toggle: () => void;
+}
+
+export const useToggleState = (initialValue?: boolean): ToggleStateApi => {
```
bensinjin: This feels like you see another use case for this hook it's conceivable but probably too early to know.

niko-pg: I just thought maybe it's good to keep all hooks that can be used generally into one directory. I will move it back if you think it's too early

src/components/feedback/feedback_modal_container.tsx
niko-pg: One of the good advantages of hooks is that the state management can be separated from the component. My common method is:
- Separate logic from JSX
- Try to reuse and compose hooks as much as possible. With hooks, it's more convenient to share state
- Hooks can have their own lifecycle/cleanups through `useEffect`. (So components wouldn't have to manage it)
- Hooks can be a good way to organize states. Say for example, with class component, everything is in one place:
```
this.state = {
   modalVisible: false,
   formInput: '',
   tableData,
}

setFormInput = () => {
   ...
}

showModal = () =>  {

}

hideModal = () => {
  ....
}

fetchTableData = () => {
    ...
}
```
With hooks, you can abstract this out to three different hooks, with each hook taking care of their own indivisible state: `useTableData`, `useForm` and `useModalState`

src/components/feedback/feedback_modal_container.tsx
```diff
@@ -1,108 +1,165 @@
...
 import React, { Dispatch, SetStateAction, useCallback, useState } from 'react';
 
-import { useToggleState, ToggleStateApi } from '../../hooks/use_toggle_state';
-
 import { FeedbackOptionsModalComponent } from './feedback_options_modal_component';
 import { FeedbackOtherRemoveServiceModal } from './feedback_other_remove_service_component';
 
+const SUGGESTION_CONTENT = {
+    OTHER: {
+        header: t`Other`,
```
tomy-pg: It was very useful to see you use the `t` macro here. I think we can clean-up some of our lingui work following your approach here.

src/components/feedback/styles/index.ts
rasmus-storjohann-PG: I do like to see components in files whose names match the components, makes it a lot easier to navigate and get a sense of where things are in a codebase. The filename index.ts tells me very little about what is in that file. That said, there are a few index.ts files in the project currently so I am not opposed to it generally.

rasmus-storjohann-PG:
No need to fix this, but I do prefer rebase onto `feedback_longlived_feature_branch` rather than merging it in, makes for tidier PRs.

src/components/feedback/feedback_other_remove_service_component.tsx
```diff
@@ -0,0 +1,133 @@
...
+                <HeaderComponent headerLabel={props.headerLabel} onClose={props.onClose} />
+                <ContentComponent inputLabel={props.inputLabel} placeholder={props.placeholder} />
+                <FooterComponent disabled={true} onSubmit={(): undefined => undefined} />
```
tomy-pg: Interesting to see you that you made both of these screens into a modal that has a header and footer. We had planned to refactor how our headers work to simplify things and to also help make the back button on the header work similar to how you have it here on our other feedback screens. I think when the back button is clicked, UX wants this screen to be shown to the feedback modal to be shown to the user: see [mock](https://app.zeplin.io/project/5bc7bfa42e72260acb03aa86/screen/5e0fe741e677e717f5c9cd2d) and with your implementation this can be done. 

However your implementation may make things difficult when you try to implement the third acceptance criteria in showing this [screen](https://app.zeplin.io/project/5bc7bfa42e72260acb03aa86/screen/5e0fe74134bdd11c327cb29f) when a user clicks the submit button as it's another modal. (I believe this is something you ran into the other night)

src/components/feedback/feedback_modal_container.tsx
rasmus-storjohann-PG: +1

src/components/feedback/feedback_modal_container.tsx
bensinjin: Nope, just curious, thx

src/components/feedback/feedback_other_remove_service_component.tsx
bensinjin: Np if it looks fine across platforms

src/components/feedback/feedback_other_remove_service_component.tsx
rasmus-storjohann-PG: @niko-pg I think you need to confirm that this will work on both platforms or else change the two feedback screens from modals to views.

src/components/feedback/feedback_modal_container.tsx
bensinjin: I'm not sure this works without the `i18n` object - has it been tested? Usually you need to use the HOC to inject the object and then you can used the `t` macro, something like:
```
        <I18n>
          {({ i18n }) => (
            <Button onPress={markAsRead} title={i18n._(t`mark messages as read`)} />
          )}
        </I18n>
```

src/components/feedback/feedback_other_remove_service_component.tsx
niko-pg: Confirmed that this works on both platforms per my testing

src/components/feedback/feedback_modal_container.tsx
niko-pg: These strings are passed down as props and then once when it reaches the component, that's where I use the `i18n` object:

https://github.com/pg-irc/pathways-frontend/pull/1063/files#diff-1ec3e05bd9eb776929d7d70da01458fbR70

bensinjin: Ah yes.

src/components/services/service_detail_component.tsx
```diff
@@ -116,11 +114,10 @@ export const ServiceDetailComponent = (props: Props): JSX.Element => {
                     {...props}
                 />
                 <DividerComponent />
-                <FeedbackButton isVisible={!feedbackEnabled} onPress={(): void => setShowFeedbackOptionsModal(true)} />
-                <FeedbackOptionsModalComponent
-                    isVisible={showFeedbackOptionsModal}
-                    setIsVisible={setShowFeedbackOptionsModal}
+                <FeedbackModalContainer
+                    feedbackEnabled={feedbackEnabled}
```
bensinjin: This boolean actually refers to only the feedback toggle for the individual fields I'm also not sure why `onFeedbackButtonPress` is being passed into this component as it's for toggling the field feedback. We should be adding additional properties to the feedback state in the `ServiceDetailComponent` for the "removeThisServiceFeedback" and the "otherFeedback" then, we should allow the modal forms to update it through a callback.

src/components/feedback/feedback_other_remove_service_component.tsx
```diff
@@ -0,0 +1,213 @@
...
+const HeaderComponent = (props: HeaderComponentProps): JSX.Element => {
+    const goBackShowModal = useServiceDetailRoute(
+        // TODO useQuery only supports string Records for now
+        { optionsModalVisible: 'false' },
+    );
+
+    const goBackHideModal = useServiceDetailRoute(
+        // TODO useQuery only supports string Records for now
+        { optionsModalVisible: 'true' },
+    );
```
tomy-pg: I think the booleans within these two are backwards.

src/components/feedback/feedback_other_remove_service_component.tsx
```diff
@@ -0,0 +1,213 @@
...
+        label: t`What is your reason for removal?`,
+        placeholder: t`e.g. Service is permanently closed`,
+    },
+};
+
+const useServiceDetailRoute = (queryParam: Record<string, string>): () => void => {
+    const params = useParams<RouteParams>();
+    const history = useHistory();
+
+    return replaceRouteWithParameters(
```
bensinjin, line 77: I'm noticing after this is called and you land back on the service detail screen you have to press back twice to get back to the service listing screen because your history is now: service listing screen, service detail screen, service detail screen. It'd be really great if feedback didn't have to be its own screen, but maybe that's not possible...

niko-pg: Hey, sorry I haven't been responsive with this issue. My first attempt were modals but there was an underlying issue with showing more than one modal for iOS. Handling of the modal state became really subtle that we've decided to pull them out as screens. I'm gonna take a look at this PR again tonight and see if there's a better way we can do

rasmus-storjohann-PG: I'm going to merge this PR to unblock @bensinjin. @niko-pg if you want to make additional improvements you can put up a new PR against the same issue number.

Issue #1048, opened 2020-03-12
Find related services near arbitrary location


rasmus-storjohann-PG:
Open the app on some topic, click find relates services, observe that the list of results is sorted by distance from the device location. Desired is to be able to enter the name of a location such as "Surrey" in order to get results sorted by the distance to the named location, as consistent as possible with the results from the Algolia search.

### Acceptance Criteria 
- [ ]  There is a field on the service listing page that allows the user to enter a location
- [ ] If the location is specified, the app hits geocoder to get the lat/long for the location
- [ ] If the field is empty, the device location is used [to be confirmed, see mocks when they're ready]
- [ ] The location from geocoder or from the device is submitted in the request to the server

Note that we are in the process of doing a fair amount of refactoring around the search. This will affect the code for calling the geocoder endpoint. The plan is to merge your PR first. So use existing code to access geocoder, but don't refactor it or make a lot of changes outside the component.

bensinjin:
Moving @cherrielam 's ticket for this here:

We would like to add the location search function to the topic-services page. This is because: As a user, I would like to adjust the service results on a topic in order to search in the location that is of interest to me. This is particularly important for users who want to use AA pre-arrival.

Mock is [here](https://app.zeplin.io/project/5bc7bfa42e72260acb03aa86/dashboard?seid=5e6ab3e4e90d37189ac6645e)

### Acceptance Criteria 
- [ ]  Topic>Services List has location search function as reflected in mocks, and service list updates according to the input location.

### Steps for QA: 
1.

rasmus-storjohann-PG:
QA: Open the service search for a topic, observe many services are listed, focus on search input field, enter "Kelowna", hit Search, observe fewer services are listed, focus on search input field, clear the field by hitting X, hit search, expected is that many services are listed, observe that the listing still shows results for Kelowna.

bensinjin:
Thanks, the fix has been pushed. I've noticed another potential issue. Because this component only triggers the loading screen when we actually start searching for services there is a delay in displaying it while the location is being fetched. There are two potential fixes:

A) Easy approach: leave the search expanded, which displays the buttons as inactive while the location lookup completes, collapse it after to indicate we've found the location

B) Longer/Refactor approach: Move the `isFetchingLatLng` state up into the parent component so we can display the loading screen when child component performs the lookup

Let me know your thoughts, thx!

cherrielam:
[...]

PR #1064, opened 2020-03-18
Issue #1048 Services manual location

bensinjin:
Please review these points:

- The new services location form sets the existing, but not in use, manual user location store value along with a new label property. This allows us to not have to  change the default behaviours of the services saga which will use the manual user location if it exists.
- Having a fixed header, outside the flatlist, now allows us to simplify our "refresh" logic, we don't need to set a timestamp in order to trigger the refresh
- If location lookup fails from the location form currently it does so silently and manual location is not updated, if this is not sufficient we could add lookup error message to the form 
- I've intentionally sparingly used outside code from the existing search files due to the large refactor that is currently being done for search.

### DIFF
--- a/src/components/services/service_list_component.tsx
+++ b/src/components/services/service_list_component.tsx
@@ -47,16 +47,11 @@ export interface ServicesUpdater {
 
 type Props = ServiceListProps & ServiceListActions & ServicesUpdater & RouterProps;
 
-type Timestamp = number;
-type TimestampSetter = Dispatch<SetStateAction<Timestamp>>;
-
 export const ServiceListComponent = (props: Props): JSX.Element => {
-    const [lastScreenRefresh, setLastScreenRefresh]: readonly [Timestamp, TimestampSetter] = useState(Date.now());
-    useEffect(() => { refreshServices(props); }, [lastScreenRefresh, props.manualUserLocation]);
-    const refreshScreen = (): void => setLastScreenRefresh(Date.now());
+    useEffect(refreshServices(props), [props.manualUserLocation]);
 
     if (isSelectorErrorServicesForTopic(props.topicServicesOrError)) {
-        return renderErrorComponent(props, refreshScreen);
+        return renderErrorComponent(props, refreshServices(props));
     }
 
     if (isLoadingServices(props.topicServicesOrError) || isInitialEmptyTopicServices(props.topicServicesOrError)) {
@@ -64,7 +59,7 @@ export const ServiceListComponent = (props: Props): JSX.Element => {
     }
 
     if (isValidEmptyTopicServices(props.topicServicesOrError)) {
-        return renderEmptyComponent(props, refreshScreen);
+        return renderEmptyComponent(props, refreshServices(props));
     }
 
     return (
@@ -80,11 +75,13 @@ export const ServiceListComponent = (props: Props): JSX.Element => {
     );
 };
 
-const refreshServices = (props: Props): void => {
-    if (servicesFetchRequired(props.topicServicesOrError) || hasManualUserLocation(props.manualUserLocation)) {
-        props.dispatchServicesRequest();
+const refreshServices = (props: Props): () => void => (
+    (): void => {
+        if (servicesFetchRequired(props.topicServicesOrError) || hasManualUserLocation(props.manualUserLocation)) {
+            props.dispatchServicesRequest();
+        }
     }
-};
+);
 
 const servicesFetchRequired = (topicServicesOrError: SelectorTopicServices): boolean => (
     topicServicesOrError.type === constants.ERROR_SERVICES_FOR_TOPIC ||
--- a/src/components/services/service_list_component.tsx
+++ b/src/components/services/service_list_component.tsx
@@ -24,6 +24,7 @@ import { MessageComponent } from '../partial_localization/message_component';
 import { HidePartialLocalizationMessageAction } from '../../stores/user_profile';
 import { SetManualUserLocationAction } from '../../stores/manual_user_location';
 import { ServiceListLocationSearchComponent } from './service_list_location_search_component';
+import { SearchListSeparator } from '../search/separators';
 
 export interface ServiceListProps {
     readonly topic: Topic;
@@ -63,15 +64,21 @@ export const ServiceListComponent = (props: Props): JSX.Element => {
     }
 
     return (
-        <View>
-            {renderHeader(props)}
+        <ServiceListWrapper {...props}>
             <FlatList
-                style={{ backgroundColor: colors.lightGrey }}
+                style={{ backgroundColor: colors.white }}
                 data={getServicesIfValid(props.topicServicesOrError)}
                 keyExtractor={(service: HumanServiceData): string => service.id}
                 renderItem={renderServiceListItem(props)}
+                ItemSeparatorComponent={SearchListSeparator}
+                ListHeaderComponent={
+                    <MessageComponent
+                        isVisible={props.showPartialLocalizationMessage}
+                        hidePartialLocalizationMessage={props.hidePartialLocalizationMessage}
+                    />
+                }
             />
-        </View>
+        </ServiceListWrapper>
     );
 };
 
@@ -90,7 +97,16 @@ const servicesFetchRequired = (topicServicesOrError: SelectorTopicServices): boo
 );
 
 const hasManualUserLocation = (manualUserLocation: UserLocation): boolean => (
-    !!(manualUserLocation.label && manualUserLocation.latLong.lat && manualUserLocation.latLong.lng)
+    !!(manualUserLocation.label && manualUserLocation.latLong)
+);
+
+type ServiceListWrapperProps = Props & { readonly children: JSX.Element };
+
+const ServiceListWrapper = (props: ServiceListWrapperProps): JSX.Element => (
+    <View style={{ flex: 1 }}>
+        {renderHeader(props)}
+        {props.children}
+    </View>
 );
 
 const renderErrorComponent = (props: Props, refreshScreen: () => void): JSX.Element => {
@@ -98,41 +114,37 @@ const renderErrorComponent = (props: Props, refreshScreen: () => void): JSX.Elem
     const sentryMessage = getSentryMessageForError(errorType, constants.SENTRY_SERVICES_LISTING_ERROR_CONTEXT);
     Sentry.captureMessage(sentryMessage);
     return (
-        <ErrorScreenSwitcherComponent
-            refreshScreen={refreshScreen}
-            errorType={errorType}
-            header={renderHeader(props)}
-        />
+        <ServiceListWrapper {...props}>
+            <ErrorScreenSwitcherComponent
+                refreshScreen={refreshScreen}
+                errorType={errorType}
+            />
+        </ServiceListWrapper>
     );
 };
 
 const renderLoadingComponent = (props: Props): JSX.Element => (
-    <LoadingServiceListComponent
-        header={renderHeader(props)}
-    />
+    <ServiceListWrapper {...props}>
+        <LoadingServiceListComponent />
+    </ServiceListWrapper>
 );
 
 const renderEmptyComponent = (props: Props, refreshScreen: () => void): JSX.Element => (
-    <EmptyServiceListComponent
-        title={<Trans>No services to show</Trans>}
-        imageSource={emptyTopicServicesList}
-        refreshScreen={refreshScreen}
-        header={renderHeader(props)}
-    />
+    <ServiceListWrapper {...props}>
+        <EmptyServiceListComponent
+            title={<Trans>No services to show</Trans>}
+            imageSource={emptyTopicServicesList}
+            refreshScreen={refreshScreen}
+        />
+    </ServiceListWrapper>
 );
 
 const renderHeader = (props: Props): JSX.Element => (
-    <View>
-        <ServiceListHeaderComponent
-            title={props.topic.title}
-            manualUserLocation={props.manualUserLocation}
-            setManualUserLocation={props.setManualUserLocation}
-        />
-        <MessageComponent
-            isVisible={props.showPartialLocalizationMessage}
-            hidePartialLocalizationMessage={props.hidePartialLocalizationMessage}
-        />
-    </View>
+    <ServiceListHeaderComponent
+        title={props.topic.title}
+        manualUserLocation={props.manualUserLocation}
+        setManualUserLocation={props.setManualUserLocation}
+    />
 );
 
 const determineErrorType = (topicServicesOrError: SelectorTopicServices): Errors => {
@@ -181,9 +193,9 @@ export const ServiceListHeaderComponent = (props: ServiceListHeaderComponentProp
     <View
         style={{
             backgroundColor: colors.teal,
-            paddingTop: 25,
+            paddingTop: 10,
             paddingHorizontal: 10,
-            paddingBottom: 10,
+            paddingBottom: 4,
         }}
     >
         <Text style={[textStyles.headlineH2StyleWhiteLeft, { paddingHorizontal: 10}]}>
--- a/src/components/services/service_list_component.tsx
+++ b/src/components/services/service_list_component.tsx
@@ -84,20 +84,8 @@ export const ServiceListComponent = (props: Props): JSX.Element => {
 
 const refreshServices = (props: Props): () => void => (
     (): void => {
-        if (servicesFetchRequired(props.topicServicesOrError) || hasManualUserLocation(props.manualUserLocation)) {
-            props.dispatchServicesRequest();
-        }
+        props.dispatchServicesRequest();
     }
-);
-
-const servicesFetchRequired = (topicServicesOrError: SelectorTopicServices): boolean => (
-    topicServicesOrError.type === constants.ERROR_SERVICES_FOR_TOPIC ||
-    topicServicesOrError.type === constants.VALID_SERVICES_FOR_TOPIC && topicServicesOrError.isExpired ||
-    topicServicesOrError.type === constants.INITIAL_EMPTY_SERVICES_FOR_TOPIC
-);
-
-const hasManualUserLocation = (manualUserLocation: UserLocation): boolean => (
-    !!(manualUserLocation.label && manualUserLocation.latLong)
 );
 
 type ServiceListWrapperProps = Props & { readonly children: JSX.Element };
--- a/src/components/services/service_list_component.tsx
+++ b/src/components/services/service_list_component.tsx
@@ -14,7 +14,7 @@ import { isSelectorErrorServicesForTopic } from '../../selectors/services/is_sel
 import * as constants from '../../application/constants';
 import { ErrorScreenSwitcherComponent } from '../error_screens/ErrorScreenSwitcherComponent';
 import { Errors } from '../../validation/errors/types';
-import { LatLong } from '../../validation/latlong/types';
+import { UserLocation } from '../../validation/latlong/types';
 import { getSentryMessageForError } from '../../validation/errors/sentry_messages';
 import { Routes, RouterProps, goToRouteWithParameter } from '../../application/routing';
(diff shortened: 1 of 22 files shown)


src/components/services/service_list_location_search_component.tsx
```diff
@@ -0,0 +1,263 @@
...
+    const latLong = await fetchFn();
+    setIsFetchingLatLng(false);
+    if (!latLong) return;
+    setManualUserLocation({
+        label: locationLabel,
+        latLong,
+    });
+};
+
+// TODO: These both may be handy elsewhere and could be moved into a helper file.
```
bensinjin: Re. code reuse: I wanted to use: https://github.com/pg-irc/pathways-frontend/blob/develop/src/components/search/api/use_fetch_lat_long_from_location.ts#L13 but this hook actually fetches the lat long whenever the location string changes, which I do not want. I exported what I needed from that file but the functionality I have here doesn't exist in its current form.

src/components/services/service_list_location_search_component.tsx
```diff
@@ -202,28 +202,39 @@ const getSearchOnPress = (
...
-    const latLong = await fetchLatLngFromDevice();
+    const latLong = await fetchFn();
     setIsFetchingLatLng(false);
-
     if (!latLong) return;
     setManualUserLocation({
-        label: MY_LOCATION,
+        label: locationLabel,
         latLong,
     });
```
rasmus-storjohann-PG: Use `try...finally` to call `setIsFetchingLatLng(false);`

bensinjin:
> Nice work, happy to see quite a bit of simplification in this PR. There are some bugs though, the My location button doesn't really do anything, it clicks but it doesn't update the string in the location input field. If it would help, you could rebase your PR onto the search refactor PR so you can see your code in that context. The big refactor is very close to done.

@rasmus-storjohann-PG your "My Location" look is likely failing, nothing occurs when this happens (we don't set the label)

src/components/services/service_list_component.tsx
```diff
@@ -80,11 +75,13 @@ export const ServiceListComponent = (props: Props): JSX.Element => {
...
-    if (servicesFetchRequired(props.topicServicesOrError) || hasManualUserLocation(props.manualUserLocation)) {
-        props.dispatchServicesRequest();
+const refreshServices = (props: Props): () => void => (
+    (): void => {
+        if (servicesFetchRequired(props.topicServicesOrError) || hasManualUserLocation(props.manualUserLocation)) {
+            props.dispatchServicesRequest();
+        }
     }
-};
+);
```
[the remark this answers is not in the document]
bensinjin: It's just leveraging the closure that needs access to props so it's a function that returns a function like: `(props) => () => { // use props here... }`. Our overly strict tslint `typedef` rule makes things like this so hard to read and makes it look wayyy more complicated than it is. Personally I think it's time we reviewed the `typedef` tslint rule unless we feel it's actually preventing bugs.

src/stores/manual_user_location.ts
```diff
@@ -33,7 +36,7 @@ export const reducer = (store: ManualUserLocationStore = buildDefaultStore(), ac
         return { ...store, userLocation: action.payload.userLocation };
     }
     if (action.type === constants.CLEAR_MANUAL_USER_LOCATION) {
-        return { ...store, userLocation: undefined };
+        return { ...store, ...buildDefaultStore() };
```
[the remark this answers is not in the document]
bensinjin: You're right, fixed.

src/validation/latlong/types.ts
```diff
@@ -2,3 +2,8 @@ export interface LatLong {
     readonly lat: number;
     readonly lng: number;
 }
+
+export interface UserLocation {
+    readonly label: string;
```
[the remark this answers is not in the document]
bensinjin: Correct, it's the location "label" or "string" or "text" or really whatever you can think of :P

src/components/services/service_list_location_search_component.tsx
bensinjin: So the current strategy is not do anything if a lat lng fetch fails so I don't think we actually want a finally clause and I think it does make sense to just keep the last successful lookup. Using the `try..catch` approach here actually doesn't really work either because the scope of the `latLong` variable would be limited to the first `try` block and then inaccessible from outside of it (this is due to how `const` and `let` work).

src/components/services/service_list_component.tsx
```diff
@@ -174,47 +173,25 @@ const renderServiceListItem = (props: Props): ({ item }: ServiceItemInfo) => JSX
...
-    return (
-        <View style={{
+export const ServiceListHeaderComponent = (props: ServiceListHeaderComponentProps): JSX.Element => (
+    <View
+        style={{
             backgroundColor: colors.teal,
-            marginHorizontal: -10,
-            marginTop: -10,
-            padding: 20,
+            paddingTop: 25,
```
niko-pg: Not really something from your PR, but in general, I'm proactively suggesting that we should start abstracting the styles out of the component when necessary. Lots of benefits we can reap with this approach which includes less component clutters, better organization and styles would only be modified in one place.

src/components/services/service_list_location_search_component.tsx
```diff
@@ -0,0 +1,263 @@
...
+import { getDeviceLocation, DeviceLocation } from '../../async/location';
+import { isLocationFetchTimeoutError, isNoLocationPermissionError } from '../../validation/errors/is_error';
+
+interface Props {
+    readonly manualUserLocation: UserLocation;
+    readonly setManualUserLocation: (userLocation: UserLocation) => SetManualUserLocationAction;
+}
+
+export const ServiceListLocationSearchComponent = (props: Props): JSX.Element => {
+    const [locationInputValue, setLocationInputValue]: readonly [string, (s: string) => void] = useState(props.manualUserLocation.label);
```
niko-pg: Should insert `<boolean>` here?

src/components/services/service_list_location_search_component.tsx
```diff
@@ -0,0 +1,263 @@
...
+                        props.setManualUserLocation,
+                    )}
+                    isFetchingLatLng={props.isFetchingLatLng}
+                />
+            </View>
+        </View>
+    );
+};
+
+interface SearchInputProps {
```
niko-pg: NIT: My personal preference, would be to reuse types as much as possible. Since you declared `SearchInputProps`, `ExpandedSearchProps` could actually inherit from this:

```
interface SearchInputProps {
    readonly locationInputValue: string;
    readonly setLocationInputValue: (s: string) => void;
}

interface ExpandedSearchProps extends SearchInputProps {
    readonly isFetchingLatLng: boolean;
    readonly setIsFetchingLatLng: (b: boolean) => void;
    readonly setManualUserLocation: Props['setManualUserLocation'];
}
```

Same thing can also be applied to the other types here. I'll leave it to you if you want to apply this or not

src/components/services/service_list_location_search_component.tsx
```diff
@@ -0,0 +1,263 @@
...
+                    />
+                )
+            }
+        </I18n>
+    </View>
+);
+
+const LocationIcon = (): JSX.Element => (
+    <Icon
+        name={'location-on'}
```
niko-pg: I dont think the brackets are needed here

src/components/services/service_list_location_search_component.tsx
```diff
@@ -0,0 +1,263 @@
...
+        type='MaterialIcons'
+        style={{ color: colors.teal, fontSize: values.smallIconSize, padding: 10 }}
+    />
+);
+
+const CloseIcon = (): JSX.Element => (
+    <Icon
+        name={'window-close'}
```
niko-pg: Same here

src/components/services/service_list_location_search_component.tsx
```diff
@@ -0,0 +1,263 @@
...
+            <Trans>My Location</Trans>
+        </Text>
+    </TouchableOpacity>
+);
+
+const getSearchOnPress = (
+    setIsFetchingLatLng: (isFetchingLatLng: boolean) => void,
+    locationInputValue: string,
+    setManualUserLocation: Props['setManualUserLocation'],
+): () => Promise<void> => async (): Promise<void> => {
```
niko-pg: Is the `async` needed here?

src/components/services/service_list_location_search_component.tsx
```diff
@@ -0,0 +1,263 @@
...
+    if (locationInputValue === MY_LOCATION) return;
```
niko-pg: I just noticed that the return type might not be needed since I don't see anything being returned? 
nor I don't see any `await` inside.

src/components/services/service_list_location_search_component.tsx
```diff
@@ -0,0 +1,263 @@
...
+        fetchFn,
+        setManualUserLocation,
+        locationInputValue,
+    );
+};
+
+const getLocateOnPress = (
+    setIsFetchingLatLng: (isFetchingLatLng: boolean) => void,
+    setManualUserLocation: Props['setManualUserLocation'],
+): () => Promise<void> => async (): Promise<void> => {
```
niko-pg: Same comment as above. Is the type here really needed? I don't see anything being returned

src/components/services/service_list_location_search_component.tsx
niko-pg: Would declaring `latLong` outside the `try...catch` work though? And also if the lat lng fetch fails, the code will just throw an error and not having a chance to call `setIsFetchingLatLng(false);`

Not sure of `fetchFn`'s implementation does but it seems like we're not guarding this whole thing for any potential crashes.

I agree with Rasmus, probably still best to wrap this thing with `try catch finally`.

src/components/services/service_list_location_search_component.tsx
```diff
@@ -0,0 +1,263 @@
...
+        .then(JSON.parse)
+        .then(toGeoCoderLatLong)
+        .catch((error: string) => {
+            console.log(error);
+            return undefined;
+        })
+);
+
+const fetchLatLngFromDevice = async (): Promise<LatLong | undefined> => {
+    const deviceLocationResponse: DeviceLocation = await getDeviceLocation();
```
niko-pg: Is `getDeviceLocation` handling the error internally? If not, probably best to wrap with `try catch`

src/sagas/services.ts
```diff
@@ -23,7 +23,7 @@ export function* updateServicesForTopic(action: actions.BuildServicesRequestActi
             return yield put(actions.buildServicesError(topicId, Errors.Offline));
         }
 
-        const deviceLocationResponse: DeviceLocation = yield call(getDeviceLocation, action.payload.manualUserLocation);
+        const deviceLocationResponse: DeviceLocation = yield call(getDeviceLocation, action.payload.manualUserLocation.latLong);
```
niko-pg: Same comment here. Do we know that `getDeviceLocation` is safe that it's not gonna throw any error?

src/components/services/service_list_location_search_component.tsx
```diff
@@ -0,0 +1,263 @@
...
+            {
+                backgroundColor: colors.white,
+                opacity: props.isFetchingLatLng ? values.disabledOpacity : 1,
+            },
+        ]}
+        onPress={props.onPress}
+        disabled={props.isFetchingLatLng}
+    >
+        <Icon
+            type={'MaterialIcons'} name={'my-location'}
```
niko-pg: I dont think the brackets are needed.

src/components/services/service_list_component.tsx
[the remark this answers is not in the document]
bensinjin: There is no third layer here, this is a function that closes over props and returns a function. I think what you're referring to is the type declarations ie: `() => void` which muddies the code. Both `useEffect` and `onPress` use `refreshServices` so this is just code reuse, you're right the closure closes over the current render's props but then the next render all these functions will be updated, can you demonstrate the out of sync problem, that'd be interesting for sure.

Either way, lemme know how you guys want this to look I'm totally cool updating it.

src/components/services/service_list_component.tsx
bensinjin: You mean moving them into stylesheets in the file?

src/components/services/service_list_location_search_component.tsx
bensinjin: Yeah, I hear ya, from personal experience however inheritance in TS interfaces is a bit of nightmare when trying to read the code later as you're constantly jumping around trying to figure out the actual shape of the type. Also just because two TS interfaces share members does not mean one `IS A` nother. I also structure it like this so we can easily pull these components out into their own files if we want to later.

src/components/services/service_list_location_search_component.tsx
bensinjin: They're not

src/components/services/service_list_location_search_component.tsx
bensinjin: Nope!  Refactored and missed, I'll tidy.

src/components/services/service_list_location_search_component.tsx
bensinjin: same as above

src/components/services/service_list_component.tsx
niko-pg: Right. But this shouldn't be addressed in your review. Just throwing this suggestion out there.

src/components/services/service_list_location_search_component.tsx
niko-pg: Fair enough. I'm basing this comment on a pattern that I typically use whenever I declare props, I always take out the `default` from `actual` props. In this case, I guess yeah it's different

```
interface DefaultProps {
}

interface ActualProps extends DefaultProps {
}
```

Anyhow, good point. The inheritance part doesn't make sense here so I think we're good.

src/components/services/service_list_component.tsx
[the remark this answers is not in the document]
niko-pg: Oh right! I thought there were three lol. My whole comment is based from the knowledge that there are three layers, so you can ignore.

In general, having many closures could lead to unnoticed stale values. So it's something that should be avoided for most of the time. This blog post does a good job of explaining this: https://dmitripavlutin.com/react-hooks-stale-closures/

In this particular case, I think the effect would still work properly. Though, readability is still a concern.  I'm fine with going as is, I don't want to hold this review if this is only gonna lead to a bigger refactor.

src/components/services/service_list_location_search_component.tsx
bensinjin: Yep! If you used `var`, this just kinda goes against our conventions, in this specific instance I know on failure the `fetchFn` returns `undefined` I do agree that this is not as defensive as it could be. I've added a `catch()` call that now does this explicitly from the callsite.

src/components/services/service_list_location_search_component.tsx
bensinjin: Fixed!

src/components/services/service_list_location_search_component.tsx
bensinjin: Fixed!

src/components/services/service_list_location_search_component.tsx
bensinjin: Fixed!

src/components/services/service_list_location_search_component.tsx
bensinjin: It does and returns errors of specific types for us so we can check those (see next line).

src/components/services/service_list_location_search_component.tsx
bensinjin: This is more of a styling thing I think we've just gotten used to doing you're right though it's not necessary.

src/components/services/service_list_component.tsx
bensinjin: Yeah I think early on we decided we were ok with inline styles in places where they were only used once otherwise put them in our `application/styles.ts` Stylesheet. This was primarily because naming styles and jumping around files took longer, the gains however is with stylesheets the runtime can cache them which may be worth it. Convention shows Stylesheets are likely more popular I just don't want to get into debates over style names, if we can agree on that I'm down ;)

bensinjin:
I think I've addressed the outstanding issues around this but there is still the question of how the "My Location" button should work. I've assumed that pressing it will try and get your device location and if successful the search will be performed otherwise nothing happens. Is this right? Could people test it and see how it feels to them?

bensinjin:
We've agreed "My Location" should follow how it behaves on the main services search.

rasmus-storjohann-PG:
When there is a location in the search bar, the search bar is not clickable, only the X is. The whole text bar should be clickable and give focus to the input line. Click on the X clear the field and gives focus, clicking on the rest of the input field just gives focus without clearing it

src/components/services/service_list_location_search_component.tsx
rasmus-storjohann-PG: You can avoid the `var` by pulling that logic out into a helper function so each time you would assign, you do a return instead. (@niko-pg we try to keep all our state read-only)

bensinjin:
@rasmus-storjohann-PG fix pushed, the Travis failures look unrelated.

src/sagas/services.ts
bensinjin: (See above)

bensinjin:
At this point we're waiting on some feedback from @cherrielam as to how the functionality should work.

cherrielam:
@bensinjin Looks good, a few notes:

**Behaviour and styling are different between the _Location search on topic-services_ and _Search feature_. Can we make these consistent? Specifically:** 

1. Clicking "My Location" on _Topic-services_ runs the search immediately. Clicking "My Location" on Search just fills the input field with "My Location" and the user still has to click the Search button to run the search. My preference is that both searches function as the latter ("My location" only fills the form field) to keep it consistent between the two. Note: There is a bug currently on _Topic-services_ where once "My Location" is entered, if the user clicks the field (not the x), they can bring it into focus and edit the text that says "My Location" but clicking "Search" button does not run the search anymore.

2. "My Location" buttons are different height between _Topic-services_ and _Search_. Let's go with the height that is on the _Search_ (where the "My location" button is the same height as the "Search" button). It looks a little cleaner.

IMG_8213

IMG_8215

3. The "Info only in English" card scrolls with the services content in _Search_. It does not scroll in _topic-services_. Let's make it scroll with the services content (as in _Search_) so it doesn't block so much of the user's view. Also because in the one that is static (on _topic-services_) it looks buggy that there is the beige horizontal bar underneath the card.

IMG_8214

IMG_8212

**And a few styling requests on the _topic-services_**

4. Can we reduce the space above the topic title on the topic-services page to reduce the real estate that the teal part takes up?

5. Bug: The whole page can scroll when it is loading (including the header with the search bar), but once it stops loading only the services scroll. 

RPReplay_Final1585090942


6. Corner-case bug: If the location is very long, it just runs onto the right side... Is it possible to hide the overflow?
IMG_8218

src/components/services/service_list_location_search_component.tsx
rasmus-storjohann-PG: I'm with @bensinjin on this conversation, inheritance doesn't tend to help readability. It has it's uses, but I tend to avoid it.

src/components/services/service_list_location_search_component.tsx
rasmus-storjohann-PG: no strong preference

bensinjin:
Thanks @cherrielam! I think I've covered everything outlined, can I get one more review?

bensinjin:
Think I addressed everything here, can I get one more review @cherrielam ?

bensinjin:
Clear button added when you enter text as per @rasmus-storjohann-PG 's feedback.

cherrielam:
Thanks @bensinjin ! Looks great, can we increase the padding below the topic title in the topic-services page? The same padding as before is good, I just wanted to reduce the padding above the topic title last time.

Also is it possible to make the "My Location" button disappear when it's already set to "My Location"? In the same behaviour that the _Search_ currently has. No worries if not.

bensinjin:
Fixed!

PR #1074, opened 2020-03-23
Feature/issue 1024 refactor header component

tomy-pg:


src/components/main/main_component.tsx
```diff
@@ -14,122 +13,79 @@ import { AppModalsComponent } from './app_modals_component';
...
-                </Drawer>
-            </Root>
-        );
-    }
-
-    closeDrawer(): void {
-        this.setState({ isHeaderMenuShowing: false });
-    }
+        Notifications.addListener(notificationListener(props.history));
+    }, []);
```
bensinjin: `props.local` should be a dependency here.

src/components/services/service_list_component.tsx
```diff
@@ -62,18 +65,23 @@ export const ServiceListComponent = (props: Props): JSX.Element => {
...
-            refreshing={isLoadingServices(props.topicServicesOrError)}
-            onRefresh={(): void => refreshServices(props)}
-            refreshControl={renderRefreshControl(isLoadingServices(props.topicServicesOrError))}
-            data={getServicesIfValid(props.topicServicesOrError)}
-            keyExtractor={(service: HumanServiceData): string => service.id}
-            renderItem={renderServiceListItem(props)}
-            ListHeaderComponent={renderHeader(props)}
-        />
+        <View style={{ flex: 1 }}>
+            <TwoButtonHeaderComponent
```
bensinjin: So what about the other instances like errors or empty? Do we still render the header in those cases?

tomy-pg: Yeah good catch, I definitely missed those cases.

tomy-pg:
@bensinjin Yes for every new screen we will have to define a Header.

src/components/services/service_list_component.tsx
```diff
@@ -62,18 +65,23 @@ export const ServiceListComponent = (props: Props): JSX.Element => {
...
+                {...props}
+                {...{ textColor: colors.white, backgroundColor: colors.teal }}
```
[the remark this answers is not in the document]
tomy-pg: Doing this gave me a few errors:
```
Left side of comma operator is unused and has no side effects.ts
unused expression, expected an assignment or function call (edited) 
Cannot find name ‘textColor’.
```

src/components/bookmarks/bookmarks_component.tsx
```diff
@@ -8,28 +8,51 @@ import { Trans, I18n } from '@lingui/react';
...
+
+type Props = BookmarksProps & BookmarkActions & TaskListActions & RouterProps ;
 
-export const BookmarksComponent: React.StatelessComponent<Props> = (props: Props): JSX.Element => {
+export const BookmarksComponent = (props: Props): JSX.Element => {
     return (
-        <Container style={{ backgroundColor: colors.lightGrey }}>
-            <HeaderComponent/>
+        <View style={{ flex: 1 }}>
+            <Header {...props} />
```
niko-pg: Is this being rendered properly? I think `Header` should be inside a `Container` if we're following Native base's suggested structure.

src/components/explore/explore_detail_component.tsx
```diff
@@ -57,3 +62,15 @@ const TaskListHeaderComponent = (props: Props): JSX.Element => (
...
+    const textColor = colors.black;
+    const backgroundColor = colors.lightGrey;
+    const leftButton = <BackButtonComponent history={props.history} textColor={textColor} />;
+    const rightButton =
+        <MenuButtonComponent
+            onPress={props.openHeaderMenu}
+            textColor={textColor}
+        />;
+    return renderHeader({ backgroundColor, leftButton, rightButtons: [rightButton] });
+};
```
niko-pg: This pattern seems to be repeated most of the time. Is there a way we can abstract this out as a common component and reuse it?

src/components/main/main_component.tsx
```diff
@@ -14,122 +13,83 @@ import { AppModalsComponent } from './app_modals_component';
...
         );
-    }
+    }, [props.location, props.locale]);
 
-    closeDrawer(): void {
-        this.setState({ isHeaderMenuShowing: false });
-    }
+    useEffect((): void => {
+        Notifications.addListener(notificationListener(props.history));
+    }, []);
```
niko-pg: NotifcationListener should be closed on cleanup.

```
useEffect(
  () => {
      const listener = Notifications.addListener(notificationListener(props.history));

      return () => listener();
  },
  [],
)
```

src/components/main/main_component.tsx
```diff
@@ -14,122 +13,83 @@ import { AppModalsComponent } from './app_modals_component';
...
-                    </Container>
-                </Drawer>
-            </Root>
+export const MainComponent = (props: Props): JSX.Element => {
+    useEffect((): void => {
+        props.history.listen((location: Location, _: Action): RouteChangedAction =>
+            props.sendAnalyticsData(location, props.locale),
         );
-    }
+    }, [props.location, props.locale]);
```
niko-pg: Hmmm I think this dependency `(props.location)` is not needed since this is already given by the callback. Actually the deps may not even be needed at all since we only want to initializer the listener on mount

Also I think this should unlisten during cleanup.

```
useEffect(
  () => {
     const unlisten = props.history.listen((location, action) => {
         console.log('You changed the page to: ' location.pathname)
      });

     return () => unlisten();
  },
  [],
)
```

src/components/recommended_topics/recommended_topics_component.tsx
```diff
@@ -18,29 +18,49 @@ import { RecommendedIconComponent } from './recommended_icon_component';
...
-};
+}
 
-type Props = RecommendedTopicsProps & TaskListActions & RouterProps & AnalyticsAction;
+type Props = RecommendedTopicsProps & RecommendedTopicsActions & TaskListActions & RouterProps;
 
 export const RecommendedTopicsComponent: React.StatelessComponent<Props> = (props: Props): JSX.Element => (
-    <TaskListComponent
+    <View style={{ flex: 1 }}>
+        <Header {...props} />
```
niko-pg: One thing I want to suggest in the codebase is that maybe we should minimize the use of prop spreading. As much as possible, we should be explicit so we know which props are being received by the component.

There is actually an airbnb eslint rule for this: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-props-no-spreading.md

src/components/recommended_topics/recommended_topics_component.tsx
```diff
@@ -18,29 +18,49 @@ import { RecommendedIconComponent } from './recommended_icon_component';
...
         emptyTaskListContent={<EmptyTopicListComponent message={<Trans>No topics to recommend</Trans>}/>}
         headerContent={<TaskListHeaderComponent {...props} />}
     />
+    </View>
 );
 
+const Header = (props: Props): JSX.Element => {
+    const textColor = colors.teal;
+    const backgroundColor = colors.white;
+    const leftButton = <HelpButtonComponent history={props.history} />;
```
niko-pg: NIT: there is a `useHistory` hook which you can call on any component which gives you access to the `history`object. This can avoid the tedious passing down of history prop.

src/components/topics/topic_detail_component.tsx
```diff
@@ -0,0 +1,109 @@
...
+                    paddingTop: 18,
+                },
+            ]}
+        >
+            <Trans>RELATED TOPICS</Trans>
+        </Text>
+    </View>
+);
+
+const onServicesTextPress = (props: Props): () => void => {
```
niko-pg: NIT: `onServicesTextPress` is actually not an accurate name for this because this function is actually not the callback but rather the callback creator. I usually go for `createOnServicesTextPress` name in this scenario.

src/components/main/main_component.tsx
tomy-pg: Unsubscribing and un-listening are really good suggestions. Fixed!

As well, I have removed having location and locale as dependencies. Including these two properties as dependencies introduced a performance issue which made navigating through the app really slow.

src/components/recommended_topics/recommended_topics_component.tsx
tomy-pg: I think so far we have gone by preference with this.

src/components/recommended_topics/recommended_topics_component.tsx
tomy-pg: I like. Perhaps we can add this to a new issue and address it all at once since this PR is already getting pretty large.

src/components/bookmarks/bookmarks_component.tsx
tomy-pg: I actually ended up removing the `Container` from this component. The components parent, `MainComponent` already surrounds its children with the `<Container/>`.

src/components/help/help_component.tsx
```diff
@@ -67,45 +69,52 @@ export interface HelpComponentActions {
...
+            {...props}
+            {...{ textColor: colors.teal, backgroundColor: colors.white }}
+        />
+        <Content padder style={applicationStyles.body}>
+            <View
+                style={{
+                    backgroundColor: colors.white,
+                    marginTop: -10,
+                    marginHorizontal: -10,
+                    padding: 10,
```
niko-pg: NIT: extra space

src/components/help/help_component.tsx
```diff
@@ -67,45 +69,52 @@ export interface HelpComponentActions {
...
+                    marginBottom: 10,
+                }}
+            >
+                <Text style={[textStyles.headlineH1StyleBlackLeft, { paddingHorizontal: values.backgroundTextPadding }]}>
+                    <Trans>{'Help & Support'}</Trans>
```
niko-pg: NIT: Unnecessary brackets? noticed because other texts don't have brackets. Not sure if we want to be consistent. I'll leave it to you if you want to change.

src/components/main/main_component.tsx
```diff
@@ -14,122 +13,90 @@ import { AppModalsComponent } from './app_modals_component';
...
+        const unlisten = props.history.listen((_: Location): void => {
+            return;
+        });
+        return (): void => unlisten();
+    }, []);
+
+    useEffect((): EffectCallback => {
+            const notificationsListener = Notifications.addListener(notificationListener(props.history));
+            // tslint:disable-next-line: no-unused-expression
+            notificationsListener;
```
niko-pg: Is this empty expression needed?

src/components/main/main_component.tsx
```diff
@@ -14,122 +13,90 @@ import { AppModalsComponent } from './app_modals_component';
...
-    openDisclaimerModal(): void {
-        this.setState({ isDisclaimerModalOpen: true });
-    }
-
-    onHardwareBackButtonPress(): boolean {
-        if (this.state.isHeaderMenuShowing) {
-            this.closeDrawer();
+        const unlisten = props.history.listen((_: Location): void => {
+            return;
+        });
```
niko-pg: I dont think this is needed? The first listener should be the one assigned to `unlisten`.

```
 const unlisten = props.history.listen((location: Location, _: Action): RouteChangedAction =>
     props.sendAnalyticsData(location, props.locale),
 );
```

src/components/main/main_component.tsx
```diff
@@ -30,106 +30,63 @@ interface MainProps {
...
-                            isDisclaimerVisible={this.state.isDisclaimerModalOpen}
-                            closeAboutModal={this.closeAboutModal}
-                            closeDisclaimerModal={this.closeDisclaimerModal}
-                        />
-                    </Container>
-                </Drawer>
-            </Root>
+    useEffect((): void => {
+        props.history.listen((location: Location, _: Action): RouteChangedAction =>
+            props.sendAnalyticsData(location, props.locale),
```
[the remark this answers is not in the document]
niko-pg: This is declaring a listener, so I think calling this once (on mount) is sufficient.

src/components/main/app_modals_component.tsx
```diff
@@ -1,25 +1,33 @@
...
             />
         </>
     );
-};
\ No newline at end of file
+};
+
+const isAboutModalVisible = (headerMenuState: HeaderMenuStore): boolean => (
+    headerMenuState === HeaderMenuStore.AboutModalIsOpen
+);
```
rasmus-storjohann-PG: Here you refactored more than you had to. When you changed the store from using booleans to using an enum, you could have isolated that change to just the store, and just changed the selectors that I think you already had to do what these functions are doing. That way selectors would have done their job of isolating the component code from the store code, letting you change one without affecting the other.

src/components/services/service_list_component.tsx
```diff
@@ -172,26 +177,34 @@ const renderServiceListItem = (props: Props): ({ item }: ServiceItemInfo) => JSX
...
-        <Text style={[textStyles.headlineH2StyleWhiteLeft, { paddingHorizontal: 10, marginBottom: 10 }]}>
-            {props.title}
-        </Text>
-        <ServiceListLocationSearchComponent
-            manualUserLocation={props.manualUserLocation}
-            setManualUserLocation={props.setManualUserLocation}
+    <View>
+        <MenuAndBackButtonHeaderComponent
+            {...props}
+            {...{ textColor: colors.white, backgroundColor: colors.teal }}
```
rasmus-storjohann-PG: I was able to pretty it up a bit:
```ts
<MenuAndBackButtonHeaderComponent
            {...{ ...props, textColor: colors.white, backgroundColor: colors.teal }}
        />
```
but the real solution is to explicitly spell out the props that you are passing in. Anyway, this is nit picky and at this stage of the PR you don't need to fix this.

src/components/explore/explore_detail_component.tsx
rasmus-storjohann-PG: It has been abstracted, `renderHeader()` is doing what you're asking for.

src/components/help/help_component.tsx
tomy-pg: The brackets are there because of the `&` I believe.

PR #1081, opened 2020-03-25
Feature/issue 1058 reintroduce load more button

ian-lin-PG:


src/components/search/api/fetch_search_results_from_query.ts
```diff
@@ -33,12 +34,14 @@ export const fetchSearchResultsFromQuery = async (searchTerm: string, latLong: L
...
             body: JSON.stringify({
                 query: searchTerm,
+                page: searchPage,
                 hitsPerPage: '20',
                 aroundLatLng: latLong ? toAlgoliaParameter(latLong) : '',
             }),
         });
 
         const responseJSON: AlgoliaResponse = await response.json();
+        setNbPages(responseJSON.nbPages);
```
tomy-pg: Not a fan of Algolia's `nbHits nbPages nbTypos...` variables.
This could just become numberOfPages.

ian-lin-PG: I changed the setter function to `setNumberOfPages`, however, the `nbHits nbPages nbTypos...` are all from algolia response and we need it for validation purposes.

Issue #1046, opened 2020-03-12
Re-introduce bookmarking to topic services and search services lists


tomy-pg:
Since we have refactored search to using Algolia's rest API in Issue #1030, we can now re-introduce bookmarking.
Bookmarking was previously removed in Issue #1026.

Unbookmarking services / topics will still result in the Flatlist to reset to the top in the bookmarks page. This will be addressed in #1047

### Acceptance Criteria: 
- [ ] Bookmarking in both topic services and search services is re-implemented

### Steps for QA:
1. Navigate to the search page and make a search
2. Bookmark some of the services in the list.
- [ ] Ensure when bookmarking that you are not brought back to the top of the page
3. Navigate to the bookmark page
- [ ] Ensure the bookmarked services are in the bookmarked services tab

tomy-pg:
This pr also addresses issue #1069

aron-chen-pg:
Tested on both Samsung 8 and iPhone x. The described behaviours were all functional.

- [ ] Bookmarking in both topic services and search services are re-implemented
- [ ]  Ensure when bookmarking that you are not brought back to the top of the page
- [ ]  Ensure the bookmarked services are in the bookmarked services tab

A couple of issues to be aware of, not sure if they already exist elsewhere.
1. On Search, clicking into a service and clicking back will bring users back to the top of the page. On Bookmarks, clicking into and back from a service will bring users back to the bookmarked topics view. This problem exists for both iOS and Android. This posts a high usability issue when users attempt to navigate between different services.
Screenshots captured on iPhone X.
![image](https://user-images.githubusercontent.com/54860686/78425073-ed486180-7626-11ea-8b8b-b0f8e8b676bd.png)
![image](https://user-images.githubusercontent.com/54860686/78425166-c63e5f80-7627-11ea-8802-0aa774e794f9.png)

2. When saving/removing a bookmark, the notification appears right on top of the bottom navigation bar. While the notifications are on, users cannot click onto any of the tabs. This issue exists for both Android and iOS. The notification should be visually above the bottom navigation bar. Please refer to [mockup](https://app.zeplin.io/project/5bc7bfa42e72260acb03aa86/screen/5dcb61e36f58b8bf61595e6f/) for notification placement
Screenshot captured on iPhone X
![91534653_1492220687624452_6781309634427748352_n](https://user-images.githubusercontent.com/54860686/78427211-c559fd80-7628-11ea-8d05-c5d9b80469f2.jpg)

[...]

PR #1086, opened 2020-03-30
Issue #1046 re-introduce bookmarking for service list items

tomy-pg:
Re-implementing the bookmarking functionality that was removed in this [issue](https://github.com/pg-irc/pathways-frontend/pulls?q=1026).

This was previously removed because the scroll position would reset when users bookmarked/unbookmarked a search service.


src/components/bookmarks/bookmarks_component.tsx
```diff
@@ -7,27 +7,25 @@ import { View, Text, Container } from 'native-base';
...
-                    <TabSwitcher i18n={i18nRenderProp.i18n} {...props}/>
-                )}
-            </I18n>
-        </Container>
-    );
-};
+export const BookmarksComponent = (props: Props): JSX.Element => (
+    <Container style={{ backgroundColor: colors.lightGrey }}>
+        <HeaderComponent/>
+        <I18n>
```
niko-pg: Not something about your review. But can be a good point of discussion. 

I think the render prop approach is the establised pattern in the codebase now but looking at the lingui docs, seems like there's a `useLingui` hook that's already available. I think this is a more convenient than the render prop pattern.

```
const Component = () => {
  const { i18n } = useLingui()

   return (
      <a href="" aria-title={i18n._("Link to documentation")}>
         <img />
      </a>
   )
};
```

src/components/bookmarks/service_bookmarks_component.tsx
```diff
@@ -40,6 +40,9 @@ export const renderServiceItems = (props: Props): ({ item }: ServiceItemInfo) =>
             service={item}
             onPress={goToRouteWithParameter(Routes.ServiceDetail, item.id, props.history)}
             currentPath={props.location.pathname}
+            isBookmarked={item.bookmarked}
+            bookmarkService={props.bookmarkService}
+            unbookmarkService={props.unbookmarkService}
```
niko-pg: NIT: Just a comment, I'll leave it to you if you want to change. But I think a better approach here is to pass an `onBoomark` and `onUnbookmark` callbacks that use the redux actions instead of passing down the raw redux action itself. 

```
const onBookmark = (service) => props.bookmarlService(service);

const onUnbookmark = (service) => props. unbookmarkService(service);
```

This approach is better because we are not giving the Presentational components any access to redux actions. Additionally, `ServiceListItemComponent` would make it easier to be reused as it's not coupled to any redux props.

src/components/bookmarks/bookmarks_component.tsx
rasmus-storjohann-PG: Looking [here](https://github.com/lingui/js-lingui/issues/390) the simplest and most up to date solution is
```ts
import { t } from "@lingui/macro"

function Example(){
  const alt = t`Image caption`

  return (
     <img src="..." alt={alt} />
  )
}
```
Let's not use hooks just for the sake of using hooks.

niko-pg: I'm not suggesting hooks just for the sake of usage. Hooks were made to replace render props and HOC. Even the link above acknowledges and they plan to remove the render props soon (`<I18n>`). So no point in using something that's going to be deprecated.

Anyhow, this was before knowing that the `t` macro can now be used directly which is a lot cleaner than using any hooks or render props so I think that's somethinf we can agree 😀

tomy-pg: The `useLingui` hook will be something we will have to migrate to when we upgrade to version 3 (we are currently on 2.9).

Just taking a look at Lingui's [documentation](https://lingui.js.org/releases/migration-3.html)

In version 3 of Lingui, it is stated in their migration guide that the `i18n` render prop have been removed for the `useLingui` hook.

rasmus-storjohann-PG: OK, I was wrong, looks like the hook will be the future here, so if it exists now, feel free to start using it.

src/components/services/render_service_items.tsx
```diff
@@ -0,0 +1,32 @@
...
+    readonly bookmarkService: (service: HumanServiceData) => BookmarkServiceAction;
+    readonly unbookmarkService: (service: HumanServiceData) => UnbookmarkServiceAction;
+}
+
+export type ServiceItemInfo = ListRenderItemInfo<HumanServiceData>;
+
+export const renderServiceItems = R.curry((props: ServiceItemsProps, itemInfo: ServiceItemInfo): JSX.Element => {
+    const service = itemInfo.item;
+    const onBookmark = (): BookmarkServiceAction => props.bookmarkService(service);
+    const onUnbookmark = (): UnbookmarkServiceAction => props.unbookmarkService(service);
```
niko-pg, line 21: I think this is good and at least the redux action is now only used by the parent. And there is now a clear distinction on which is the smart component and which is just a presentational component.

In most list scenarios I've used and saw the common pattern is more like:
```
const onBookmark = (service) => {
  props.bookmarkService(service);
}
```
Then in the ServiceListItemComponent, it's something like:
```
const ServiceListItemComponent = React.memo((props) => {
  const onBookmark = () => {
     props.onBookmark(props.service);
  }
})
```
But I do realize that the way the current service items are structured are pretty different. The example above is a common pattern for rendering FlatList or VirtualizedList. We're not using those, so no need to change it this way.

Issue #985, opened 2020-01-16
Save feedback data to airtable or similar cloud service


rasmus-storjohann-PG:
### Acceptance Criteria 
- [x] When clicking the submit button on the modal ([mock](https://zpl.io/VOnjoDl)), submit the suggested updates to the server.
- [x] There is some form of schema or plan for how the data is stored in the server, including suggested changes to specific fields, suggestions to remove a service, general "other" feedback on a server, user email, full name, SPO role.
- [x] Suggested changes are submitted only for the fields where the check box is selected ([mock](https://zpl.io/anQ019J)). (So if you check a checkbox, type in something, uncheck the checkbox and hit submit, what you just typed in is not included in the submitted data) 
- [x] After saving is complete, clear fields (to avoid spamming the same data), and show message "Thank you for your contribution" ([mock](https://zpl.io/VDd674q)).

bensinjin:
Questions:

Currently the "Remove this service" and "Other" links from the "Suggest an update" modal take you to a new screen which would have its own feedback state, then, after you press submit, and the content is sent to Airtable, you're taken back to the service detail to submit your email. If we want the email address to be submitted along with the feedback we'll likely have to move feedback into Redux so it can be accessed from both screens or if this is not how it works we'll have a new table in Airtable for email addresses + service ids?

bensinjin:
Talked to Rasmus a bit:


Ben  11:50 AM
Ok .. so currently the other feedback options actually take you to new screens in the app

rasmus  11:51 AM
We want the email attached to the feedback so we can use it to assess the feedback reliability and respond to the user as well.






Ben  11:51 AM
yeah that's what I was thinking
11:51
we may need to move the feedback into Redux then because we actually jump between different screens now to give feedback
11:52
the "other" non-field feedback is actually done on a new screen, which is kinda awkward

rasmus  11:56 AM
It's a bit of a pain to put the feedback data in redux now, but it'll give us flexibility, such as other feedback being its own screen is no longer a problem.

bensinjin:
For now I'm waiting a bit for changes from Niko and Tomy to come in as we're all needing to work on the same screens. I've got a hook in `feature/Issue_985_Send_feedback` for sending feedback. For the new Feedback screens ("Other" and "Remove service") we're going to try sending the feedback text back to the service detail screen as a query parameter which will populate the feedback object.

bensinjin:
### Testing setup
[...]

PR #1101, opened 2020-04-03
Issue #985 Send feedback to Airtable

bensinjin:
See instructions here: https://github.com/pg-irc/pathways-frontend/issues/985#issuecomment-608665614. I think when we move to production with this we'll want a dedicated account to use the Airtable API key of, not one of our own.

Note: There are existing know bugs please do not report the following on this PR:

- After navigating to new feedback screen and coming back to the service detail screen the back button needs to be pressed multiple times to work as expected
- The close button on the feedback screen opens the options menu when navigating back to the service detail screen. We'd expect the back button to do this.


src/application/routing.ts
```diff
@@ -18,9 +18,9 @@ export type MatchParameters = Partial<{
 }>;
 
 export type QueryParameters = Partial<{
-    readonly mode: 'OTHER' | 'REMOVE_SERVICE';
-    readonly optionsModalVisible: 'true' | 'false';
-    readonly receiveUpdatesModalVisible: 'true' | 'false';
+    readonly feedbackContentMode: 'OTHER' | 'REMOVE_SERVICE';
+    readonly feedbackOptionsModalMode: 'visible' | 'hidden';
+    readonly feedbackSubmitModalMode: 'visible' | 'hidden';
```
tomy-pg, line 23: This is a very nice refactor.

src/components/feedback/hooks/use_send_feedback.ts
```diff
@@ -0,0 +1,131 @@
...
+
+    return JSON.stringify({ fields: R.reduce(buildFieldValuePair, {}, feedbackFields) });
+};
+
+const getRequestUrl = (): string => {
+    if (!AIRTABLE_TABLE_ID) {
+        throw new Error('AIRTABLE_TABLE_ID is missing.');
+    }
+
+    return `https://api.airtable.com/v0/${AIRTABLE_TABLE_ID}/Table%201`;
```
tomy-pg: We should use the `build-url` package here, similar to other places in our app where we append strings to make URL.

bensinjin: Thx for pointing me to that, fixed!

src/application/routing.ts
niko-pg: Definitely better than the stringified booleans.

src/application/routing.ts
```diff
@@ -92,19 +106,16 @@ export const goToRouteWithoutParameter = (route: Routes, history: History): () =
...
+    const path = queryParams ?
+        `${routePathWithParameter(route, parameter)}${buildUrl('', { queryParams })}`
+        :
+        routePathWithParameter(route, parameter);
     // tslint:disable-next-line:no-expression-statement
     return (): void => history.push(path);
 };
 
-export const replaceRouteWithParameters = (route: Routes, parameter: string, queryParams: Record<string, string>, history: History): () => void => {
+export const replaceRouteWithParameters = (route: Routes, parameter: string, queryParams: QueryParameters, history: History): () => void => {
```
niko-pg, line 118: I guess we can rename this as `replaceRouteWithParameter` to be consistent with `goToRouteWithParameter`

src/application/toast.ts
```diff
@@ -0,0 +1,15 @@
+// tslint:disable:no-expression-statement
+import { textStyles, applicationStyles } from './styles';
+import { Toast } from 'native-base';
+
+export const showToast = (text: string): void => {
+    const toastProps = {
+        style: applicationStyles.toast,
+        textStyle: textStyles.toast,
+        duration: 2000,
+    };
```
niko-pg, line 10: Optional. Shouldn't we just pass this directly to `Toast.show`?

```
Toast.show({
        style: applicationStyles.toast,
        textStyle: textStyles.toast,
        duration: 2000,
        text,
    });
```

src/components/feedback/hooks/use_send_feedback.ts
```diff
@@ -0,0 +1,136 @@
...
+    };
+};
+
+export interface UseSendFeedback {
+    readonly isSendingFeedback: boolean;
+    readonly sendFeedback: () => Promise<void>;
+}
+
+export const useSendFeedback = (feedback: Feedback, clearFeedback: () => void): UseSendFeedback => {
+    let sendCancelled = false;
```
niko-pg, line 58: I think this should be a `ref`. What's gonna happen here is that this variable will be re-declared every re-render which is not what we want:

```
const sendCancelledRef = useRef<boolean>(false);
```
Making this a ref would persist this variable throughout the component's lifetime.

https://reactjs.org/docs/hooks-reference.html#useref

src/components/feedback/hooks/use_send_feedback.ts
```diff
@@ -0,0 +1,136 @@
...
+    useEffect(() => () => { sendCancelled = true; }, []);
```
niko-pg, line 59: Should be changed to once we use the ref:
```
sendCancelledRef.current = true;
```

src/components/feedback/hooks/use_send_feedback.ts
```diff
@@ -0,0 +1,136 @@
...
+    const [isSendingFeedback, setisSendingFeedback]: readonly[boolean, Dispatch<SetStateAction<boolean>>] = useState(false);
+    const sendFeedback = async (): Promise<void> => {
+        try {
+            setisSendingFeedback(true);
+            await handleResponse(await fetch(
```
niko-pg, line 64: NIT: Probably just me, probably better to separate into two lines so the flow is clearer?

src/components/feedback/hooks/use_send_feedback.ts
```diff
@@ -0,0 +1,136 @@
...
+    if (!AIRTABLE_API_KEY) {
+        throw new Error('AIRTABLE_API_KEY is missing.');
+    }
+
+    return {
+        method: 'POST',
+        headers: {
+            Accept: 'application/json',
+            'Content-Type': 'application/json',
+            'Authorization': `Bearer ${AIRTABLE_API_KEY}`,
```
niko-pg, line 125: NIT: unnecessary single quotes?

src/components/feedback/hooks/use_send_feedback.ts
```diff
@@ -0,0 +1,136 @@
...
+        },
+        body: feedbackJSON,
+    };
+};
+
+const handleResponse = async (response: Response): Promise<void> => {
+    if (!response.ok) {
```
niko-pg, line 132: Hey I might be missing something but where's the handling if the response is ok?

EDIT: Oh I see, we're just sending the data. we're not expecting any response

src/components/feedback/styles/index.ts
```diff
@@ -159,6 +160,11 @@ const receiveUpdatesStyles = StyleSheet.create<ReceiveUpdatesStyles>({
         paddingHorizontal: 15,
     },
 
+    finishButtonSending: {
+        // justifyContent: 'flex-start',
```
niko-pg, line 164: leftover comment?

src/hooks/use_query.ts
```diff
@@ -1,21 +1,26 @@
 // tslint:disable: no-expression-statement
 import { useLocation } from 'react-router-native';
+import { ParsedQueryParameters } from '../application/routing';
 
 // TODO Replace with a third party library once we start supporting other types.
-function parseSearch(search: string): Record<string, string> {
+function parseSearch(search: string): ParsedQueryParameters {
```
niko-pg, line 6: Cool. I like the JSON support. Not something about your review, but maybe we can use this JSON approach to support different types? Only problem is that this param object will be encapsulated into one giant search param. 
```
// Currently, visible and count will be parsed as strings (show === 'true', count === '3')
service/:serviceId?visible=true&count=3

// With the giant param approach
service/:serviceId?params="%7B%22visible%22:true,%22count%22:3%7D"

(Parsing params would preserve the type of visible as boolean and count as int)
JSON.parse(decodeURIComponent(params));
```
Just a point of discussion. A robust approach here would be to have it's own `React Context` and use Compounding Components pattern but there's not much usage yet that warrants this kind of complexity

src/components/feedback/feedback_modal_container.tsx
```diff
@@ -69,22 +80,39 @@ export const FeedbackModalContainer = ({
         setOptionsModalVisible(false);
     };
 
-    const onHideReceiveUpdatesModal = (): void => {
-        setReceiveUpdatesModalVisible(false);
+    const onHideReceiveUpdatesModal = (i18n: I18n) => async () => {
+        try {
+            await sendFeedback();
+        } finally {
```
niko-pg, line 86: Should we somehow display something if ever the sending failed?

src/application/routing.ts
bensinjin: Tidied up, thx!

src/application/toast.ts
bensinjin: Yep much better, fixed.

src/components/feedback/hooks/use_send_feedback.ts
bensinjin: Yeah I think a ref would make more sense here,  fixed.

src/components/feedback/hooks/use_send_feedback.ts
bensinjin: fixed

src/components/feedback/hooks/use_send_feedback.ts
bensinjin: Agreed, fixed.

src/components/feedback/hooks/use_send_feedback.ts
bensinjin: Nope! Fixed.

src/components/feedback/hooks/use_send_feedback.ts
bensinjin: Yeah I thought of this, I'm not sure we have any requirements at this point to handle the failure case. This is all pretty thin thread.

src/components/feedback/styles/index.ts
bensinjin: Removed.

src/hooks/use_query.ts
bensinjin: You're right, currently if you want to preserve your type you need to use JSON as the individual params are treated as strings. I did consider your "giant" param approach but then went with what I considered to be less disruptive. Either way it's a good point and I'm happy to take a crack at an implementation @rasmus-storjohann-PG do you wanna weigh in?

src/components/feedback/feedback_modal_container.tsx
bensinjin: Perhaps? It hasn't been a requirement and I think the goal is to sort of test how this is being used in the wild before we build any more than the necessities.

src/components/feedback/feedback_other_remove_service_component.tsx
```diff
@@ -182,25 +184,38 @@ const FooterComponent = (props: FooterComponentProps): JSX.Element => {
     );
 };
 
+const getFeedbackJSON = (mode: QueryParameters['feedbackContentMode'], input: string, serviceId: string): string => {
```
rasmus-storjohann-PG, line 187: So this `QueryParameters['feedbackContentMode']` way of specifying a type can be over-used, consider instead defining a type and using it both here and in the definition of `QueryParameters` instead, it's simpler and usually clearer.

src/components/feedback/feedback_other_remove_service_component.tsx
```diff
@@ -182,25 +184,38 @@ const FooterComponent = (props: FooterComponentProps): JSX.Element => {
...
+    const feedback: Feedback = { ...getEmptyFeedback(), bc211Id: { value: serviceId, shouldSend: true }};
+    const feedbackField = mode === 'OTHER' ?
+        { other: {...feedback.other, value: input } }
+        :
+        { removalReason: {...feedback.removalReason, value: input }};
+    return JSON.stringify({ ...feedback, ...feedbackField });
```
rasmus-storjohann-PG, line 193: This function does too many things, if you break it up you can make the code so much easier to understand

src/application/routing.ts
```diff
@@ -92,23 +106,13 @@ export const goToRouteWithoutParameter = (route: Routes, history: History): () =
...
+);
 
-export const replaceRouteWithParameters = (route: Routes, parameter: string, queryParams: Record<string, string>, history: History): () => void => {
-    const path = `${routePathWithParameter(route, parameter)}${buildUrl('', { queryParams })}`;
-    // tslint:disable-next-line:no-expression-statement
-    return (): void => history.replace(path);
-};
+export const replaceRouteWithParameter = (route: Routes, parameter: string, history: History, queryParams?: QueryParameters): () => void => (
+    (): void => history.replace(getPathWithOptionalQuery(routePathWithParameter(route, parameter), queryParams))
+);
```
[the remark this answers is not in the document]
bensinjin: This is equivalent to `goToRouteWithParameter()` but it calls `replace()` instead of `push()`. It's not specifically for feedback and wasn't introduced with this work just tidied up a bit. We previously decided to support optional query parameters and I'm not sure it makes sense to make them required? Do you want new functions specifically for routes with query parameters?

src/hooks/use_query.ts
```diff
@@ -1,21 +1,26 @@
...
+        } catch (error) {
+            result[key] = value;
+        }
     }
 
     return result;
 }
 
-export const useQuery = (): Record<string, string> => {
+export const useQuery = (): ParsedQueryParameters => {
```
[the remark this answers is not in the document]
bensinjin: It's actually not meant to be specific to feedback it just so happens that currently only feedback related things use it. The idea is that we get query parameters in a string form but then turn them into and object form. We could make the `ParsedQueryParameters` type the union of types that represent the expected query parameters on a screen by screen basis or something? So you'd have something like: `type ParsedQueryParameters = FeedbackScreenParams & ServiceDetailScreenParams` - might add clarity?

src/components/feedback/feedback_modal_container.tsx
```diff
@@ -69,22 +80,39 @@ export const FeedbackModalContainer = ({
...
+                            isVisible={receiveUpdatesModalVisible}
+                            onHide={onHideReceiveUpdatesModal(i18n)}
+                            isSendingFeedback={isSendingFeedback}
+                            setFeedback={setFeedback}
+                            feedback={feedback}
+                        />
+                    </>
+                )
+            }
+        </I18n>
```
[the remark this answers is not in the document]
bensinjin: I don't think we have access to it yet without a lingui upgrade.

src/application/routing.ts
```diff
@@ -136,4 +140,8 @@ export const getParametersFromPath = (location: Location, route: Routes): MatchP
         exact: true,
     });
     return match.params;
-};
\ No newline at end of file
+};
+
+const getPathWithOptionalQuery = (path: string, queryParams?: QueryParameters): string => (
```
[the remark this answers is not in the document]
bensinjin: I'm not sure I follow .. it builds path with a query parameter if it's provided.

src/application/routing.ts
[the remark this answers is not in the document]
rasmus-storjohann-PG: It doesn't look specific to feedback, but `QueryParameters` makes it feedback specific, and if `QueryParameters` was better named, it would be readily apparent. Making it generic would lead to it taking a type parameter `<T>`, which I don't think is a good way to go.

src/hooks/use_query.ts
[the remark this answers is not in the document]
rasmus-storjohann-PG: So I have only recently learned that this anti-pattern has a name, Avoid Hasty Abstractions. You anticipate future evolution of the code base and write the code accordingly. It is much better to write the code to solve the specific problem if your current story, and worry about what it will evolve into later. Two reasons for this, the first is that we're often wrong about how our code base will evolve, and secondly, the abstraction makes the code harder to understand. When the abstraction is good, this is a price worth paying, but when it is hasty, as in this case, the cost outweighs the benefit.

src/application/routing.ts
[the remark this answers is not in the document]
rasmus-storjohann-PG: Rename `QueryParameters` to something that reflects what it actually represents, and it'll become apparent

src/components/feedback/feedback_modal_container.tsx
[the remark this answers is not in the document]
rasmus-storjohann-PG: I don't think this kind of change needs to be done in a big sweep over the code base, I'd be fine with some code using this form and some use the hook, once it becomes available.

src/application/routing.ts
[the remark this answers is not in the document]
rasmus-storjohann-PG: https://www.sandimetz.com/blog/2016/1/20/the-wrong-abstraction

src/hooks/use_query.ts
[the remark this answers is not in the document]
niko-pg: I think what happened here is that since `useQuery` is related to `application/routing`, it made sense to put it on the app level directory. I guess name can be improved to `useUrlQuery` ?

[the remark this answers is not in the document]
bensinjin: Yep, although I don't know if I consider taking a type that could've been anything (it was previously an object with a string key and string value) and narrowing it to something that better represents the possible values we expect a hasty abstraction. Maybe it's the name `ParsedQueryParameters` you don't like? Keep in mind we're just getting url query parameters here it's not generally a very strongly typed data format.

src/components/feedback/feedback_other_remove_service_component.tsx
```diff
@@ -20,15 +20,17 @@ import React, { Dispatch, SetStateAction, useState } from 'react';
...
 
 import { colors, textStyles } from '../../application/styles';
-import { Routes, replaceRouteWithParameters } from '../../application/routing';
+import { Routes, replaceRouteWithParameter, QueryParameters } from '../../application/routing';
 import { useQuery } from '../../hooks/use_query';
 import { CloseButtonComponent } from '../close_button/close_button_component';
 import { MultiLineButtonComponent } from '../mutiline_button/multiline_button_component';
 
 import { otherRemoveServiceStyles as styles } from './styles';
+import { getEmptyFeedback, Feedback } from './hooks/use_send_feedback';
```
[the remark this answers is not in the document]
niko-pg: I misplaced my comment in another thread. Pasting it here:

I think what happened here is that since useQuery is related to `application/routing`, it made sense to put it on the app level directory. I guess name can be improved to `useUrlQuery`?

src/components/feedback/hooks/use_send_feedback.ts
```diff
@@ -0,0 +1,131 @@
...
+        authorName: emptyFeedbackField,
+        authorJobTitle: emptyFeedbackField,
+        authorOrganization: emptyFeedbackField,
+    };
+};
+
+export interface UseSendFeedback {
+    readonly isSendingFeedback: boolean;
+    readonly sendFeedback: () => Promise<void>;
+}
```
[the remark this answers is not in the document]
bensinjin: I see your rename, I guess it's ok but this type is not technically a "SendFeedbackPromise".

src/application/routing.ts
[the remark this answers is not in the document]
bensinjin: What would you like done here?

src/hooks/use_query.ts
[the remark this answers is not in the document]
bensinjin: I think avoiding the most obvious abstraction - parsed query parameters - hurts readability here. It's a common abstraction that people understand and I tend to agree if it's confusing it can be remedied by maybe adding "Url". Technically this function returns a very generic object, saying what we expect is _for_ "feedback" is even further away from saying it's for "Screen A" or "Screen B" which is how you generally treat url query parameters.

Either way, you've got the rename in place, what else needs to be fixed? Should this be `useFeedbackQuery`?

src/components/feedback/feedback_modal_container.tsx
```diff
@@ -1,51 +1,62 @@
...
 import { FeedbackOptionsModalComponent } from './feedback_options_modal_component';
+import { UseSendFeedback, Feedback } from './hooks/use_send_feedback';
+import { showToast } from '../../application/toast';
 
 interface FeedbackModalContainerProps {
     readonly feedbackEnabled: boolean;
     readonly onSuggestAnUpdatePress: () => void;
     readonly serviceId: string;
+    readonly query: ParsedQueryParameters;
+    readonly sendFeedback: UseSendFeedback['sendFeedback'];
```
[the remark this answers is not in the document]
bensinjin: You fixed.

src/components/feedback/feedback_modal_container.tsx
```diff
@@ -1,51 +1,62 @@
...
 
 export const FeedbackModalContainer = ({
     feedbackEnabled,
     onSuggestAnUpdatePress: onSuggestAnUpdate,
     serviceId,
+    query,
+    sendFeedback,
+    isSendingFeedback,
+    setFeedback,
+    feedback,
```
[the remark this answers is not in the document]
bensinjin: So should we be changing the whole function to do `props.`?

src/components/feedback/feedback_other_remove_service_component.tsx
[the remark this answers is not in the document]
bensinjin: Do you want this hook moved into the `services` directory because that's where it's used?

src/application/routing.ts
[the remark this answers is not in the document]
bensinjin: Should this be renamed to `getPathWithOptionalFeedbackState`?

src/components/feedback/feedback_other_remove_service_component.tsx
```diff
@@ -70,15 +72,14 @@ const SUGGESTION_CONTENT: SuggestionContentMap = {
     },
 };
 
-const useServiceDetailRoute = (queryParam: Record<string, string>): () => void => {
-    const params = useParams<RouteParams>();
+const useServiceDetailRoute = (serviceId: string, queryParam: QueryParameters): () => void => {
```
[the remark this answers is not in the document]
bensinjin: Agreed, fixed.

src/application/routing.ts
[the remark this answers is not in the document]
bensinjin: I'll try and come up with a better name.

src/components/feedback/hooks/use_send_feedback.ts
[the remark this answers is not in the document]
rasmus-storjohann-PG: Feel free to tweak the names

src/application/routing.ts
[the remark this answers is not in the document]
bensinjin: Fixed.

src/hooks/use_query.ts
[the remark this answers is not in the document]
bensinjin: Fixed.

src/application/routing.ts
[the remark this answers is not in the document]
bensinjin: Fixed, I've chosen the aforementioned name.

src/components/feedback/feedback_other_remove_service_component.tsx
bensinjin: Fixed.

src/components/feedback/feedback_other_remove_service_component.tsx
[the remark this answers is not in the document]
bensinjin: Fixed. I've move this into the services directory and removed the `hooks` dir.

src/components/feedback/feedback_other_remove_service_component.tsx
bensinjin: Fixed.

src/hooks/use_query.ts
rasmus-storjohann-PG: I haven't responded because I'm totally lost here. I am going to review this PR and just ask questions.

src/components/services/use_send_feedback.ts
```diff
@@ -0,0 +1,136 @@
...
+    readonly phone: FeedbackField;
+    readonly website: FeedbackField;
+    readonly email: FeedbackField;
+    readonly removalReason: FeedbackField;
+    readonly other: FeedbackField;
+    readonly authorIsEmployee: FeedbackField;
+    readonly authorEmail: FeedbackField;
+    readonly authorName: FeedbackField;
+    readonly authorOrganization: FeedbackField;
+    readonly authorJobTitle: FeedbackField;
```
rasmus-storjohann-PG, line 27: Typescript has a great type system, you should use it. All the author properties are optional, make it into its own interface and make that optional in this one. The feedback data is either name, org, etc, OR it's removalReason, OR it's other feedback. Use the type system to express these things.

src/components/feedback/feedback_modal_container.tsx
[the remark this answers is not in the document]
rasmus-storjohann-PG: Yes

src/components/feedback/feedback_receive_updates_modal.tsx
```diff
@@ -1,38 +1,50 @@
...
-    const [emailInput, setEmailInput]: readonly [string, Dispatch<SetStateAction<string>>]
-        = useState<string>('');
+    const onPressIsEmployee = (): void =>
+        setFeedback({
+            ...feedback,
+            authorIsEmployee: {
+                ...feedback.authorIsEmployee,
+                value: feedback.authorIsEmployee.value === 'true' ? 'false' : 'true',
+            },
+        });
```
rasmus-storjohann-PG, line 45: Complex state update code in component is smelly, there is no separation of concerns. With my changes, this all happens in a redux reducer, simplifying the component code.

PR #1102, opened 2020-04-06
Feature/issue 1096 fix extra margin on status bar

tomy-pg:


src/components/main/get_status_bar_height_for_platform.ts
```diff
@@ -0,0 +1,13 @@
...
+import Constants, { AppOwnership } from 'expo-constants';
+
+export const getStatusBarHeightForPlatform = (): number => {
+    if (Platform.OS === 'ios') {
+        return 0;
+    }
+    return isStandaloneAndroidBuild() ? StatusBar.currentHeight : 0;
+};
+
+const isStandaloneAndroidBuild = (): boolean => (
```
bensinjin, line 11: This technically doesn't check if this is an Android build, maybe make it return something like:
```
Platform.OS === 'android' && Contants.appOwnership === AppOwnership.Standalone
```
Then you can just make your `getStatusBarHeightForPlatform()` return something like:
```
isStandaloneAndroidBuild() ? StatusBar.currentHeight : 0;
```

rasmus-storjohann-PG: We already have `isAndroid()`, it would be nice to use it consistently, that way we have a way to find all the platform specific logic in the app.

tomy-pg:
@rasmus-storjohann-PG thanks for pointing that out. I think I have addressed this issue. I had also tried some of the solutions people were proposing in the thread below, such as setting properties in app.json, but it seemed like adding a Statusbar to the main component was the only one that worked:
https://github.com/expo/expo/issues/7635

Issue #1076, opened 2020-03-25
Update "Information is only in English" dialog to indicate that "Search is in beta"


cherrielam:
On the Search page, we would like to update the dialog that informs users that the services information is only in English to also say that "Search is in beta". 

Mock is now here: https://zpl.io/bAyJZ88 

The copy that needs to be translated is: "Search is currently in beta and only available in English." and "Give Feedback" - @jparlina and @tomy-pg can you add to weblate for translations?

We have removed the "Don't show again" button on this dialog to reduce the clutter of buttons, and replaced it with the "Give Feedback" button. This means that the _Search in beta_ dialog is not dismissable, but it does scroll with the rest of the cards, so should hopefully not be too intrusive.

Clicking "Give Feedback" should open an email titled "Arrival Advisor - Feedback on Search" in the subject line, to: info@arrivaladvisor.ca, and if possible, in the body: Version # - populate with the app version that the user has.

**Note:** On _topic-services_, the dialog should still be the previous one that says "Information about services is currently only available in English.... Don't show again". So the dialogs on _topic-services_ and _Search_ will be different.

Also the padding around the text in the dialog is different from the Services cards. Please make the padding the same.

### Acceptance Criteria 
- [ ]  On _Search_, the dialog should now say "Search is currently in beta and only available in English. For support in other languages, please call BC211" and button "Give Feedback"
- [ ] The dialog is localized in all languages
- [ ] Clicking "Give Feedback" opens the user's email client with subject line: "Arrival Advisor - Feedback on Search" mailto: info@arrivaladvisor.ca, and if possible Version # of the user's app in the email body.
- [ ] The _Search_ dialog is different from the dialog on _topic-services_. _Topic-services_' dialog should still say "Information about services is currently only available in English... Don't show again"
- [ ] _Topic-services_ dialog is dismissable, _Search_ dialog is not. Both scroll with the services.
- [ ] Padding around the text in the dialog is the same as the padding around the text in the Services cards.

### Steps for QA: 
1.

tomy-pg:
The new copy has been added to Weblate.

tomy-pg:
QA:
Get Feedback is not translated in app because of the difference between capitalization in the source string and the app string. This has been fixed in Weblate. See here: https://github.com/pg-irc/ui-strings/commit/ac48f7684bf6e28809fed588d40f4389fbfcbe7d

jparlina:
[...]

PR #1105, opened 2020-04-06
Issue #1076 Apply adjustments to services message card

bensinjin:


src/components/partial_localization/message_component.tsx
```diff
@@ -1,64 +1,44 @@
...
-import { HidePartialLocalizationMessageAction } from '../../stores/user_profile';
 import { View, Text } from 'native-base';
 import { textStyles, colors } from '../../application/styles';
-import { Trans } from '@lingui/react';
 import { TouchableOpacity } from 'react-native';
-import { openURL } from '../link/link';
 import { EmptyComponent } from '../empty_component/empty_component';
 
-export interface MessageComponentProps {
+export interface Props {
```
rasmus-storjohann-PG, line 7: We usually avoid exporting types called `Props` to not pollute the larger namespaces with non-specific names like this. Do you have a particular reason to make this change?

bensinjin: It was meant to just be renamed because it was exported but never imported anywhere else. Fixed the "export" part.

rasmus-storjohann-PG: Yep, if it's not exported, then Props is the preferred name

Issue #1103, opened 2020-04-06
Arabic search input and placeholder is LTR


tomy-pg:
Arabic text should flow right-to-left.
![Screen Shot 2020-04-06 at 11.12.31 AM.png](https://images.zenhubusercontent.com/5c6c8cff9d62b128eab447ff/54a33b26-2e51-45bf-bc81-61c7602dfad8)

### Acceptance Criteria 
- [ ]  Arabic search input and placeholders flow right-to-left.

### Steps for QA: 
1.

tomy-pg:
Staging version 1.27.0:
Pass

PR #1110, opened 2020-04-07
Issue #1103 Remove dead style code, fix LTR placeholder in RTL env

bensinjin:


src/application/styles.ts
```diff
@@ -170,9 +172,6 @@ export const textStyles = StyleSheet.create({
         letterSpacing,
         ...getNormalFontStylesForOS(),
     },
-    alwaysLeftAlign: {
```
bensinjin, line 173: This, and the style removed below were not in use.

PR #1111, opened 2020-04-07
Feature/issue 935 add search google analytic

ian-lin-PG:


src/sagas/analytics/events.ts
```diff
@@ -35,6 +35,12 @@ export const sendSearchQueriedEvent = (searchTerm: string, searchLocation: strin
     analytics.hit(event);
 };
 
+export const sendOpenService = (service: HumanServiceData): void => {
```
tomy-pg, line 38: I think `sendServiceDetailPressed` could be a clearer name. What do you think?

src/sagas/analytics/events.ts
```diff
@@ -72,7 +78,7 @@ const createScreenHit = (path: string): any => (
 );
 
 // tslint:disable-next-line:no-any
-const createEvent = (category: string, action: string, label?: string, value?: number): any => (
+const createEvent = (category: string, action: string, label?: string | readonly string[], value?: number): any => (
```
rasmus-storjohann-PG, line 94: This is smelly, I prefer functions with well defined parameter types. Both optional and union types are best avoided. You are going to have to concatenate these sub-labels anyway, right? So the caller might as well do that and pass in the one label.

src/sagas/analytics/events.ts
```diff
@@ -16,9 +17,15 @@ export const sendAnswerChosenEvent = (answerId: string): void => {
     analytics.hit(event);
 };
 
-export const sendBookmarkAddedEvent = (topicId: string): void => {
+export const sendBookmarkTopicEvent = (topicId: string): void => {
     const analytics = createAnalytics();
-    const event = createEvent('Bookmarks', 'BookmarkAdded', topicId);
+    const event = createEvent('Bookmarks', 'TopicBookmarkAdded', topicId);
+    analytics.hit(event);
```
rasmus-storjohann-PG, line 23: There's a lot of repetition of  `const analytics = createAnalytics();` and `analytics.hit(event);` here, it can be DRYed up

src/sagas/analytics/watch_analytics.ts
```diff
@@ -67,6 +69,9 @@ async function sendAnalyticsDataAsync(action: WatchedAction): Promise<void> {
         case constants.BOOKMARK_SERVICE:
             events.sendBookmarkServiceEvent(action.payload.service);
             break;
+        case constants.OPEN_SERVICE:
+            events.sendOpenService(action.payload.service);
+            break;
```
rasmus-storjohann-PG, line 74: Hmmm, I'm wondering whether we could just watch the existing route transition action `RouteChangedAction` for this. With this new event, we're emitting two actions for one UX interaction, which is not parsinomious.

src/components/services/render_service_items.tsx
```diff
@@ -19,11 +20,16 @@ export const renderServiceItems = R.curry((props: ServiceItemsProps, itemInfo: S
     const service = itemInfo.item;
     const onBookmark = (): BookmarkServiceAction => props.bookmarkService(service);
     const onUnbookmark = (): UnbookmarkServiceAction => props.unbookmarkService(service);
+    const onOpenService = (): void => {
+        // tslint:disable: no-expression-statement
+        props.openService(service);
+        goToRouteWithParameter(Routes.ServiceDetail, service.id, props.history)();
```
rasmus-storjohann-PG, line 26: Yeah, you should not have to do both of these two things....

src/components/services/service_list_connected_component.tsx
```diff
@@ -1,7 +1,9 @@
 import { Dispatch } from 'redux';
 import { Store } from '../../stores';
-import { buildServicesRequest, BuildServicesRequestAction, ServicesAction, BookmarkServiceAction,
-        UnbookmarkServiceAction, bookmarkService, unbookmarkService } from '../../stores/services/actions';
+import {
+    buildServicesRequest, BuildServicesRequestAction, ServicesAction, BookmarkServiceAction,
+    UnbookmarkServiceAction, bookmarkService, unbookmarkService, openService, OpenServiceAction,
+} from '../../stores/services/actions';
```
rasmus-storjohann-PG, line 6: So there are several examples of code like this in the code base, but I do prefer to avoid long lists of imports like this, it's feels very natural to end up with
```ts
import * as actions from '../../stores/services/actions';
```
and then `actions.buildServicesRequest()` at the call site. I really like that we can use this namespace syntax when we're importing a lot of symbols from a module, but not if we only import a few.

src/sagas/analytics/watch_analytics.ts
rasmus-storjohann-PG: It means you have to call a selector on the store to get the additional data off of the service, looking forward to seeing the solution to that. ;-)

src/components/services/render_service_items.tsx
```diff
@@ -19,11 +20,16 @@ export const renderServiceItems = R.curry((props: ServiceItemsProps, itemInfo: S
     const service = itemInfo.item;
     const onBookmark = (): BookmarkServiceAction => props.bookmarkService(service);
     const onUnbookmark = (): UnbookmarkServiceAction => props.unbookmarkService(service);
+    const onOpenService = (): void => {
+        // tslint:disable: no-expression-statement
```
rasmus-storjohann-PG, line 24: So this should be `tslint:disable-next-line`, not `tslint:disable`. While you're fixing that, can you can the codebase for other places where this mistake has been made. I'd expect all comments like this that are not at the top of a file should be `tslint:disable-next-line`

src/sagas/analytics/watch_analytics.ts
ian-lin-PG: `RouteChangedAction` will only have service id following the path format of `/service/:serviceId`. I don't know if just having the serviceId will be meaningful. I can change the `RouteChangedAction` for service to include `service_name` and `organization_name` as url parameters.

I think we need to decide on if additional service information will be meaningful and how to go about collecting those information.

niko-pg: Would be interesting to see if the url search params that  was implemented would be applicable here? 

https://github.com/pg-irc/pathways-frontend/pull/1101/files/fd12583e51c2d1b19da99a1d52204c0adf44fa7d#diff-bcf71221d616232d3c3689a7eda9a062R109 

Search params would make sense for ephemeral data. But if the information is available in the store, I think that would be better

src/components/services/service_list_connected_component.tsx
rasmus-storjohann-PG: Oh it wouldn't be `actions['UserLocation']`, that's awful, it would be `actions.UserLocation`.

niko-pg: Another idea to solve this problem is by having `types` be on their own file. This is one of the patterns I've  seen especially when types get reused in other files. By doing this we'll prevent having a lot of imports under one module. Downside is there's going to be a lot of back and forth happening and more boilder plate code 

import {
    buildServicesRequest, bookmarkService, unbookmarkService, openService, 
} from '../../stores/services/actions';
import {  
    BuildServicesRequestAction, 
    ServicesAction, 
    BookmarkServiceAction,
    UnbookmarkServiceAction, 
} from '../../stores/services/types';

What do you think? @rasmus-storjohann-PG

src/sagas/analytics/watch_analytics.ts
ian-lin-PG: @rasmus-storjohann-PG having looked around the `RouteChangedAction ` on google analytics, I don't see any place to add in the additional service information after picking off the information from the store. The screenhit event only takes in the current path and presents that information in the analytics.

src/components/services/service_list_connected_component.tsx
rasmus-storjohann-PG: Putting types in a separate file is not a bad pattern, I've done that in a few places in this codebase. It would naturally lend itself to namespace style imports too, which I generally like.

src/sagas/analytics/events.ts
rasmus-storjohann-PG: To be more specific, I am not opposed to union types, they can be really expressive, but `string | string[]` is smelly because it's kinda like "just throw a string in there, or multiple if you have them". Compare to the feedback UX where feedback can be `ServiceFeedback | RemoveServiceFeedback | OtherFeedback`, in that context, union type is a much better tool. Make sense?

src/sagas/analytics/events.ts
ian-lin-PG: I agree open `sendOpenService` is a bit vague, I propose a combination of `sendServiceDetailOpened`. We are not pressing the ServiceDetail per se, but we are opening the ServiceDetail page.

src/components/services/render_service_items.tsx
ian-lin-PG: I chose `tslint:disable` because the following 2 lines both need to have the linter ignored. Duplicating the `tslint:disable-next-line` twice in a row looks too smelly, I think it's fair that I move this statement to the top of the file.

Issue #1120, opened 2020-04-16
Subsequent search should direct users to top of search results


jparlina:
On Staging V1.28.0 Server V1.3.27. for both iOS and Android:
When doing a new search after scrolling down to the middle of previous search results, the screen is still stuck in the middle of that previous search result. Subsequent search should push users back to the top of the search results. 

### Acceptance Criteria 
- [ ]  Subsequent search should direct users to the top of the new search results

### Steps for QA: 
1. Go to Search feature
2. Enter keyword and location, click Search
3. Scroll down the search results 
4. Enter a new keyword and/or location, then click Search
- [x] You should be directed to the top of the new search results

jparlina:
QA'd on staging  V 1.29.0 Server 1.3.28 on iOS - PASS

PR #1121, opened 2020-04-16
Issue #1120 scroll to top when new search query is executed

tomy-pg:


src/components/search/search_results_component.tsx
```diff
@@ -65,20 +66,36 @@ export const SearchResultsComponent = (props: Props): JSX.Element => {
...
+            flatListRef.current.scrollToOffset({ animated: false, offset: props.searchOffset });
+        }
+    }, [props.searchOffset]);
+    return (
+        <View style={{ flexDirection: 'column', backgroundColor: colors.lightGrey, flex: 1 }}>
+            {renderLoadingScreen(props.isLoading)}
+            <FlatList
+                ref={flatListRef}
+                onScroll={(e: any): void => props.setScrollOffset(e.nativeEvent.contentOffset.y)}
+                initialNumToRender={props.searchOffset ? props.searchResults.length : 20}
```
rasmus-storjohann-PG, line 87: Very nice solution

src/components/search/search_results_component.tsx
```diff
@@ -89,6 +106,7 @@ const renderSearchHit = R.curry((props: Props, itemInfo: ListRenderItemInfo<Sear
     const service: HumanServiceData = toHumanServiceData(item, props.bookmarkedServicesIds);
     const onPress = (): void => {
         props.saveService(service);
+        props.saveSearchOffset(props.scrollOffset);
         props.openServiceDetail(service);
         goToRouteWithParameter(Routes.ServiceDetail, service.id, props.history)();
```
rasmus-storjohann-PG, line 111: This is starting to feel like tech debt, having to do four things that are really mostly the same thing again and again, but I don't want it fixed in this PR. It's more something to think about how this could be simplified.

src/components/search/search_results_component.tsx
```diff
@@ -65,20 +66,36 @@ export const SearchResultsComponent = (props: Props): JSX.Element => {
...
+    useEffect((): void => {
+        if (props.searchTerm) {
+            flatListRef.current.scrollToOffset({ animated: false, offset: 0 });
+        }
+    }, [props.searchTerm, props.searchLocation]);
+    useEffect((): void => {
+        if (props.searchResults.length > 0) {
+            flatListRef.current.scrollToOffset({ animated: false, offset: props.searchOffset });
+        }
+    }, [props.searchOffset]);
```
rasmus-storjohann-PG, line 80: The useEffects are nice, the second argument is used effectively. I would say it could be made even nicer with making a function with a clear name to show what's happening, something like this
```ts
    useEffect((): void => {
        scrollToOffset(flatListRef, props);
    }, [props.searchOffset]);
```

src/stores/persisted_data.ts
```diff
@@ -17,6 +17,7 @@ export interface PersistedData {
     readonly searchLatLong: LatLong;
     readonly searchPage: number;
     readonly numberOfSearchPages: number;
+    readonly searchOffset: number;
```
rasmus-storjohann-PG, line 20: The only thing I don't understand in this PR is why the search offset needs to be persisted.

PR #1132, opened 2020-04-23
Issue 985 & Friends - Send feedback and move modal state to Redux

bensinjin:
After discussions with Niko and Rasmus we came to the conclusion feedback related state should be moved into Redux. This is the gist of what's in this beastly (sorry!) PR:

- All feedback state has been moved to Redux (thx Rasmus) this includes what screen your on, what modal your viewing, and the feedback data itself
- Missing feedback discard confirmation modal has been added, this was necessary in order to properly test the functionality
- Feedback is pulled from the store into a format that can be JSON.stringified and is then sent to Airtable

Throughout my work on this piece I had to decide on a format for components that have multiple local variables, hook variables and callbacks. I landed on a format that organizes the component as you would a class, for example:
```
ServiceDetailComponent = () => {
    const myVar = 'var'; // Local variables first
    const [myState, setMyState] = useState(); // Hooks second
    useEffect(doSomething, [props.something])

    function doSomething() { } // Hook effects or other hook related functions (these can't be expressions because we require them to be hoisted)

    const onSomeButtonPress = () => undefined; // Other callbacks or "methods"

    return <View>...</View>
} 
```
Let me know what you guys think.


src/components/services/service_detail_component.tsx
```diff
@@ -127,15 +131,35 @@ export const ServiceDetailComponent = (props: Props): JSX.Element => {
...
 
-    const closeModal = (): void => {
+    const closeChooseFeedbackModeModal = (): void => {
         props.close();
     };
 
-    // TODO Wire this up to the submit button for service feedback.
-    // const onSubmitPress = (): void => {
+    // TODO Wire up to submit button.
+    // const onSubmit = (): void => {
```
bensinjin, line 142: @tomy-pg Uncomment this ,the onBackPress, and onClosePress for your work that's left to do when we're in edit mode for the service.

src/components/services/service_detail_connected_component.tsx
```diff
@@ -9,23 +9,76 @@ import { selectBookmarkedServicesIds } from '../../selectors/services/select_boo
...
+    FinishAction,
+    close,
+    CloseAction,
+    DiscardChangesAction,
+    discardChanges,
+    BackAction,
+    back,
+    CancelDiscardChangesAction,
+    cancelDiscardChanges,
+} from '../../stores/feedback';
```
rasmus-storjohann-PG, line 37: See, I think this whole file would probably benefit from this just being `import * as stores from '../../stores/feedback'`, but it's up to you.

src/selectors/__tests__/feedback.test.ts
```diff
@@ -0,0 +1,164 @@
...
+import { Feedback, UserInformation } from '../../stores/feedback/types';
+import { getEmptyServiceFeedback } from '../../stores/feedback';
+
+describe('picking feedback properties to send with: pickSendableFeedback()', () => {
+
+    it('will pick properties that are strings', () => {
+        const feedback: Feedback = {
+            type: 'other_feedback',
+            value: aString(),
+        };
```
rasmus-storjohann-PG, line 16: This is where you can use `aString()` to improve the readability of your test. Reading this test, I get the sense that `'other_feedback'` is a special value that affects the behaviour of the code being tested, but is that true? As I dig, I see that it is, but this code is still confusing... heading over to the definition of `pickSendableFeedback()` now...

src/selectors/feedback/pick_sendable_feedback.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export const pickSendableFeedback = (store: FeedbackStore): Feedback => (
+    R.pickBy(
+        isStringOrSendableFeedbackField,
+        store.feedback,
+    )
+);
+
+const isStringOrSendableFeedbackField = (item: string | FeedbackField): boolean =>
+    typeof item === 'string' || item.shouldSend;
```
rasmus-storjohann-PG, line 12: So this helper function here is I think confusing things, because it kinda does more than one thing... as given away by the types it accepts. I think if you can simplify this function to just be `(item: string | FeedbackField): boolean` you'd have some ripple effects through the PR that would make things nicer...

rasmus-storjohann-PG: OK, so this is the simplification: In pickSendableFeedback, you can test the `type` field on the `store.feedback`, and from that the compiler will understand which of the three different types of feedback you're working with. The you can fork off one branch that handles the type with the `FeedbackField` members and a different branch to handle the types that have the string. It'll be nicer.

src/selectors/feedback/to_feedback_post_data_author.ts
```diff
@@ -0,0 +1,10 @@
+import { UserInformation } from '../../stores/feedback/types';
+import { FeedbackPostDataAuthor }  from './types';
+
+export const toFeedbackPostDataAuthor = (userInformation: UserInformation): FeedbackPostDataAuthor => ({
+    authorIsEmployee: userInformation.isEmployee === true ? 'true' : undefined,
+    authorEmail: userInformation.email,
+    authorName: userInformation.name,
+    authorJobTitle: userInformation.jobTitle,
+    authorOrganization: userInformation.organizationName,
+});
```
rasmus-storjohann-PG, line 10: Nice, keep the store format and the wire format independent

src/stores/feedback/index.ts
```diff
@@ -0,0 +1,73 @@
...
+
+export const getEmptyUserInfo = (): UserInformation => ({
+    email: '',
+    name: '',
+    organizationName: '',
+    jobTitle: '',
+    isEmployee: undefined,
+});
+
+export const getEmptyServiceFeedback = (config: { readonly shouldSend: boolean } = undefined): ServiceFeedback => {
```
rasmus-storjohann-PG, line 60: The config looks overly complicated, can't you just pass in a `boolean` or at most a `boolean?`?

src/stores/feedback/reducer.ts
```diff
@@ -0,0 +1,64 @@
...
+            return {...store, screen: FeedbackScreen.ServiceDetail, modal: FeedbackModal.None };
+        default:
+            return store;
+    }
+};
+
+const submitOrDiscardReducer = (store: FeedbackStore, action: ReducerActions): FeedbackStore => {
+    switch (action.type) {
+        case constants.SUBMIT:
+            return { ...store, modal: FeedbackModal.ReceiveUpdatesModal, feedback: action.payload.feedback, screen: FeedbackScreen.ServiceDetail };
```
rasmus-storjohann-PG, line 52: I was thinking of this flow, do you always want ServiceDetail here? What if you were on other feedback and hit submit, you want to stay on the OtherScreen, no?

src/components/feedback/feedback_modal_container.tsx
```diff
@@ -1,110 +1,46 @@
...
+            isSendingFeedback={props.isSendingFeedback}
+            setUserInformation={props.setUserInformation}
+            userInformation={props.userInformation}
+        />
+        <FeedbackDiscardChangesModal
+            isVisible={props.showFeedbackDiscardChangesModal}
+            onDiscardPress={props.onDiscardPress}
+            onKeepEditingPress={props.onKeepEditingPress}
+        />
+    </>
```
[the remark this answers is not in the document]
bensinjin: Fixed

src/components/feedback/feedback_other_remove_service_component.tsx
```diff
@@ -182,32 +162,58 @@ const FooterComponent = (props: FooterComponentProps): JSX.Element => {
...
+        props.close();
+    };
+
+    const onDiscardPress = (): void => {
+        props.discardFeedback();
+    };
+
+    const onKeepEditingPress = (): void => {
+        props.cancelDiscardFeedback();
+    };
```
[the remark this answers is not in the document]
bensinjin: Fixed. I think the idea here was to avoid child components that expected functions that returned store objects, to sort of create a boundary between the connected component and "the others". After pondering this a bit longer I came to the conclusion this line of thinking is more applicable to reusable components, does that seem right? Another thing that I noticed is the code is much clearer if we avoid using the super generic (albeit common) `onSomeButtonPress()` for naming our on press callbacks.

src/components/feedback/feedback_other_remove_service_component.tsx
```diff
@@ -182,32 +162,58 @@ const FooterComponent = (props: FooterComponentProps): JSX.Element => {
...
-    // TODO: Formulate a more robust typed query param getter
-    const content: SuggestionContent = SUGGESTION_CONTENT[query.mode];
+export const FeedbackOtherRemoveServiceComponent = (props: FeedbackOtherRemoveServiceProps): JSX.Element => {
+    const isOtherFeedback = props.feedbackScreen === FeedbackScreen.OtherChangesPage;
+    const content: SuggestionContent = isOtherFeedback ? SUGGESTION_CONTENT.OTHER : SUGGESTION_CONTENT.REMOVE_SERVICE;
+    const history = useHistory();
+    const [feedback, setFeedback]: readonly[string, Dispatch<SetStateAction<string>>] = useState<string>('');
+    useEffect(navigateOnScreenChange, [props.feedbackScreen]);
+
+    function navigateOnScreenChange(): void {
```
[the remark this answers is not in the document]
bensinjin: Fixed - now called "navigateToFeedbackScreen" which maybe makes it more obvious.

src/components/services/service_detail_component.tsx
```diff
@@ -32,121 +32,224 @@ import { BookmarkServiceAction, UnbookmarkServiceAction } from '../../stores/ser
...
+        props.chooseRemoveService();
+    };
+
+    const onOtherChangesPress = (): void => {
+        props.chooseOtherChanges();
+    };
+
+    const closeChooseFeedbackModeModal = (): void => {
+        props.close();
+    };
```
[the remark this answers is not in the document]

src/components/services/service_detail_component.tsx
```diff
@@ -32,121 +32,224 @@ import { BookmarkServiceAction, UnbookmarkServiceAction } from '../../stores/ser
...
+                        showFeedbackDiscardChangesModal={props.feedbackModal === FeedbackModal.ConfirmDiscardChangesModal}
+                        userInformation={userInformation}
+                        setUserInformation={setUserInformation}
+                        onChangeNameOrDetailsPress={onChangeNameOrDetailsPress}
+                        onRemoveThisServicePress={onRemoveThisServicePress}
+                        onOtherChangesPress={onOtherChangesPress}
+                        onFinishPress={onFinishPress}
+                        closeChooseFeedbackModeModal={closeChooseFeedbackModeModal}
+                        onDiscardPress={onDiscardPress}
+                        onKeepEditingPress={onKeepEditingPress}
```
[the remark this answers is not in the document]
bensinjin: Fixed the one line functions. There are about 6 props from the parent component that would not be applicable to this component and then a couple that come from local state ... I realize it'd be nice to shorten up this props list but I really prefer not spreading in the parent props if you're ok with that?

src/components/services/service_detail_component.tsx
```diff
@@ -32,121 +32,224 @@ import { BookmarkServiceAction, UnbookmarkServiceAction } from '../../stores/ser
...
+                        feedbackEnabled={isEditableService}
+                        feedbackDisabledComponent={<Organization name={props.service.organizationName} history={props.history} />}
+                    />
+                    <DividerComponent />
+                    <FeedbackComponent
+                        setFeedbackField={setFeedbackForField('description')}
+                        feedbackField={feedback.description}
+                        fieldLabel={<Trans>Description</Trans>}
+                        fieldValue={props.service.description}
+                        feedbackEnabled={isEditableService}
```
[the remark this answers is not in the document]

src/components/services/service_detail_component.tsx
```diff
@@ -31,122 +32,224 @@ import { BookmarkServiceAction, UnbookmarkServiceAction } from '../../stores/ser
...
 
 export const ServiceDetailComponent = (props: Props): JSX.Element => {
-    const [feedback, setFeedback]: readonly[Feedback, Dispatch<SetStateAction<Feedback>>]
-        = useState<Feedback>(getDefaultFeedbackValues());
-    const [feedbackEnabled, setFeedbackEnabled]: readonly[boolean, Dispatch<SetStateAction<boolean>>]
-        = useState<boolean>(false);
-
+    const isEditableService = props.feedbackScreen === FeedbackScreen.EditableServiceDetailPage;
     const scrollViewRef = useRef<KeyboardAwareScrollView>(undefined);
+    const [feedback, setFeedback]: readonly[ServiceFeedback, Dispatch<SetStateAction<ServiceFeedback>>] = useState(getEmptyServiceFeedback());
```
niko-pg, line 87: Just throwing this comment out

useReducer is the preferred hook for managing an object state. But the downsides of it is you'll need to create a local reducer + plus actions which may not be worth. The downside of using `useState` is it overwrites the current value rather than merging.

A clever but heretic way of achieving the classic `setState` of class components is to do:
`const [feedback, setFeedback] = useReducer((a, b) => ({ ...a, ...b  }), getEmptyServiceFeedback())`

You can now do:
`setFeedback({ name: 'new name' })` instead of
`setFeedback({ ...feedback, name: 'new name' })`

src/components/services/service_detail_component.tsx
```diff
@@ -32,121 +32,224 @@ import { BookmarkServiceAction, UnbookmarkServiceAction } from '../../stores/ser
...
+        }
+    }
 
-    const setFeedbackForField = R.curry((field: keyof Feedback, value: string): void => (
-        setFeedback({...feedback, [field]: value})
+    function sendFeedbackWhenReady(): void {
+        if (feedbackIsReadyToSend) {
+            sendFeedback();
+        }
+    }
```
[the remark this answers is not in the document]
niko-pg: +1. Moving to the sagas makes more sense now since most of the info was moved to the global state. But can punt so we dont hold this PR

src/stores/feedback/reducer.ts
tomy-pg: It appears in the mocks that when you hit submit and trigger the `ReceiveUpdatesModal`, you are already in the `ServiceDetail` screen.

src/components/services/service_detail_component.tsx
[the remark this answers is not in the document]
niko-pg: Explicit prop passing is better IMO than the prop spreading `{...props}` so it's much clearer what goes into the component. It is also good for maintainability by avoiding passing unintentional extra props. Not sure if this is something you want to consider. 

https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-props-no-spreading.md

src/components/services/service_detail_connected_component.tsx
niko-pg: How about a separate file for types? or nah?

src/selectors/feedback/to_feedback_post_data_content.ts
```diff
@@ -0,0 +1,36 @@
+import { Feedback, FeedbackField } from '../../stores/feedback/types';
+import { FeedbackPostDataContent }  from './types';
+
+export const toFeedbackPostDataContent = (feedback: Feedback, serviceId: string): FeedbackPostDataContent => {
+    switch (feedback.type) {
+        case 'service_feedback':
+            return {
+                bc211Id: serviceId,
+                name: undefinedOrFeedbackFieldValue(feedback.name),
```
niko-pg, line 9: I think this can be simplified by using optional chaining: `feedback?.name`

Though not sure if optional chaining is already configured. Ignore this comment if not

src/stores/feedback/reducer.ts
```diff
@@ -0,0 +1,64 @@
...
+        case constants.CLOSE:
+            return { ...store, modal: FeedbackModal.ConfirmDiscardChangesModal};
+        case constants.BACK:
+            return { ...store, modal:  FeedbackModal.ConfirmDiscardChangesModal};
```
niko-pg, line 56: Minor, should we just let the CLOSE case fall?
```
case constants.CLOSE:
case constants.BACK:
            return { ...store, modal:  FeedbackModal.ConfirmDiscardChangesModal};
```

src/stores/feedback/types.ts
```diff
@@ -0,0 +1,56 @@
...
+    ReceiveUpdatesModal,
+    ConfirmDiscardChangesModal,
+}
+
+export interface UserInformation {
+    readonly email: string;
+    readonly name: string;
+    readonly organizationName: string;
+    readonly jobTitle: string;
+    readonly isEmployee: undefined | boolean;
```
niko-pg, line 27: Should this be `readonly isEmployee?: boolean`

Having the isEmployee as undefined should be the same as not having it as property at all. or no?

src/components/feedback/feedback_other_remove_service_component.tsx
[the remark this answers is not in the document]
niko-pg: I think a better distinguishment is having a boundary between smart/connected components vs presentational components.

Presentational components should be aim towards clearer event callback names and must be agnostic of whether what we're passing is a redux action or just some ordinary function:
```
<FeedbackModal
    onBackPress={props.back}
    onDiscard={props.discardChanges}
/>
```
better than
```
<FeedbackModal
    back={props.back}
    discardChanges={props.discardChanges}
/>
```

src/stores/feedback/reducer.ts
bensinjin: Yeah the mocks suggest after feedback is submitted we display the receive updates modal over the service detail screen. I assumed this was going to be the flow regardless of what type of feedback you're submitting, it does feel more natural to me.

src/application/routing.ts
```diff
@@ -9,14 +8,14 @@ import * as R from 'ramda';
 // The property names of this structure are defined by the corresponding
 // route definitions, e.g. parsing a url '/learn/1' which matches
 // '/learn/:learnId' will put '1' in the attribute `learnId`.
-export interface MatchParameters {
-    readonly learnId?: LearnId;
-    readonly topicId?: TopicId;
-    readonly organizationId?: string;
-    readonly serviceId?: string;
-}
+export type RouteParameters = Partial<{
```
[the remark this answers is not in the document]
rasmus-storjohann-PG: There's room for a nice clean up here, agreed

src/components/main/main_page_switcher.tsx
```diff
@@ -38,7 +38,7 @@ export const MainPageSwitcherComponent: React.StatelessComponent<Props> = (props
         <Route exact path={routePathDefinition(Routes.Search)} component={SearchConnectedComponent} />
         <Route exact path={routePathDefinition(Routes.OrganizationDetail)} component={OrganizationDetailComponent} />
         <Route exact path={routePathDefinition(Routes.ServiceDetail)} component={ServiceDetailConnectedComponent} />
-        <Route exact path={routePathDefinition(Routes.Feedback)} component={FeedbackOtherRemoveServiceModal} />
+        <Route exact path={routePathDefinition(Routes.Feedback)} component={FeedbackOtherRemoveServiceConnectedComponent} />
```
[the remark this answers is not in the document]
rasmus-storjohann-PG: yes, I think we discussed this before

src/components/services/service_detail_component.tsx
rasmus-storjohann-PG: Clever. A few things I'd like to address here: "useReducer is the preferred hook" says who? I can see how it would be preferred in some cases, but clearly overkill in others. Not sure whether your solution is "heretical", it's pretty nice actually. I bet ramda has a function for `(a, b) => ({ ...a, ...b }` which should be used, it'll be even nicer because the code will be more readable with the function name there, I won't have to parse `(a, b) => ({ ...a, ...b }` to understand what's going on. I like good names.

src/selectors/feedback/to_feedback_post_data_content.ts
rasmus-storjohann-PG: I like this solution, if it's available.

src/stores/feedback/reducer.ts
rasmus-storjohann-PG: I think this will change once we get Cherrie to QA this, I believe there is a subtle difference in what the two buttons do

src/stores/feedback/types.ts
rasmus-storjohann-PG: The more specific we can be with the types, the more expressive the code and the fewer bugs and misunderstandings, so I will always prefer `boolean` to `boolean?`. Here I don't think there is any meaningful difference between `false` and `undefined`, so it should be just `boolean`.

src/components/services/service_detail_component.tsx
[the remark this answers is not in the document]
rasmus-storjohann-PG: OK. It's smelly to pass in so many props, maybe lower level components ought to be `connect()`ed. Leave it for now, I want to get this in.

src/components/services/service_detail_connected_component.tsx
rasmus-storjohann-PG: Not really needed. It's in the nature of connect() that it pulls in a lot of different things from different files. That's why I really like having separate files for FooComponent and FooConnectedComponent. Because connect interacts with code from different layers in the architecture, it can be helpful to use the namespace syntax to make it clear where the different symbols are coming from.

src/components/services/service_detail_component.tsx
nikojohnarellano: I based that statement above about `useReducer` from here: https://reactjs.org/docs/hooks-reference.html#functional-updates

```

Unlike the setState method found in class components, useState does not automatically merge update objects. You can replicate this behavior by combining the function updater form with object spread syntax:

setState(prevState => {
  // Object.assign would also work
  return {...prevState, ...updatedValues};
});

Another option is useReducer, which is more suited for managing state objects that contain multiple sub-values.
```

src/stores/feedback/types.ts
bensinjin: Yeah, I was thinking this should be undefined or true ... I didn't want to explicitly write that an entry wasn't from an employee when they hadn't made an explicit action.

rasmus-storjohann-PG: true/false is better

src/stores/feedback/index.ts
bensinjin: Yeah, my attempt at making the call site more clear - I agree it's overboard, I'll tidy.

src/components/services/service_detail_component.tsx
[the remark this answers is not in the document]
bensinjin: I agree! @rasmus-storjohann-PG did you want me to get this in on this PR?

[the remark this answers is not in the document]
rasmus-storjohann-PG: Yes please. You have the option of putting all the state to be sent to the server into the action, but you could instead call the selector from the saga to give you the JSON you need to transmit to the server, see which approach you like better.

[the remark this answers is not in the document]
bensinjin: Ok I'll be working on wrapping this up later this week.

src/components/feedback/feedback_other_remove_service_component.tsx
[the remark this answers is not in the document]
bensinjin: @niko-pg I can't find any useful documentation on why you wouldn't name these, my reasoning is to keep things tidy and give a hint as to what the effect does. I know this isn't conventional but I don't really understand why? Do you know?

src/components/feedback/feedback_other_remove_service_component.tsx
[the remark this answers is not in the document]
bensinjin: Good point, I think what you have here is better.

src/selectors/feedback/pick_sendable_feedback.ts
bensinjin: Fixed.

src/selectors/feedback/to_feedback_post_data_content.ts
bensinjin: Fixed (it's available)

src/stores/feedback/types.ts
bensinjin: Fixed.

src/selectors/feedback/pick_sendable_feedback.ts
bensinjin: Fixed, also made it so we don't pick the type field.

src/stores/feedback/index.ts
bensinjin: Fixed

src/selectors/feedback/pick_sendable_feedback.ts
bensinjin: I've made this clearer by using the type field.

Issue #958, opened 2020-01-02
Hide header and navigation tab on scroll


ian-lin-PG:
The header and navigation tab are always present on the search page. They should disappear when scrolling.

### Acceptance Criteria 
- [ ]  On scroll should look like this: https://zpl.io/VkzBNkK
Video demo: https://drive.google.com/open?id=1UYm9VnwCy0fEdp0WG6uRN9PuVXdS6SJn

### Steps for QA: 
1.  Go to search page
2. Type in a search term
3. Type in a search location
4. Scroll through the search results and see the page looks like the mock

ArtiBPG:
@aron-chen-pg can you add the mock here?

aron-chen-pg:
Updated the issue with Zeplin mockup and video demonstration.

niko-pg:
Hey, is the header supposed to be hidden as well when scrolling? The mock shows otherwise.

ian-lin-PG:
@niko-pg sorry for taking so long to get back to you. The header typically means the bar that includes the hamburger on the top. In the mock the hamburger was hidden before the scrolling action. I will have @aron-chen-pg to clarify the specifications.

aron-chen-pg:
@niko-pg Like Ian explained, we refer the header as the horizontal space that contains the hamburger menu and the "Find a service" header string. I have updated the video demo so you could see how the part should be hidden when scrolling. Let me know if you have any more questions.

aron-chen-pg:
Tested on Samsung 8, version 1.33.0
When clicking onto the minimized input field while the header and the bottom footer are hiding, they should resurface. Like this:
![98526207_672349580268683_3742676850111938560_n](https://user-images.githubusercontent.com/54860686/82099010-0b3fc200-96bb-11ea-9da2-0d214de7aee2.jpg)

Instead, right now the app does not re-surface the header nor the footer when input field is being clicked on:
![96841367_247454109696562_7580288085376630784_n](https://user-images.githubusercontent.com/54860686/82099054-2b6f8100-96bb-11ea-8ba6-67f536e3d66d.jpg)

tomy-pg:
Staging version 1.33.0:

After scrolling down and a search service list and seeing the header and footer disappear, clicking the back button on my Samsung s9 to navigate to a previous screen also shows the screen without a footer.

[...]

PR #1136, opened 2020-04-24
Issue #958 Hide navigation bar and header on scroll

niko-pg:


src/components/main/footer_component.tsx
```diff
@@ -1,31 +1,76 @@
...
 
 export interface FooterProps {
     readonly history: History;
     readonly location: Location;
 }
 
 export const FooterComponent: React.StatelessComponent<FooterProps> = (props: FooterProps): JSX.Element => {
+    const  { clampedScroll }: ScrollAnimationContext = useContext(ScrollContext);
+
+    const footerHeight = Platform.OS === 'android' ? FOOTER_HEIGHT : FOOTER_HEIGHT + FOOTER_SPACING;
```
tomy-pg, line 29: This logic will need to be adjusted to account for different iPhone devices. I think what you may need to do is add `FOOTER` + `FOOTER_SPACING` only to iPhone's that have 6.5 inch displays.
If you do this with all iPhones, iPhones with 5.5 inch displays will have a large white space between the end of the device screen and the footer.

src/components/main/main_component.tsx
```diff
@@ -43,26 +46,78 @@ interface MainProps {
...
+            }
+            return true;
+        };
+
+        BackHandler.addEventListener('hardwareBackPress', (): boolean => {
+            if (!props.isHeaderMenuVisible) {
+                goBack(props.history);
+            } else {
+                props.closeHeaderMenu();
+            }
```
rasmus-storjohann-PG, line 107: So this is a comment on the code as it was before this PR, so not related to your work. This is just an example of how removing a negation very often makes code clearer. Just take that if statement and flip it around, and it becomes much easier to read:
```ts
if (props.isHeaderMenuVisible) {
    props.closeHeaderMenu();
} else {
    goBack(props.history);
}
```
I know you must think I'm picky, but I believe that it's thousands of little things like this that makes the difference between code that is just OK and code that is really good and maintainable.

src/components/main/main_component.tsx
```diff
@@ -43,26 +46,78 @@ interface MainProps {
...
+            return true;
+        });
+
+        return (): void => {
+            BackHandler.removeEventListener('hardwareBackPress', onHardwareBackButtonPress);
+        };
+    }, []);
```
rasmus-storjohann-PG, line 114: I would suggest using functions like this to make the logic clearer
```ts
const addEventsListener = () => {
   ...
};

const removeEventListener = () => {
   ...
};

addEventListener();
return removeEventListener;
```

src/components/main/main_component.tsx
```diff
@@ -81,17 +136,18 @@ export const MainComponent = (props: Props): JSX.Element => {
                     />
                 }
             >
-                <Container>
+                <Animated.View style={{ flex: 1  }}>
+                    <ScrollContext.Provider value={scrollAnimationContext}>
```
rasmus-storjohann-PG, line 140: I'm not a big fan of the context API, whether with hooks or without, since they're a way to disguise the fact that there are effectively global variables. However, in this case I can see how they've valid since they are used specifically for global state related to the animation.

src/components/main/main_component.tsx
```diff
@@ -1,23 +1,26 @@
...
+import { Drawer, Root } from 'native-base';
+import React, { useRef, useEffect, EffectCallback } from 'react';
+import { BackHandler, StatusBar } from 'react-native';
+import Animated from 'react-native-reanimated';
+
 import { RouterProps, goBack } from '../../application/routing';
-import { Location, Action } from 'history';
-import { RouteChangedAction } from '../../stores/router_actions';
 import { Locale } from '../../locale';
-import { HardwareBackButtonHandlerComponent } from './hardware_back_button_handler_component';
```
rasmus-storjohann-PG, line 11: I would have expected this file to be removed in this PR, since you're not needing it anymore

src/components/main/main_component.tsx
```diff
@@ -43,26 +46,78 @@ interface MainProps {
...
 
-    const onHardwareBackButtonPress = (): boolean => {
-        if (!props.isHeaderMenuVisible) {
-            goBack(props.history);
-        } else {
-            props.closeHeaderMenu();
-        }
-        return true;
+    // Apply hardware back button press listener
+    useEffect((): EffectCallback => {
```
rasmus-storjohann-PG, line 92: I would like to see this hook moved to its own file

src/components/search/search_component.tsx
```diff
@@ -1,28 +1,37 @@
...
+import { renderHeader } from '../main/render_header';
+
+import { fetchSearchResultsFromQuery } from './api/fetch_search_results_from_query';
+import { fetchLatLongFromLocation } from './api/fetch_lat_long_from_location';
+import { SearchInputComponent } from './search_input_component';
+import { SearchResultsComponent } from './search_results_component';
+import { useOnlineStatus } from './use_online_status';
+
+// Based from inspecting React Native's Element inspector
+const HEADER_HEIGHT: number = 44;
```
rasmus-storjohann-PG, line 34: Should this constant be part of the styles file?

src/components/search/search_component.tsx
```diff
@@ -88,7 +97,12 @@ export const SearchComponent = (props: Props): JSX.Element => {
 
     const onLoadMore = async (): Promise<void> => {
         try {
-            const moreResults = await fetchSearchResultsFromQuery(props.searchTerm, props.searchPage + 1, props.searchLatLong, props.saveNumberOfSearchPages);
+            const moreResults = await fetchSearchResultsFromQuery(
+                props.searchTerm,
+                props.searchPage + 1,
+                props.searchLatLong,
+                props.saveNumberOfSearchPages,
+            );
```
rasmus-storjohann-PG, line 105: Again, avoid cleanup in commits with complex change in functionality

src/components/search/search_component.tsx
```diff
@@ -132,10 +146,23 @@ const useDisableAnalyticsOnEasterEgg = (location: string, disableAnalytics: (dis
 };
 
 const Header = (props: MenuAndBackButtonHeaderProps): JSX.Element => {
+    const { clampedScroll }: ScrollAnimationContext = useContext(ScrollContext);
+
+    const headerTranslate = Animated.interpolate(clampedScroll, {
+        inputRange: [0, (HEADER_HEIGHT + Constants.statusBarHeight)],
+        outputRange: [HEADER_HEIGHT + Constants.statusBarHeight, Constants.statusBarHeight],
+        extrapolate: Animated.Extrapolate.CLAMP,
+    });
```
rasmus-storjohann-PG, line 155: `headerTranslate` is not a good name. Further down the code `style={{ height: headerTranslate }}` gives a hint to a better name, this is not a "translate", it'a a "height", so maybe "animiatedHeaderHeight"?

src/components/search/search_component.tsx
```diff
@@ -132,10 +146,23 @@ const useDisableAnalyticsOnEasterEgg = (location: string, disableAnalytics: (dis
...
             onPress={props.openHeaderMenu}
             textColor={props.textColor}
         />;
-    return renderHeader({ ...props, rightButtons: [rightButton] });
+
+    return <Animated.View style={{ height: headerTranslate }}>
+        {
+            renderHeader({ ...props, rightButtons: [rightButton] })
+        }
+    </Animated.View>;
```
rasmus-storjohann-PG, line 167: This is purely aesthetic, but I think it's usually possible to change `renderHeader({ ...props, rightButtons: [rightButton] })` into `<RenderHeader...>`, which would just look better. It can be quite a lot of work though, so it's totally optional.

src/components/search/search_results_component.tsx
```diff
@@ -67,33 +73,45 @@ export const SearchResultsComponent = (props: Props): JSX.Element => {
 };
 
 const renderComponentWithResults = (props: Props): JSX.Element => {
-    const flatListRef = useRef<FlatList<SearchServiceData>>(undefined);
+    // $FIXME Animated.createAnimatedComponent only has any as its return type
+    const flatListRef = useRef<any>(undefined);
```
rasmus-storjohann-PG, line 77: Fix this. If the third party library forces you to use any, that's life.

src/components/search/search_results_component.tsx
```diff
@@ -67,33 +73,45 @@ export const SearchResultsComponent = (props: Props): JSX.Element => {
...
+            flatListRef.current.getNode().scrollToOffset({ animated: false, offset: props.searchOffset });
         }
     }, [props.searchOffset]);
+
     return (
-        <View style={{ flexDirection: 'column', backgroundColor: colors.lightGrey, flex: 1 }}>
+        <Animated.View style={{ flexDirection: 'column', backgroundColor: colors.lightGrey, flex: 1 }}>
             {renderLoadingScreen(props.isLoading)}
-            <FlatList
+            <AnimatedFlatList
```
rasmus-storjohann-PG, line 96: Is this part of the same change as the animation of the header and footer? I'm confused, separate commits would have made this clearer.

src/components/search/search_results_component.tsx
```diff
@@ -67,33 +73,45 @@ export const SearchResultsComponent = (props: Props): JSX.Element => {
...
                 ref={flatListRef}
-                onScroll={(e: any): void => props.setScrollOffset(e.nativeEvent.contentOffset.y)}
+                onScroll={Animated.event(
+                    [{nativeEvent: {contentOffset: { y: scrollAnim }}}],
+                    {
+                        listener: ((e: any): void => props.setScrollOffset(e.nativeEvent.contentOffset.y)),
+                        useNativeDriver: true,
+                    },
+                )}
```
rasmus-storjohann-PG, line 104: Need to break this apart with some helper functions. See, we always use names of functions and variables to communicate what the code is doing. With this inline style of programming, you just have way fewer names, meaning you have fewer means of expressing what is going on, and it makes the code difficult to understand.

src/components/main/footer_component.tsx
```diff
@@ -1,31 +1,76 @@
...
+
+    // Animate Footer Height
```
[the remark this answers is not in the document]
niko-pg: Fixed

src/components/main/main_component.tsx
```diff
@@ -1,23 +1,26 @@
...
-import { Container, Drawer, Root } from 'native-base';
-import { MainPageSwitcherComponent } from './main_page_switcher';
-import { FooterComponent, FooterProps } from './footer_component';
-import { HeaderMenuConnectedComponent } from '../header_menu/header_menu_connected_component';
+import { Action, History, Location } from 'history';
+import { Notifications } from 'expo';
+import { Drawer, Root } from 'native-base';
+import React, { useRef, useEffect, EffectCallback } from 'react';
+import { BackHandler, StatusBar } from 'react-native';
+import Animated from 'react-native-reanimated';
```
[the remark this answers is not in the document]
niko-pg: Yeah cleaning up of imports is somewhat an instinct of mine ever since. Even arranging in alphabetical order 

I decided to revert all the import cleanups and would like to propose to install this plugin: 
https://github.com/webschik/tslint-import-rules#readme
 
This should be on a separate review if you will permit. I'd expect the plugin to automatically fix all imports throughout the codebase

src/components/main/footer_component.tsx
niko-pg: Solved the problem by only specifying the `footer spacing` for iPhoneX. Based it off when looking at the source code inside node_modules and saw how Native Base handles the Footer component.

Footer's render method:
```
  render() {
    const { style } = this.props;
    const variables = this.context.theme
      ? this.context.theme['@@shoutem.theme/themeStyle'].variables
      : variable;
    return variables.isIphoneX ? (
      <View
        ref={c => (this._root = c)}
        {...this.props}
        onLayout={e => this.layoutChange(e.nativeEvent.layout)}
        style={[
          style,
          {
            height: this.calculateHeight(
              this.state.orientation,
              variables.Inset
            ),
            paddingBottom: this.calculatePadder(
              this.state.orientation,
              variables.Inset
            )
          }
        ]}
      />
    ) : (
      <View ref={c => (this._root = c)} {...this.props} />
    );
  }
```

src/components/main/footer_component.tsx
```diff
@@ -1,31 +1,76 @@
...
+    const footerTranslatedHeight = Animated.interpolate(clampedScroll, {
+        inputRange: [0, footerHeight],
+        outputRange: [footerHeight, 0],
+        extrapolate: Animated.Extrapolate.CLAMP,
+    });
+
+    // Animate Footer Tab Height
+    const footerTabTranslatedHeight = Animated.interpolate(clampedScroll, {
+        // Only translate the footer tab height once the extra spacing of the footer container is gone
+        inputRange: [FOOTER_HEIGHT, footerHeight],
```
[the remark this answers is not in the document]
niko-pg: Fixed.

src/components/main/main_component.tsx
```diff
@@ -43,26 +46,78 @@ interface MainProps {
...
+export interface ScrollAnimationContext {
+    readonly scrollAnim: Animated.Value<number>;
+    readonly offsetAnim: Animated.Value<number>;
+    readonly clampedScroll: Animated.Node<number>;
+}
+
+export const ScrollContext = React.createContext<ScrollAnimationContext | undefined>(undefined);
+
 export const MainComponent = (props: Props): JSX.Element => {
+    const scrollAnim = useRef<Animated.Value<number>>(new Animated.Value(0));
```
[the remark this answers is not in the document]
niko-pg: Fixed

src/components/main/main_component.tsx
```diff
@@ -43,26 +46,78 @@ interface MainProps {
...
-        return true;
+    // Apply hardware back button press listener
+    useEffect((): EffectCallback => {
+        const onHardwareBackButtonPress = (): boolean => {
+            if (!props.isHeaderMenuVisible) {
+                goBack(props.history);
+            } else {
+                props.closeHeaderMenu();
+            }
+            return true;
```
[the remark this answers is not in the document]
niko-pg: Fixed. This is a typo. Not intended to be repeated

src/components/main/main_component.tsx
niko-pg: Fixed.

src/components/main/main_component.tsx
niko-pg: Fixed.

src/components/main/main_component.tsx
niko-pg: Fixed.

src/components/main/main_component.tsx
niko-pg: Fixed.

src/components/search/search_component.tsx
niko-pg: Fixed.

src/components/search/search_results_component.tsx
niko-pg: Fixed.

src/components/search/search_results_component.tsx
niko-pg: Yes, this is part of the change. `FlatList` needs to be converted to `AnimatedFlatList` so we can capture the event and map to `scrollAnimatedValue`

src/components/search/search_component.tsx
niko-pg: Fixed

src/components/search/search_results_component.tsx
niko-pg: Fixed. Broke into separate variables

src/components/main/footer_component.tsx
```diff
@@ -63,10 +81,26 @@ export const FooterComponent: React.StatelessComponent<FooterProps> = (props: Fo
...
+                <NavigationButton
+                    icon='bookmark'
+                    isActive={isOnBookmarksPage(props)}
+                    onPress={goToRouteWithoutParameter(Routes.Bookmarks, props.history)}
+                />
+                 <NavigationButton
+                    icon='search'
+                    isActive={isOnSearchPage(props)}
+                    onPress={goToRouteWithoutParameter(Routes.Search, props.history)}
+                />
```
rasmus-storjohann-PG, line 73: I like this change

src/components/main/footer_component.tsx
```diff
@@ -39,6 +44,19 @@ const getFooterStyles = (): FooterStyles => {
...
+        <Button vertical onPress={props.onPress} style={{ flexWrap: 'nowrap' }}>
+            <Icon
+                type='FontAwesome'
+                name={props.icon}
+                active={props.isActive}
+                style={{ fontSize: values.navigationIconSize,  color: props.isActive ? colors.white : colors.teal }}
+            />
+        </Button>
+    );
+};
```
rasmus-storjohann-PG, line 42: Picky style thing, just a reminder that we prefer to drop the return when possible, i.e. prefer this
```ts
const NavigationButton = (props: NavigationButtonProps): JSX.Element => (
   <Button vertical onPress={props.onPress}...
```
over
```ts
const NavigationButton = (props: NavigationButtonProps): JSX.Element => {
    return (
         <Button vertical onPress={props.onPress}...
```

src/components/main/main_component.tsx
```diff
@@ -44,6 +45,18 @@ interface MainProps {
...
+    useHardwareBackButtonPress((): boolean => {
+        const shouldNotBubbleUpEvent = true;
+
+        if (props.isHeaderMenuVisible) {
+            props.closeHeaderMenu();
+        } else {
+            goBack(props.history);
+        }
+
+        return shouldNotBubbleUpEvent;
```
rasmus-storjohann-PG, line 57: I'm not clear on what this does ... what's the code that pays attention to this return value?

src/components/search/search_component.tsx
```diff
@@ -138,21 +134,15 @@ const useDisableAnalyticsOnEasterEgg = (location: string, disableAnalytics: (dis
...
+    const { headerAnimatedHeight }: ScrollAnimationContext = useContext(ScrollContext) as ScrollAnimationContext;
 
     const rightButton =
         <MenuButtonComponent
             onPress={props.openHeaderMenu}
             textColor={props.textColor}
         />;
 
-    return <Animated.View style={{ height: headerTranslate }}>
+    return <Animated.View style={{ height: headerAnimatedHeight }}>
```
rasmus-storjohann-PG: You don't have to change this, but for naming, just write the way you would talk. If you were talking about this code with another human being, you'd never say "header animated height", you'd say "animated header height", right? So write that (next time). @bensinjin this is something you can work on also.

src/components/main/scroll_animation_context.tsx
```diff
@@ -79,11 +79,26 @@ export const createScrollAnimationContext = (): ScrollAnimationContext => {
...
+                        y: scrollAnimatedValueRef.current,
+                    },
+                },
+            }],
+            {
+                listener: onScroll,
+                useNativeDriver: true,
+            },
+        );
+    };
```
rasmus-storjohann-PG, line 110: That took some digging to figure out, I'm sure!

src/components/main/use_hardware_back_button_press.ts
```diff
@@ -1,11 +1,20 @@
...
     useEffect((): EffectCallback => {
-        // tslint:disable-next-line: no-expression-statement
-        const backHandler = BackHandler.addEventListener('hardwareBackPress', onHardwareBackPress);
+        const addListener = (): void => {
+            BackHandler.addEventListener('hardwareBackPress', onHardwareBackPress);
+        };
+
+        const removeListener = (): void => {
+            BackHandler.removeEventListener('hardwareBackPress', onHardwareBackPress);
+        };
```
rasmus-storjohann-PG, line 14: If you find this verbose, this might feel lighter:
```ts
const addListener = (): void => BackHandler.addEventListener('hardwareBackPress', onHardwareBackPress);
const removeListener = (): void => BackHandler.removeEventListener('hardwareBackPress', onHardwareBackPress);
...
```

src/components/main/header_component.tsx
```diff
@@ -5,14 +5,14 @@ import { EmptyComponent } from '../empty_component/empty_component';
...
-interface RenderHeaderProps {
+interface HeaderComponent {
     readonly backgroundColor: string;
     readonly leftButton?: JSX.Element;
     readonly rightButtons: ReadonlyArray<JSX.Element>;
     readonly title?: JSX.Element;
 }
 
-export const renderHeader = (props: RenderHeaderProps): JSX.Element => {
+export const HeaderComponent = (props: HeaderComponent): JSX.Element => {
```
rasmus-storjohann-PG, line 15: This must be a typo, surprised the compiler didn't complain.

src/components/main/footer_component.tsx
rasmus-storjohann-PG: You don't need to fix this here, just for the future.

rasmus-storjohann-PG:
Rebase is needed

rasmus-storjohann-PG:
Oh and some of the commit messages aren't right

src/components/main/main_component.tsx
```diff
@@ -43,26 +46,78 @@ interface MainProps {
 
 type Props = MainComponentProps & MainComponentActions;
 
+export interface ScrollAnimationContext {
+    readonly scrollAnim: Animated.Value<number>;
+    readonly offsetAnim: Animated.Value<number>;
+    readonly clampedScroll: Animated.Node<number>;
+}
+
+export const ScrollContext = React.createContext<ScrollAnimationContext | undefined>(undefined);
```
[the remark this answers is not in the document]
niko-pg: Fixed

src/components/main/main_component.tsx
niko-pg: From the react native docs (https://reactnative.dev/docs/backhandler):

```
The event subscriptions are called in reverse order (i.e. the last registered subscription is called first).

- If one subscription returns true, then subscriptions registered earlier will not be called.
- If no subscription returns true or none are registered, it programmatically invokes the default back button functionality to exit the app.
```
In terms of why we return true I'm not sure. The handler has always been like this, maybe it was decided that we dont want the default behaviour of the back button?

src/components/main/header_component.tsx
niko-pg: Fixed

src/components/search/search_component.tsx
niko-pg: Great point. Will work on this in the future!

src/components/main/use_hardware_back_button_press.ts
niko-pg: Fixed

niko-pg:
Thanks for the review @rasmus-storjohann-PG ! Pushed the last round of code review fixes plus rebased and reworded the wrong commit messages

rasmus-storjohann-PG:
OK, I'll have a look at this PR later today.

src/components/main/main_component.tsx
rasmus-storjohann-PG: Thanks for the info, lets leave existing behaviour unchanged

rasmus-storjohann-PG:
Rebase needed, and UX glitch seen on Android, so we're close but not quite there yet.

rasmus-storjohann-PG:
Hey @niko-pg looks like there's a rebase needed, looks tricky so I didn't attempt it.

Issue #1131, opened 2020-04-22
Bottom navigation bar gets pushed above keyboard


aron-chen-pg:
When clicking on to a text input field, the bottom navigation bar, which was supposed to be covered underneath the keyboard would get pushed above instead. This happens throughout the app, including Search and Topic Services. The issue was tested for both iOS (iPhones) and Android (Samsung 8) and was only found on Android. 

Below is the video demonstration for staging version 1.29.1 on Android:
![Bottom navigation behaviour.gif](https://images.zenhubusercontent.com/5d6ea767bc3e240001df648b/e5ec13f2-9b7e-446a-a3a8-f286b8c5d453)

### Acceptance Criteria 
- [ ]  When a text input field is clicked/focused, the bottom nav is covered underneath the keyboard.

### Steps for QA: 
1.  Go to Search.
2.  Click on any text input field.

tomy-pg:
Staging 1.32.0 - passes

PR #1139, opened 2020-04-29
Issue #1131 Hide footer if Keyboard opens

ian-lin-PG:


src/components/main/footer_component.tsx
```diff
@@ -13,7 +14,27 @@ export interface FooterProps {
 
 export const FooterComponent: React.StatelessComponent<FooterProps> = (props: FooterProps): JSX.Element => {
 
-    if (isFooterHidden(props)) {
+    const [keyboardIsVisible, setKeyboardIsVisible]: readonly [boolean, Dispatch<SetStateAction<boolean>>] = useState(false);
+
+    React.useEffect((): () => void => {
```
tomy-pg, line 19: If you import useEffect like you do with useState, prefixing React. before useEffect is no longer needed.

ian-lin-PG: Good catch, I was lazy and didn't import properly.

src/components/main/footer_component.tsx
```diff
@@ -13,7 +14,27 @@ export interface FooterProps {
...
+        Keyboard.addListener('keyboardDidShow', keyboardOpens);
+        Keyboard.addListener('keyboardDidHide', keyboardCloses);
+
+        return (): void => {
+            Keyboard.removeListener('keyboardDidShow', keyboardOpens);
+            Keyboard.removeListener('keyboardDidHide', keyboardCloses);
+        };
+    }, [keyboardIsVisible]);
+
+    const keyboardOpens = (): void => {
```
bensinjin, line 29: It kinda feels like the listeners to the keyboard even should live even further up the component tree. Maybe in the `main_component`? Then you could set Redux state when it opens and closes and just `selectKeyboardIsOpen()` in child components like the footer.  Also, I think the dependency for the effect is just: `[]` because you only want to register the listeners when the app mounts then then clean up when it unmounts.

rasmus-storjohann-PG: OK, so I would like to establish some conventions for `useEffect()`. @niko-pg you'll be interested in this too, we talked about this at length.

First, for this kind of register/unregister flow,  which I think we will see quite a lot, I would like to see it spelled out like this, the names of the two opposite functions will make it nice and clear what's going on.
```ts
useEffect(() => {
    const addListener = () => {
        ...
    };
    const removeListener = () => {
        ...
    };
    addListener();
    return removeListener();
});
```
Secondly, `useEffect()` has the potential of introducing subtle bugs related to calls from inside the useEffect() to code outside it. To avoid these kinds of bugs, the rule will be that anything that is (directly or indirectly) accessed outside of the `useEffect()` needs to be included in the array passed as the second argument to `useEffect()`. It should be clear that we want that to not get out of hand, with very long lists of things passed in. The solution then is to move helper functions inside the `useEffect()`. So, in this case, it would mean moving keyboardOpens and keyboardCloses inside the `useEffect()`. That will make this `useEffect()` dependent on just `[setKeyboardIsVisible]`.

I prefer so see the listener here where it is needed, rather than higher in the component hierarchy and stored in redux. In this case, the flag is just needed here so a local state management solution is most natural.

bensinjin: Your state point makes sense but I guess I don't understand why we want to re-register the listeners whenever the boolean changes?

niko-pg: +1 on defining functions inside `useEffect` if it's only used within `useEffect`. 

And just want to clarify the proposed convention above. If converting Ian's code above to the sample code. Would it be:
```
useEffect(() => {
     const keyboardOpens = (): void => {
        ....
     }

     const keyboardCloses = (): void => {
        ....
     }

    const addListener = () => {
        Keyboard.addListener('keyboardDidShow', keyboardOpens);
        Keyboard.addListener('keyboardDidHide', keyboardCloses);
    };
    const removeListener = () => {
          Keyboard.removeListener('keyboardDidShow', keyboardOpens);
          Keyboard.removeListener('keyboardDidHide', keyboardCloses);
    };
    addListener();
    return removeListener;
}, [);
```

Not opposing to this approach but seems verbose to be put in inside a component. Should we abstract further into our own custom hook `useListeners`  to enforce this clarity?

niko-pg: and  +1 on Ben's comment. `keyboardIsVisible` shouldn't be a dependency

bensinjin: + 1 to functions in effects

rasmus-storjohann-PG: @niko-pg I agree that it's verbose, but the current style has no structure to it at all, 
```ts
        Keyboard.addListener('keyboardDidShow', keyboardOpens);
        Keyboard.addListener('keyboardDidHide', keyboardCloses);

        return (): void => {
            Keyboard.removeListener('keyboardDidShow', keyboardOpens);
            Keyboard.removeListener('keyboardDidHide', keyboardCloses);
        };
```
it doesn't communicate the very important fact that these two blocks of code do the opposite thing, and it'll cause us confusion bugs going forward.

ian-lin-PG: Hi all new commit pushed here is a few noteworthy things:
1) This bug is related to these 2 bugs tracked on the Expo github page [this](https://github.com/expo/expo/issues/7589) and [that](https://github.com/expo/expo/issues/7815). There's no progress on the issue but remains open. [This comment](https://github.com/expo/expo/issues/7815#issuecomment-619906901) sums up the issue of `KeyboardAvoidingView` behaving inconsistently on the android platform. Ideally Expo would fix this issue before the next production release, which circumvents this PR, but hard to say.

2) There is an UI bug when running consecutive searches. The second search onwards, there's a white element about the size of the keyboard that covers the loading screen. On Tomy's phone the spinner is covered, presumably due to the dimension of screen vs keyboard on his phone. See the following image
Screenshot_20200501-151949


3) Registering `[]` or `[keyboardIsVisible]` doesn't seem to affect the functionality at all. My original thinking is that the side effect of hiding or rendering the footer is only contingent on the state of the keyboard. In this case, the `FooterComponent` only has the one state, and both the `useEffect` and `useState` are mounted at the same time. If we have a totally unrelated state or prop here... I will have to think a lot more about if the `useEffect` should be `[]` or subscribe to a few of the states.

4) Otherwise the rest of the changes in this thread are in the new commit.

rasmus-storjohann-PG: It's `setKeyboardIsVisible` that needs to be registered. The reasons are subtle, we should do a lunch&learn.

Issue #1140, opened 2020-04-29
Geocoder is throttling results


rasmus-storjohann-PG:
Geocoder is throttling results leading to very slow UX for search. Client side caching of commonly used location strings could make this a lot better. 

Acceptance criteria:
- [ ] When repeatedly searching with the same location string (a very common use case it seems to me), the geocoder request is only sent out once. Cache size of 1 could really make a difference here, or we can build something little more powerful.

tomy-pg:
Staging version 1.33.0: PASS

PR #1147, opened 2020-04-30
Issue #1140 Don't refetch the same location

bensinjin:
The only change needed here was to not refetch the location if it hadn't changed since the previous search for the topic services screen. This was already being done for the search screen. Both these values are already cached in Redux so there was no need to build anything.

### DIFF
--- a/src/components/services/service_list_location_search_component.tsx
+++ b/src/components/services/service_list_location_search_component.tsx
@@ -123,6 +123,7 @@ const ExpandedSearch = (props: ExpandedSearchProps): JSX.Element => {
                         props.locationInputValue,
                         props.setManualUserLocation,
                         props.setSearchIsCollapsed,
+                        props.locationInputValue === props.manualUserLocation.label,
                     )}
                     isDisabled={searchIsDisabled}
                 />
@@ -233,8 +234,12 @@ const getSearchOnPress = (
     locationInputValue: string,
     setManualUserLocation: Props['setManualUserLocation'],
     setSearchIsCollapsed: (b: boolean) => void,
+    locationIsUnchanged: boolean,
 ): () => void => (): void => {
     setSearchIsCollapsed(true);
+    if (locationIsUnchanged) {
+        return;
+    }
     if (locationInputValue === MY_LOCATION) {
         fetchLatLng(
             setIsFetchingLatLng,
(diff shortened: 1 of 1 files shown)


src/components/services/service_list_location_search_component.tsx
```diff
@@ -123,6 +123,7 @@ const ExpandedSearch = (props: ExpandedSearchProps): JSX.Element => {
                         props.locationInputValue,
                         props.setManualUserLocation,
                         props.setSearchIsCollapsed,
+                        props.locationInputValue === props.manualUserLocation.label,
```
rasmus-storjohann-PG, line 126: So this code is already pretty smelly and it just got a bit worse... how about this more declarative style:
```ts
const isCachedLocationValid = props.locationInputValue === props.manualUserLocation.label;
const doNothing = () => {};
const lookupLocationLatLong = () => {getSearchOnPress(
                        props.setIsFetchingLatLng,
                        props.locationInputValue,
                        props.setManualUserLocation,
                        props.setSearchIsCollapsed,
                    )}};
const onPress = isCachedLocationValid ? doNothing : lookupLocationLatLong;
 ...
<SearchButton
                    onPress={onPress} ...
```

bensinjin: Yeah, ok, I do think we want to collapse the search so it acts like the other search so I named `doNothing` `collapseSearch` - other than that followed what's here, thx!

src/components/services/service_list_location_search_component.tsx
```diff
@@ -98,6 +98,16 @@ const ExpandedSearch = (props: ExpandedSearchProps): JSX.Element => {
...
     const searchIsDisabled = props.locationInputValue.length === 0 || props.isFetchingLatLng;
+    const isCachedLocationValid = props.locationInputValue === props.manualUserLocation.label;
+    const collapseSearch = (): void => { props.setSearchIsCollapsed(true); };
+    const lookupLocationLatLong = getSearchOnPress(
+        props.setIsFetchingLatLng,
+        props.locationInputValue,
+        props.setManualUserLocation,
+        props.setSearchIsCollapsed,
+    );
+    const onSearchButtonPress = isCachedLocationValid ? collapseSearch : lookupLocationLatLong;
```
rasmus-storjohann-PG, line 109: Better, now the code is telling a story!

Issue #1152, opened 2020-05-04
Notification token language does not update immediately


ian-lin-PG:
When a user changes their language in app, their notification token should change to the new language. This allows us to send notification in the language that matches the app.

### Acceptance Criteria 
- [ ]  Token language changes when in app language changes

### Steps for QA: 
1.  Open the app in English
2. Connect to staging DB through postico
3. Open the notification table
4. Change the app language to Arabic
5. Find the token entry that changed from English to Arabic
6. Change app language to another non Arabic language (Korean in this example)
7. Check the same token entry changed to Korean
8. Change app language to yet another non Arabic language (Tagalog)
9. Check token entry updated to Tagalog, THIS IS MOST IMPORTANT TO PASS QA.

tomy-pg:
Not seeing notification token for my Samsung S9.

tomy-pg:
Staging version 1.33.0: PASS

PR #1153, opened 2020-05-05
Issue #1152 Add updateNotificationToken to onPress event

ian-lin-PG:


src/components/header_menu/header_menu_connected_component.ts
```diff
@@ -6,6 +6,7 @@ import { SaveLocaleRequestAction, saveLocaleRequest } from '../../stores/locale/
 import { LocaleInfo } from '../../locale';
 import { selectAvailableLocales } from '../../selectors/locale/select_available_locales';
 import { selectLocale } from '../../selectors/locale/select_locale';
+import { PushNotificationPostRequestAction, request } from '../../sagas/post_push_notification_token';
```
tomy-pg, line 9: could you rename / refactor some of the names in this saga file? `request` is a bit difficult to follow

src/components/header_menu/header_menu_component.tsx
```diff
@@ -1,3 +1,4 @@
+// tslint:disable: no-expression-statement
```
rasmus-storjohann-PG, line 1: I prefer these on the line that breaks the linter rule rather than for the whole file, unless the rule is broken in many places in the file.

src/application/store.ts
```diff
@@ -7,7 +7,7 @@ import { runSaga, ApplicationSaga } from '../sagas';
 import { loadFontsActions } from '../stores/fonts';
 import { loadLocaleRequest } from '../stores/locale/actions';
 import { DataPersistence } from '../stores/persisted_data';
-import { request as requestPushNotificationPost } from '../sagas/post_push_notification_token';
+import { pushNotificationTokenRequest as requestPushNotificationPost } from '../sagas/post_push_notification_token';
```
rasmus-storjohann-PG, line 10: Hmmm, this became a little awkward now. We should try to avoid having more than one name for the same thing, so I suggest you change one of these names to the other. I like `pushNotificationTokenRequest` better of the two.

src/components/header_menu/header_menu_connected_component.ts
```diff
@@ -6,7 +6,7 @@ import { SaveLocaleRequestAction, saveLocaleRequest } from '../../stores/locale/
 import { LocaleInfo } from '../../locale';
 import { selectAvailableLocales } from '../../selectors/locale/select_available_locales';
 import { selectLocale } from '../../selectors/locale/select_locale';
-import { PushNotificationPostRequestAction, request } from '../../sagas/post_push_notification_token';
+import { PushNotificationPostRequestAction, pushNotificationTokenRequest } from '../../sagas/post_push_notification_token';
```
rasmus-storjohann-PG, line 9: Here it would also be nice to have names that require the reader to not have to think. I think we can make it easier to read by sticking to a consistent `fooBar()` and `FooBarAction` for the action creator and action type.

PR #1155, opened 2020-05-07
Feature/issue 1057 partial localization

tomy-pg:


src/components/services/service_list_location_search_component.tsx
```diff
@@ -261,8 +264,12 @@ const getSearchOnPress = (
     }
 };
 
+const locationInputIsMyLocation = (locationInput: string, i18n: I18n): boolean => (
+    locationInput === MY_LOCATION || isLocalizedMyLocation(locationInput, i18n)
+);
```
rasmus-storjohann-PG: You have three functions that do almost the same thing ...

src/api/fetch_lat_long_from_location.ts
```diff
@@ -4,28 +4,40 @@ import { LatLong } from '../validation/latlong/types';
...
+    location === MY_LOCATION
+);
+
+const locationIsNotEmpty = (location: string): boolean => (
+    location !== ''
+);
+
+const userIsOnline = (onlineStatus: OnlineStatus): boolean => (
+    onlineStatus === OnlineStatus.Online
+);
```
rasmus-storjohann-PG: Nice

src/components/services/service_list_location_search_component.tsx
```diff
@@ -239,78 +238,24 @@ const LocateButton = (props: LocateButtonProps): JSX.Element => (
...
-    const latLong = await fetchFn().catch(handleFetchError);
-    setIsFetchingLatLng(false);
+    const latLong = await fetchLatLongFromLocation(location);
     setManualUserLocation({
-        label: locationLabel,
-        latLong,
+        label: locationInputValue,
+        latLong: latLong,
     });
+    setIsFetchingLatLng(false);
```
rasmus-storjohann-PG, line 274: Looking for improved error handling later in this PR...

src/components/services/service_list_location_search_component.tsx
```diff
@@ -146,6 +139,31 @@ const ExpandedSearch = (props: ExpandedSearchProps): JSX.Element => {
...
+): JSX.Element => {
+    if (!hasMyLocation(locationInputValue, i18n)) {
+        return (
+            <LocateButton
+            onPress={getLocateOnPress(setLocationInputValue, i18n)}
+            isFetchingLatLng={isFetchingLatLng}
+            />
+        );
+    }
+    return <EmptyComponent />;
```
rasmus-storjohann-PG, line 156: Picky: I'd flip this if statement around, two benefits: one less negation, and early return on the "don't need to do anything" branch.

src/components/services/service_list_location_search_component.tsx
rasmus-storjohann-PG: So the only thing I would really ask you to fix in this PR is here. Before you call `fetchLatLongFromLocation()` you call `setIsFetchingLatLng()` with true, and after it comes back you set it back to false. But what if `fetchLatLongFromLocation()` throws, then the flag never gets set back. What we need here is some form of a `finally` control flow. Can you make that work? Other than that, the other comments are optional and this thing is ready to go.

tomy-pg: Fixed.

src/components/services/service_list_location_search_component.tsx
tomy-pg: Fixed. I think this makes it easier to understand as well.

PR #1162, opened 2020-05-13
Feature/issue 1152 notifications

tomy-pg:
Solution here: https://github.com/expo/expo/issues/7606#issuecomment-608446477


bin/prepare_deploy.sh
```diff
@@ -316,14 +316,22 @@ completeManualConfiguration() {
...
+    echo "Set the ALGOLIA_SEARCH_API_KEY available from "
     echo "https://www.algolia.com/apps/MMYH1Z0D3O/api-keys/all"
     echo
+    echo "Set the Airtable API_KEY, BASE_ID and TABLE_ID. "
+    echo "Retrieve the API_KEY from https://airtable.com/account and "
+    echo "the BASE_ID and TABLE_ID from https://airtable.com/api"
+    echo
+    echo "Add the google-services.json file to the root directory of $CLIENT_DIRECTORY. "
+    echo "Retrieve this by going to https://console.firebase.google.com/u/0/ , "
+    echo "selecting ArrivalAdvisor $BUILD and navigating to project settings."
```
rasmus-storjohann-PG, line 333: OK, so this is getting to be pretty cumbersome now. 1Pass has a "document" data type, which is just a file. So I suggest we make a file that contains all of these values and stick it in 1Pass in a vault that Ian, Tomy and I have access to. We need to use a format like CSV or JSON so that we can read/parse the file from the deploy scripts and insert the various values where they need to go.

ian-lin-PG: > Just the one suggestion, @ian-lin-PG do you think that solves the problems you've been highlighting in the deployment process?

This is definitely a step in the right direction. I also recommend writing the variable names into the wording directory `.env` files. We don't need to necessarily write the value of those variables, more so as a reminder that something is missing and need to be filled in for the deployment to continue successfully.

tomy-pg: In the `prepare_deploy` script we could pull in the `env` variables using a file as suggested by @rasmus-storjohann-PG. We would set each of these as arguments IE `$AIRTABLE_API_KEY` and then have the script also check if these values are missing as per @ian-lin-PG 's suggestion.

Issue #1106, opened 2020-04-06
Navigating from Service Detail to Service List/My Bookmarks should not bring users back to top of page


jparlina:
As noted by @aron-chen-pg when testing bookmarks on issue #1046:
When a user selects a service record to get the service details from service listing under a topic or My Bookmarks, and then back to the service listing under a topic/My Bookmarks, the user should be brought back to the position of the service that they just clicked on. Users should not be brought back to the top of the page. This problem exists for both iOS and Android. This posts a high usability issue when users attempt to navigate between different services. 

### Acceptance Criteria 
- [x]  Clicking back from Service Details to topic services should bring users back to that service on the flatlist.
- [x]  Clicking back from Service Details to My Bookmarks should bring users back to that service on the flatlist.

### Steps for QA: 
1. Select a service service listing under a topic/My Bookmarks
2. Click the back. button from the Service Detail page
- [ ]  You should be directed to the position of the service that you just selected on the service listing under a topic/My Bookmarks.

rasmus-storjohann-PG:
Solved as part of #1120

jparlina:
This issue reoccurred on Staging V1.33.0, Server 1.3.30

tomy-pg:
This issue is linked to https://github.com/pg-irc/pathways-frontend/issues/1047 :
In summary, I think we will have to use a different library for tab navigation as it is affecting how the flatlist works.

tomy-pg:
I conducted a binary search of this regression and it seems to have occurred when we replaced the `flatlist` in `src/components/search/search_results_component.tsx` with an `AnimatedFlatList`. @niko-pg are you able to look into this?

niko-pg:
@tomy-pg sorry just to verify the issue here. So the bug is that the scroll positions resets to the top when a service is clicked on right?

tomy-pg:
@niko-pg that's right. This is happening in the services in the search list, topic services list and bookmarks list.

rasmus-storjohann-PG:
QA: Places in the UX where we navigate from a detail page to a list page by clicking the back button:
- [ ] Home page -> Topic detail -> Back to home page: Does not remember the position within the home page
- [ ] Bookmark page -> Topics pane -> Topic detail -> Back to bookmark page: Does not remember list position
- [x] Bookmark page -> Services pane -> Service detail -> Back to bookmark page: It remembers the list position
[...]

PR #1173, opened 2020-05-23
Issue #1106 Fix scroll offset bug in search results

niko-pg:
… the listener config.

### DIFF
--- a/src/components/search/search_results_component.tsx
+++ b/src/components/search/search_results_component.tsx
@@ -87,11 +87,7 @@ const renderComponentWithResults = (props: Props): JSX.Element => {
         }
     }, [props.searchOffset]);
 
-    const onScroll = (e: any): void => {
-        props.setScrollOffset(e.nativeEvent.contentOffset.y);
-    };
-
-    const onScrollAnimated = getAnimatedFlatListScrollHandler(onScroll);
+    const onScrollAnimated = getAnimatedFlatListScrollHandler(props.setScrollOffset);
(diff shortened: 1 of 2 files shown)


niko-pg:
If ever you're curious about the Animated api: https://software-mansion.github.io/react-native-reanimated/index.html

ian-lin-PG:
`onPress` event is not triggering consistently for Android devices. I can only tell an event is triggered properly when the item turns opaque which doesn't happen most of the time. The first item that successfully open will have this error message:
image
I have occasions have the service page open much later after I furiously tapped on the service item. When pressing the back button it does returns to correct position within the list.

niko-pg:
@tomy-pg @ian-lin-PG I've basically hit a rabbit hole while fixing that I just decided to move the setting of the offset in `onScrollEndDrag`. In most of my testing, it works. But I'm worried that I missed some edge cases. Let me know if there are still issues.

src/components/search/search_results_component.tsx
```diff
@@ -87,18 +86,18 @@ const renderComponentWithResults = (props: Props): JSX.Element => {
         }
     }, [props.searchOffset]);
 
-    const onScroll = (e: any): void => {
+    const onScrollEnd = (e: any) => {
+        console.log(e.nativeEvent.contentOffset.y)
```
ian-lin-PG, line 90: Noticed these logs in the output, otherwise looks good!

Issue #1047, opened 2020-03-12
bookmarked topics and services list resets to top when one item is unbookmarked


tomy-pg:
Currently in the bookmarking page, we are using a Tab Switcher library to navigate between bookmarked topics and services. Flatlist is having some issues working with the library and we are seeing the Flatlist reset to the top when unbookmarking services or topics.

### Acceptance Criteria 
- [x]  Unbookmarking a service / topic does not reset the list to the top. 

### Steps for QA: 
1.

PR #1176, opened 2020-05-26
Issue #1047 unbookmarking topics and services in Bookmarks resets flatlist position

tomy-pg:


src/components/bookmarks/tab_switcher.tsx
```diff
@@ -14,13 +13,16 @@ import { EmptyComponent } from '../empty_component/empty_component';
...
+
+type Props = TabSwitcherProps & BookmarksProps & ListActions & RouterProps;
 
 export const TabSwitcher = (props: Props): JSX.Element => {
-    const _ = props.i18n._.bind(props.i18n);
     const routes: TabRoutes = [
-        { key: 'topics', title: _(t`Topics`) },
-        { key: 'services', title: _(t`Services`) },
+        { key: 'topics', title: props.i18n._(t`Topics`) },
+        { key: 'services', title: props.i18n._(t`Services`) },
```
rasmus-storjohann-PG, line 26: Double check that this works, you may need the `bind()` magic in there, it's a work-around for Javascript's broken object model

tomy-pg: Just double checked this. This change has not broken anything. I checked the Tab headers in different locales and they were being translated properly and I also ran the `yarn extract-strings` command and was able to successfully extract the Tab header strings. 

Using the `I18n` type instead of `ReactI18nRenderProp` type seems to work without the bind magic.

PR #1181, opened 2020-05-28
Issue #958 Reinstate footer and header when inputs are clicked

niko-pg:


niko-pg:
Putting this is as a separate PR so other bug fixes are not blocked by this.

PR #1184, opened 2020-05-29
Feature/issue 1171 feedback ux issues

rasmus-storjohann-PG:


src/components/main/footer_component.tsx
```diff
@@ -113,6 +115,11 @@ export const FooterComponent: React.StatelessComponent<FooterProps> = (props: Fo
 const isFooterHidden = (props: FooterProps, keyboardIsVisible: boolean): boolean => {
     if (keyboardIsVisible)
         return true;
+
+    const isOnFeedbackScreen = props.feedbackScreen === FeedbackScreen.EditableServiceDetailPage;
```
tomy-pg, line 119: I think you will need to include `FeedbackScreen.RemoveServicePage` and `FeedbackScreen.OtherChangesPage`. The AC for this is to remove the footer from any feedback screen.

rasmus-storjohann-PG: Good catch! Fixed

PR #1185, opened 2020-06-02
Feature/issue 1106 maintain list scroll position

tomy-pg:


src/stores/__tests__/list_offset.test.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    describe('the bookmarked services offset state', () => {
+
+        it('is set to an offset number value by the save bookmarked services offset action', () => {
+            const oldStore = buildDefaultStore();
+            const offset = aNumber();
+            const newStore = reducer(oldStore, saveBookmarkedServicesOffset(offset));
+            expect(newStore.bookmarkedServicesOffset).toBe(offset);
+        });
+    });
```
ian-lin-PG, line 35: If possible I would like to have interaction test with the offset and the bookmark tab. There's differentiation between service book mark list and topic bookmark list.

src/stores/bookmarks_tab.ts
```diff
@@ -0,0 +1,35 @@
...
+    if (!action) {
+        return store;
+    }
+    switch (action.type) {
+        case constants.SET_BOOKMARKS_TAB:
+            return updateBookmarksTab(action.payload.index);
+        default:
+            return store;
+    }
+};
```
rasmus-storjohann-PG, line 28: You're not obligated to use a switch statement just because it's a reducer, you can do
```ts
if (action && action.type === constants.SET_BOOKMARKS_TAB) {
   return updateBookmarksTab(action.payload.index);
}
...
```

src/stores/index.ts
```diff
@@ -67,4 +70,5 @@ export const reducer = combineReducers<Store>({
     feedback: feedback.reducer,
     content: content.reducer,
     listOffset: listOffset.reducer,
+    bookmarksTab: bookmarksTab.reducer,
```
rasmus-storjohann-PG: Is the store getting a little too fine grained here? Maybe it would make sense to combine the list offset and bookmarks tab into one store, UX state or something?

src/stores/bookmarks_tab.ts
```diff
@@ -0,0 +1,35 @@
+import * as constants from '../application/constants';
+import * as helpers from './helpers/make_action';
+
+export enum BookmarksTabStore {
+    Topics,
+    Services,
+}
```
rasmus-storjohann-PG, line 7: The store would normally be a struct with one or more properties on it. When the store is just one enum, then I think that's a sign that the store is too simple and the state is too fragmented. The solution is to combine the bookmark tab and the list position into one store.

src/components/bookmarks/service_bookmarks_component.tsx
```diff
@@ -9,32 +10,53 @@ import { BookmarkServiceAction, UnbookmarkServiceAction, OpenServiceAction } fro
...
+    useEffect((): void => {
+        if (props.bookmarkedServices.length > 0) {
+            flatListRef.current.scrollToOffset({ animated: false, offset: props.listOffset });
         }
-        ItemSeparatorComponent={SearchListSeparator}
-        ListHeaderComponent={<View />}
-    />
-);
\ No newline at end of file
+    }, [props.listOffset]);
```
rasmus-storjohann-PG: Missing items from the list of dependencies is going to be a plentyful source of bugs. Here you need to include `props.bookmarkedServices` and I think `flatListRef` needs to be included as well?

src/components/services/render_service_items.tsx
```diff
@@ -14,9 +14,8 @@ export interface ServiceItemsProps {
     readonly bookmarkService: (service: HumanServiceData) => BookmarkServiceAction;
     readonly unbookmarkService: (service: HumanServiceData) => UnbookmarkServiceAction;
     readonly openServiceDetail: (service: HumanServiceData) => OpenServiceAction;
-    // TO DO make below properties mandatory. Doing this to keep commit small
-    readonly scrollOffset?: number;
-    readonly saveListOffset?: (offset: number) => SaveListOffsetAction;
+    readonly scrollOffset: number;
+    readonly saveListOffset: (offset: number) => SaveListOffsetAction;
```
rasmus-storjohann-PG: Nice

src/components/bookmarks/service_bookmarks_component.tsx
```diff
@@ -10,39 +10,44 @@ import { BookmarkServiceAction, UnbookmarkServiceAction, OpenServiceAction } fro
...
+    }, [props.bookmarkedServicesOffset]);
 
     // tslint:disable-next-line: no-any
     const onScrollEnd = (e: any) => {
         setScrollOffset(e.nativeEvent.contentOffset.y);
     };
 
+    const saveListOffset = (): void => {
+        props.saveBookmarkedServicesOffset(scrollOffset);
+    };
```
rasmus-storjohann-PG, line 46: Potential for confusion with the name `saveListOffset`, since there now are several of these list offsets being saved, the name needs to remain consistent. I might go with this, maybe, which reduces the scope of the confusing name and lets you keep it unchanged in the context of the `renderServiceItems()` function.
```ts
    const serviceItemsProps = { 
          ...props, 
          scrollOffset, 
          saveListOffset:  props.saveBookmarkedServicesOffset(scrollOffset); 
};
```

src/components/search/search_results_component.tsx
```diff
@@ -72,6 +70,7 @@ export const SearchResultsComponent = (props: Props): JSX.Element => {
 };
 
 const renderComponentWithResults = (props: Props): JSX.Element => {
+    const [scrollOffset, setScrollOffset]: readonly [number, (n: number) => void] = useState(props.searchOffset);
```
rasmus-storjohann-PG, line 73: Don't use two names for the same thing unless you have to, here I think `searchOffset` should be used consistently

src/components/bookmarks/service_bookmarks_component.tsx
```diff
@@ -37,8 +37,7 @@ export const ServiceBookmarksComponent = (props: Props): JSX.Element => {
         }
     }, [props.bookmarkedServicesOffset]);
 
-    // tslint:disable-next-line: no-any
-    const onScrollEnd = (e: any) => {
+    const onScrollEnd = (e: NativeSyntheticEvent<ScrollViewProperties>): void => {
```
rasmus-storjohann-PG, line 40: thanks

src/stores/bookmarks_tab.ts
tomy-pg: This makes a lot of sense to me. In my last few commits, you will see that I also added the `HeaderMenuStore` to the `UserExperienceStore` because that store is also just enum. representing the UX of the Header Menu. Let me know if you want me to drop those commits as there was quite a lot of file paths changed...

src/components/services/service_list_component.tsx
```diff
@@ -76,23 +76,31 @@ export const ServiceListComponent = (props: Props): JSX.Element => {
 };
 
 const ValidServiceListComponent = (props: Props): JSX.Element => {
+    const [scrollOffset, setScrollOffset]: readonly [number, (n: number) => void] = useState(props.listOffset);
```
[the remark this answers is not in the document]
tomy-pg: the local `scrollOffset` state is used to to take in the `e.nativeEvent.contentOffset.y` every time `onScrollEndDrag` is triggered. I do not want to save this value to the store every time, only when the user clicks into a Service Detail.

src/stores/bookmarks_tab.ts
rasmus-storjohann-PG: No that sounds right to me ... haven't looked at the commit yet

src/components/bookmarks/service_bookmarks_component.tsx
rasmus-storjohann-PG: I think when in doubt, include them. No harm in including an item that never changes.

src/stores/user_experience/index.ts
```diff
@@ -6,24 +6,46 @@ export interface UserExperienceStore {
...
+    action.type === constants.CLOSE_HEADER_MENU ||
+    action.type === constants.CLOSE_ABOUT_MODAL ||
+    action.type === constants.OPEN_ABOUT_MODAL ||
+    action.type === constants.CLOSE_DISCLAIMER_MODAL ||
+    action.type === constants.OPEN_DISCLAIMER_MODAL ||
+    action.type === constants.SAVE_LOCALE_REQUEST;
+
+    if (isHeaderMenuAction) {
+        return headerMenuReducer(store, action);
+    }
```
rasmus-storjohann-PG, line 47: This should go in the switch statement with fall through, so 
```ts
case constants.CLOSE_HEADER_MENU:
case constants.CLOSE_ABOUT_MODAL:
case constants.OPEN_ABOUT_MODAL:
.....
    return headerMenuReducer(store, action);
```

src/components/topics/topic_detail_connected_component.tsx
```diff
@@ -25,31 +29,36 @@ const mapStateToProps = (store: Store, ownProps: OwnProps): TopicDetailsProps =>
...
+    const manualUserLocation = selectManualUserLocation(store);
     return {
         topic,
         taskIsBookmarked,
         savedTasksIdList,
         history: ownProps.history,
         location: ownProps.location,
+        manualUserLocation,
     };
 };
```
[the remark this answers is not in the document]
tomy-pg: I can revisit this. From what I understand, it is like this to begin with because `matchParams` needs to be defined before `topic` can be defined. `topic` then needs to be defined so that `taskIsBookmarked` can be.

src/components/services/service_list_component.tsx
```diff
@@ -101,6 +101,7 @@ const ValidServiceListComponent = (props: Props): JSX.Element => {
                 renderItem={renderServiceItems(serviceItemsProps)}
                 ItemSeparatorComponent={SearchListSeparator}
                 ListHeaderComponent={<ListHeaderComponent {...props} />}
+                initialNumToRender={props.topicServicesOffset ? services.length : 20}
```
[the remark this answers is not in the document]
tomy-pg: `Flatlist` tries to optimize the amount of items rendered by default to I think 20. However, this optimization kind of fights against us when we are trying to have UX where users can return to their previous position in the `Flatlist`. For instance, if a user navigates to a service detail from a service in position 39 out of a list of 40 items and goes back, we want the initial number of items to render to be 40 or else the scrollToOffset won't work properly.

Issue #1080, opened 2020-03-25
Add click state for external links, & link styling across app


cherrielam:
Right now for links in Topic Detail, About Arrival Advisor, and Disclaimer, that open into an external page, there is no click state. This means users will not be informed whether they've clicked on to the link or not. This can cause some usability issues especially when the internet is slow or when the app is experiencing performance issues.
As a user, he/she would want to be informed on the status of the app, this includes showing a click state of a text link.

Also, the styling of external links is inconsistent between Topic Detail and the following pages: About Arrival Advisor, Disclaimer. Can we make these consistent — teal, same font weight?

Recommendation:
Since parts of Arrival Advisor already indicate a click state (eg. on Service Detail), we could just replicate the same transition to all the intext links. For instance, in the staging app, clicking on different links on service detail will reduce its opacity (example [here](https://drive.google.com/open?id=1TsLRK_TeALRaEe9vowKNs1u0ZL0aUpqQ)). 

@aron-chen-pg made a quick prototype below for demonstration of click state should look like: 
![Intext-link-clickstate](https://user-images.githubusercontent.com/54860686/77418955-de3ef500-6d84-11ea-95b3-128040641f39.gif)

For consideration: @rasmus-storjohann-PG, @bensinjin & others please weigh in —  Should (and could) we add a system alert before the user goes into the external page? In other apps, sometimes you get a confirmation link to "Open link in browser?" with options like "Don't show me again/Always open the link without asking". I am suggesting this because we know from testing that users get confused when they open an external page and are often unclear how to get back to the app.

### Acceptance Criteria 
- [ ]  Click state added on in-text links (within Topic Detail). Same click state already exists on links in Service Detail.
- [ ]  Decision made as to whether a system alert to "Open link in browser" is added. If so, this system alert should pop up when the user clicks an external link that is in browser. If possible, add a way for users to dismiss future asks.
- [ ] Styling of links in Topic Detail and About Arrival Advisor, Disclaimer is consistent. All links in teal and same font weight.

### Steps for QA: 
1.

aron-chen-pg:
Regarding opening a disclaimer dialogue, I made another prototype to kind of replicate how that might look like:
![Intext-link-interaction v2](https://user-images.githubusercontent.com/54860686/77575448-c6549780-6e90-11ea-9b60-a47b6f156f80.gif)

[...]

PR #1187, opened 2020-06-03
Feature/issue 1080 link ui fixes

ian-lin-PG:


src/components/link/link_component.tsx
```diff
@@ -27,8 +26,8 @@ export const openURL = (url: string): void => {
 export const Link = (props: LinkProps): JSX.Element => {
     const onPress = (): void => openURL(props.href);
     return (
-        <Text onPress={onPress} style={[textStyles.paragraphURL, props.style]}>
-            {props.children}
+        <Text onPress={onPress} style={props.style}>
+            {props.children + ' '}
```
niko-pg, line 30: `children`  can be any renderable element (text, component, etc). Not sure how it would act if you add to a space.

I guess if `Link` is just going to be used for texts then this should be fine..

ian-lin-PG: This change is responding to this [comment](https://github.com/pg-irc/pathways-frontend/issues/1080#issuecomment-638463711) on the issue. The `props.children` should only be url strings in the context of the link component. I will look into ways to enforce this.

Issue #1199, opened 2020-06-22
Feedback: Thank you for your contribution no longer showing


tomy-pg:
When users send feedback they have usually received a `Thank you for your contribution` message once they submit the information.

### Acceptance Criteria 
- [x]  `Thank you for your contribution` toast is shown to users when they send feedback.

### Steps for QA: 
1.

ian-lin-PG:
PASSED QA on 1.38.0

PR #1201, opened 2020-06-23
Issue #1199 use former logic for showing different modal dialogs to e…

tomy-pg:
In this PR I had to revert the using the former way of showing different modal dialogs since `onModalHide` will not trigger if `isVisible` is always set to true.


ian-lin-PG:
Tested on staging vs the PR. `Thank you for your contribution` shows up in the PR.

PR #1209, opened 2020-06-29
Feature/issue 1186 address remaining feedback polish

tomy-pg:


src/components/services/service_detail_component.tsx
```diff
@@ -134,6 +134,16 @@ export const ServiceDetailComponent = (props: Props): JSX.Element => {
         props.submitFeedback(feedbackInput);
     };
 
+    const hasNoFeedbackToSend = (): boolean => {
+        const sendableFeedback = R.pickBy(isSendableFeedbackField, feedbackInput);
+        if (R.isEmpty(sendableFeedback)) {
+            return true;
+        }
+        return false;
+    };
```
ian-lin-PG, line 145: Disabling the submit button when the airtable is empty, I see.

src/components/feedback/submit_feedback_button.tsx
```diff
@@ -5,7 +5,9 @@ import { applicationStyles, textStyles } from '../../application/styles';
 import { Text } from 'native-base';
 import { Trans } from '@lingui/react';
 
-export const SubmitFeedbackButton = (props: { readonly isVisible: boolean, readonly onPress: () => void }): JSX.Element => {
+export const SubmitFeedbackButton = (props: {
+    readonly isVisible: boolean, readonly disabled?: boolean, readonly onPress: () => void,
```
rasmus-storjohann-PG: `boolean` can represent true/false, `boolean?` can represent three states, true/false/undefined. Do you need three states here? If not, you should just use `boolean`. Many a bug comes from using types with domains that are larger than necessary, always chose the type that most accurately reflects the set of valid values.

src/components/services/service_detail_component.tsx
```diff
@@ -134,6 +134,16 @@ export const ServiceDetailComponent = (props: Props): JSX.Element => {
         props.submitFeedback(feedbackInput);
     };
 
+    const hasNoFeedbackToSend = (): boolean => {
+        const sendableFeedback = R.pickBy(isSendableFeedbackField, feedbackInput);
+        if (R.isEmpty(sendableFeedback)) {
+            return true;
+        }
+        return false;
```
rasmus-storjohann-PG, line 144: I like your overall solution here, this is the right approach. However, there is a subtle code smell here: You're using Ramda and then you're adding some logic of your own. You are not the first person who has needed this particular extra logic ... maybe there is already something in Ramda to help you...

src/components/feedback/other_remove_service_component.tsx
```diff
@@ -112,6 +110,13 @@ export const OtherRemoveServiceComponent = (props: FeedbackOtherRemoveServicePro
         }
     };
 
+    const hasNoFeedbackToSend = (): boolean => {
+        if (!feedback) {
+            return true;
+        }
+        return false;
+    };
```
rasmus-storjohann-PG, line 118: No, just return `!feedback` please.

src/components/feedback/submit_feedback_button.tsx
```diff
@@ -6,7 +6,7 @@ import { Text } from 'native-base';
 import { Trans } from '@lingui/react';
 
 export const SubmitFeedbackButton = (props: {
-    readonly isVisible: boolean, readonly disabled?: boolean, readonly onPress: () => void,
+    readonly isVisible: boolean, readonly disabled: boolean, readonly onPress: () => void,
```
rasmus-storjohann-PG, line 9: Thanks for dropping the optional

src/stores/__tests__/feedback.test.ts
```diff
@@ -147,11 +147,19 @@ describe('feedback reducer', () => {
...
     });
 
-    test('closing dialog from a data entry screen brings up confirmation modal', () => {
+    test('closing dialog from an empty data entry screen brings user back to service detail', () => {
         const oldScreen = aBoolean() ? FeedbackScreen.EditableServiceDetailPage : FeedbackScreen.OtherChangesPage;
         const oldStore = new FeedbackStoreBuilder().withScreen(oldScreen).build();
         const action = close();
         const newStore = reducer(oldStore, action);
+        expect(newStore.screen).toEqual(FeedbackScreen.ServiceDetail);
+    });
```
rasmus-storjohann-PG, line 156: Nice. The big benefit of using redux to drive this UX is to be able to unit test the UX. Most of the time we don't unit test our UX, so I'm really happy to be able to do it here.

src/components/feedback/other_remove_service_component.tsx
rasmus-storjohann-PG: Also consider that names names without negation in them are easier to read that named with negation. It's often a trade-off so it depends a bit on what's happening at the function's call sites, but generally, `hasFeedbackToSend()` is a better name than `hasNoFeedbackToSend()`, or in this case, flipping the name removes the need for having this function at all.

src/components/feedback/submit_feedback_button.tsx
tomy-pg: I hear you on this. I understand now that it's not very clear for why I made this property optional. I did this to keep the commit small. `SubmitFeedbackButton` is used in two other components as well and I didn't want to muddle this commit with changes to that component as well since the logic I was using slightly differed.

src/components/feedback/other_remove_service_component.tsx
tomy-pg: Fixed. I flipped the name and ended up removing the function completely.

src/components/services/service_detail_component.tsx
tomy-pg: I attempted to remove the negation from this function name and ended up with logic like this:
`return R.not(R.isEmpty(sendableFeedback))`. 
I am not entirely sure if the trade-off is worth it or not but I do think the naming of this function is easier to read. If I kept the function name with negation I would have logic this:
`return R.isEmpty(sendableFeedback)`. 
Let me know what you think!

src/components/services/service_detail_component.tsx
```diff
@@ -136,12 +136,9 @@ export const ServiceDetailComponent = (props: Props): JSX.Element => {
...
     };
 
-    const hasNoFeedbackToSend = (): boolean => {
+    const hasFeedbackToSend = (): boolean => {
         const sendableFeedback = R.pickBy(isSendableFeedbackField, feedbackInput);
-        if (R.isEmpty(sendableFeedback)) {
-            return true;
-        }
-        return false;
+        return R.not(R.isEmpty(sendableFeedback));
```
ian-lin-PG, line 141: Could you not do `R.isEmpty(sendableFeedback) ? false : true`. It's another variation on the same logic, but functionally should be the same.

tomy-pg: This is related to @rasmus-storjohann-PG 's earlier comment about code smell as I would be using Ramda in addition to my own logic when Ramda has its own solution with`R.not` .

PR #1219, opened 2020-07-07
Feature/issue 1212 client package upgrade

ian-lin-PG:


package.json
```diff
@@ -58,26 +58,26 @@
...
+        "react": "16.11.0",
+        "react-native": "https://github.com/expo/react-native/archive/sdk-38.0.0.tar.gz",
         "react-native-dotenv": "^0.2.0",
         "react-native-gesture-handler": "~1.6.0",
         "react-native-map-link": "^2.7.3",
         "react-native-markdown-display": "^6.1.1",
         "react-native-markdown-renderer": "^3.2.8",
         "react-native-modal": "^11.5.6",
-        "react-native-reanimated": "^1.9.0",
+        "react-native-reanimated": "~1.9.0",
```
tomy-pg, line 82: QA on iPhone XR: The scroll animation has regressed. 1/3 of the hamburger menu is hidden making it much more difficult to touch and when I scroll down the footer navigation no longer completely hides. 

One thing to note is that the scroll animation uses `react-native-reanimated`, not react-native's `animate` module. 
I recognize in the SDK upgrade that any use of react-native's `animate` module needs to set `useNativeDriver` to either true or false. I would dig in to see if `react-native-reanimated` requires the same thing now.

jestSetupFile.js
```diff
@@ -0,0 +1,3 @@
+import mockAsyncStorage from '@react-native-community/async-storage/jest/async-storage-mock';
+
+jest.mock('@react-native-community/async-storage', () => mockAsyncStorage);
```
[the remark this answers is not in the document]
tomy-pg: The tests still pass but now we are seeing the `console.error` show up in the `questionnaire.tests.ts`. I think this is fine.

package.json
rasmus-storjohann-PG: I get a warning that `useNativeDriver` needs to be set.

PR #1220, opened 2020-07-07
Feature/issue 1200 add back functionality to feedback submit modal

tomy-pg:


src/components/feedback/contact_information_modal.tsx
```diff
@@ -39,7 +39,7 @@ export const ContactInformationModal =
     return (
         <I18n>
             {({ i18n }: I18nProps): JSX.Element => (
-                <Modal isVisible={isVisible} backdropTransitionOutTiming={0} onModalHide={onModalHide(i18n)}>
+                <Modal isVisible={isVisible} backdropTransitionOutTiming={0} onBackButtonPress={back}>
```
rasmus-storjohann-PG, line 42: I think `onBackButtonPress()` is a better name than just `back()`

src/components/feedback/other_remove_service_component.tsx
```diff
@@ -135,3 +136,17 @@ export const OtherRemoveServiceComponent = (props: FeedbackOtherRemoveServicePro
...
+    }
+    switch (feedback.type) {
+        case 'remove_service':
+            return feedback.reason;
+        case 'other_feedback':
+            return feedback.value;
+        default:
+            return '';
+    }
+};
```
rasmus-storjohann-PG, line 152: I'm trying to figure out why this doesn't feel right. I think this feels like selector code, it's stuff that the component shouldn't know about. It connects to the problem that too much information is passed to the component in this line below: 
```ts
        feedback: selectFeedback(store),
```
The component doesn't need all of the feedback, it just needs a piece and `feedbackInputByType()` is the code the component needs to contain to fish out the information that should already be given to it. So, make a selector that fishes out this information, and pass that to the component. It will be immediately obvious why the component needs that specific information, rather than the less clear picture I get from seeing that it needs the entire feedback blob of data.

src/components/services/service_detail_component.tsx
```diff
@@ -143,6 +146,13 @@ export const ServiceDetailComponent = (props: Props): JSX.Element => {
 
     const isSendableFeedbackField = (value: FeedbackField): boolean => value.shouldSend === true && value.value.length > 0;
 
+    const back = (): void => {
+        if (isOtherRemoveServiceFeedback(props.feedback)) {
+            goToRouteWithParameter(Routes.OtherFeedback, props.match.params.serviceId, props.history)();
+        }
```
rasmus-storjohann-PG, line 152: So why is it that we navigate from the other feedback screen to the service detail screen when we hit submit? Was that something Cherrie specifically asked for? Seems to me we could just stay on the other feedback screen with the contact information modal over it. Then we wouldn't need to change the route to go back to the correct screen after. That would simplify this PR in a couple of places, I think.

src/stores/feedback/reducer.ts
```diff
@@ -73,3 +75,16 @@ const submitOrDiscardReducer = (store: FeedbackStore, action: ReducerActions): F
             return store;
     }
 };
+
+const backButtonReducer = (store: FeedbackStore): FeedbackStore => {
+    switch (store.feedback.type) {
+        case 'service_feedback':
+            return { ...store, modal: FeedbackModal.None, screen: FeedbackScreen.EditableServiceDetailPage };
```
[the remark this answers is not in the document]
tomy-pg: Perhaps I am misunderstanding your question but `store.feedback.type` is being used here instead of `FeedbackScreen`  because once the user clicks `Submit` and they see the `ContactInformation` modal, they are navigated back to the `Service Detail` screen. Because of this UX, `FeedbackScreen` does not contain the data I need for this reducer, whereas `feedback.type` does.

src/components/services/service_detail_component.tsx
tomy-pg: This is a part of the mocks. https://app.zeplin.io/project/5bc7bfa42e72260acb03aa86/screen/5e0fe74134bdd11c327cb29f

I agree, it would simplify this PR if users are brought back to the same feedback screen after they send in their contact information but I am unsure if this change would make the UX for this better or worse.

src/components/feedback/other_remove_service_component.tsx
tomy-pg: Ah yes, a selector for this is the much better solution.

src/components/feedback/contact_information_component.tsx
```diff
@@ -77,25 +113,52 @@ export const ContactInformationModal =
...
+const headerLabelByType = (feedbackType: string): string => {
+    if (feedbackType === 'service_feedback') {
+        return 'Change name or other details';
+    }
+
+    if (feedbackType === 'remove_service') {
+        return 'This service no longer exists';
+    }
+
+    return 'Other suggestions';
```
ian-lin-PG, line 159: When I see multiple if statements I see potential switch statements, not a big issue.

src/components/services/thank_you_message_or_empty_component.tsx
```diff
@@ -0,0 +1,17 @@
...
+import { showToast } from '../../application/toast';
+
+interface ThankYouMessageOrEmptyProps {
+    readonly i18n: I18n;
+    readonly isSendingFeedback: boolean;
+}
+
+export const ThankYouMessageOrEmptyComponent = ({i18n, isSendingFeedback}: ThankYouMessageOrEmptyProps): JSX.Element => {
+    if (isSendingFeedback) {
+        showToast(i18n._(t`Thank you for your contribution!`), 3000);
```
ian-lin-PG, line 14: I see the duration is being reused here, nice!

src/stores/feedback/types.ts
```diff
@@ -3,6 +3,7 @@ export interface FeedbackStore {
     readonly modal: FeedbackModal;
     readonly userInformation: undefined | UserInformation;
     readonly feedback: undefined | Feedback;
+    readonly hasFeedbackToSend: boolean;
```
rasmus-storjohann-PG, line 6: Is it not true that `hasFeedbackToSend === true` is equivalent to `feedback !== undefined`?

tomy-pg: Yes I am noticing this as I am working on the selector code. I will likely drop this commit as well...

src/application/toast.ts
```diff
@@ -2,11 +2,11 @@
 import { textStyles, applicationStyles } from './styles';
 import { Toast } from 'native-base';
 
-export const showToast = (text: string): void => {
+export const showToast = (text: string, duration: number): void => {
     Toast.show({
         style: applicationStyles.toast,
         textStyle: textStyles.toast,
-        duration: 2000,
+        duration,
```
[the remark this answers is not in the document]
tomy-pg: The `Thanks for your contribution` toast for feedback appears a bit earlier in the feedback flow than it once was. Previously, it showed once the `Contact Information Modal` closed. However, since this modal is now a screen, this toast is shown once the `isSending` feedback state is true, triggering the toast a bit earlier. I bumped up the duration for the toast to 3 seconds to ensure the user has ample time to see the message.

src/components/feedback/contact_information_component.tsx
tomy-pg: Agreed, I will clean this up.

src/components/feedback/contact_information_component.tsx
```diff
@@ -2,46 +2,82 @@
...
             isEmployee: !userInformation.isEmployee,
         });
 
-    const buttonLabel = userInformation.email.length ? t`Email me updates` : t`Finish without email`;
+    const onBackButtonPress = (): void => {
+        if (isOtherRemoveServiceFeedback(feedbackType)) {
+            goToRouteWithParameter(Routes.OtherFeedback, serviceId, history)();
+        } else {
+            goToRouteWithParameter(Routes.ServiceDetail, serviceId, history)();
+        }
```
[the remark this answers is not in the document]
tomy-pg: In this case, I want to do one thing or another AND trigger an action. I am cleaning this part of the code any ways to make it more simple to read.

src/selectors/feedback/select_service_feedback.ts
```diff
@@ -0,0 +1,14 @@
+import { Store } from '../../stores';
+import { selectFeedback } from './select_feedback';
+import { getEmptyServiceFeedback } from '../../stores/feedback';
+import { ServiceFeedback } from '../../stores/feedback/types';
+
+export const selectServiceFeedback = (store: Store): ServiceFeedback => {
+    const feedback = selectFeedback(store);
+
+    if (feedback && feedback.type === 'service_feedback') {
```
rasmus-storjohann-PG, line 9: Cool to see if you could do `if (feedback?.type === 'service_feedback')` here

tomy-pg: It works! Reads nicely.

PR #1226, opened 2020-07-24
Feature/issue 1142 use expo android status bar

tomy-pg:
It has been solved. Turns out native-base's `Header` component has a `androidStatusBarColor` property that was overwriting all of our changes to the `StatusBar`. We never used this property in our HeaderComponents so it was turning our bars purple with our knowing.

Please test this PR on your Android device as setting the backgroundColour only applies to them. We should no longer see a purple bar.

### DIFF
--- a/src/application/styles.ts
+++ b/src/application/styles.ts
@@ -29,6 +29,7 @@ export const colors = {
     sunshine: '#f2b134',
     topaz: '#11cac0',
     greyBorder: '#dedede',
+    transparentBlack: '#00000088',
 };
 
 export const values = {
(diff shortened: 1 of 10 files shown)


src/application/styles.ts
```diff
@@ -29,6 +29,7 @@ export const colors = {
     sunshine: '#f2b134',
     topaz: '#11cac0',
     greyBorder: '#dedede',
+    transparentBlack: '#00000088',
```
rasmus-storjohann-PG, line 32: Looking at other apps on my phone, it looks to me to be common to use an app specific colour for the top status bar, usually a slightly darker shade for the header than for the rest of the app. For our app I think it could make sense to use the teal we already have or a darker version of the same, such as `#005e66` or `#007580`, see https://www.w3schools.com/colors/colors_picker.asp as a nice tool to play around.

cherrielam: I suggest the same teal (#00838F), not sure there is a need for it to be differentiated when the background is also 00838f, if users can still see the text and battery stuff

Issue #982, opened 2020-01-16
Link from app feedback UI to web page describing how our feedback system works


rasmus-storjohann-PG:
### Acceptance Criteria 
- [ ] There is a link in the app for more information about the feedback system
- [ ] The link opens a page in the app with localized information about this, similar to the About page

### Questions
- [x] Mock is needed, link will probably be on [this](https://zpl.io/bPn5L4m) modal.

cherrielam:
Just checking the backlog and noticed this — is this still needed? @rasmus-storjohann-PG If so, I/ @aron-chen-pg should hop on this soon?

cherrielam:
### String added on the Feedback Options Modal:
Add "Learn more" (see below)
> Your suggestions will be reviewed by our team before making final changes to the service information. **Learn more**

### Feedback system explainer:

> # Services listed on Arrival Advisor
>  
> ### Where do the services listed on Arrival Advisor come from?
> The services listed on Arrival Advisor come from [bc211](http://www.bc211.ca), a nonprofit information and referral organization that publishes a searchable online directory of community, government, and social services in British Columbia. bc211’s staff members collect and update this services information.
> 
> ### I found some inaccurate or missing services information in the app. What do I do?
> You can make a suggestion by clicking “Suggest an update” at the bottom of a service’s detail page in Arrival Advisor. Then, you can provide more information in the form to let us know what needs to be updated. You can choose to leave your contact email so that our team can follow up with you. The Arrival Advisor team will work with bc211 to review your suggestions before we make the final changes.
> 
> ### How can I get my organization or service listed?
> You can read [bc211’s Inclusion Policy here](http://www.bc211.ca/inclusion-policy/) to see what organizations and services are listed. If you believe that your organization or service fits the inclusion criteria, you can email updates@bc211.ca so the bc211 team can review your organization or service. You can also visit http://www.bc211.ca for more information.

Here's what I've got so far. I'd like to get some eyes to review for any edits before it goes forward for translation (@jparlina @tomy-pg @ian-lin-PG @rasmus-storjohann-PG)? 

Also, content-wise, am I missing anything?

rasmus-storjohann-PG:
That looks good to me

jparlina:
Looks good @cherrielam. Where does the link to the explainer appear on the app?

tomy-pg:
This should be high priority as it is blocking translation QA.

cherrielam:
[...]

PR #1231, opened 2020-07-27
Feature/issue 982 feedback explainer

ian-lin-PG:


src/stores/feedback/types.ts
```diff
@@ -55,4 +56,9 @@ export interface RemoveServiceFeedback {
     readonly reason: string;
 }
 
-export type Feedback = ServiceFeedback | OtherFeedback | RemoveServiceFeedback;
+export interface ExplainFeedback {
+    readonly type: 'explain_feedback';
+    readonly reason: string;
```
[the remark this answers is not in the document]
ian-lin-PG: It was an artifact of building off the simplest Feedback types of `OtherFeedback` and `RemoveServiceFeedback`, in hindsight it serves no purpose at all. Will remove in next commit.

src/components/feedback/explain_feedback_component.tsx
```diff
@@ -1,14 +1,61 @@
...
+    const history = useHistory();
+    const onClosePress = (): void => {
+        props.close();
+        goToRouteWithParameter(Routes.ServiceDetail, props.match.params.serviceId, history)();
+    };
 
     return (
+        <Content>
+            <HeaderComponent
+                headerLabel={t`Service listed on Arrival Advisor`}
```
tomy-pg, line 29: This needs to be `Services listed on Arrival Advisor`. The **s**  is missing at the end of services, which will prevent translation from working. This will also be consistent with the mocks.

src/components/feedback/explain_feedback_component.tsx
```diff
@@ -29,13 +29,13 @@ export const ExplainFeedbackComponent = (props: props): JSX.Element => {
                 headerLabel={t`Service listed on Arrival Advisor`}
                 close={onClosePress}
             />
-        <Content style={{ marginTop: 5, paddingLeft: 15, paddingRight: 5 }}>
+        <Content style={{ paddingTop: 15, paddingHorizontal: 15 }}>
```
tomy-pg, line 32: The padding you are adding here does not seem to be working while on the Arabic locale.
Simulator Screen Shot - iPhone X - 2020-07-28 at 12 05 14

src/components/feedback/explain_feedback_component.tsx
```diff
@@ -1,14 +1,61 @@
...
+                </Trans>
+            </ParagraphComponent>
+            <Text style={[textStyles.headlineH2StyleBlackLeft, { paddingHorizontal: values.backgroundTextPadding }]}>
+                <Trans>How can I get my organization or service listed?</Trans>
             </Text>
+            <ParagraphComponent>
+                <Trans>
+                You can read bc211’s Inclusion Policy here to see what organizations and services are listed. If you believe that your organization or service fits the inclusion criteria, you can email <Link href={'mailto:support@expo.io'} style={textStyles.link}>updates@bc211.ca</Link> so the bc211 team can review your organization or service. Visit <Link href={'http://www.bc211.ca'} style={textStyles.link} >bc211</Link> for more information.
```
tomy-pg: The ending of this should say: `Visit http://www.bc211.ca for more information."` to be consistent with the mock as well as to make translations work. Currently it days `Visit bc211 for more information` which is not a translated string.

rasmus-storjohann-PG:
How hard would it be to make it so that when you close the explainer, the "Suggest an update" modal is still visible? This should be a change in the reducer so it might be very straightforward?

ian-lin-PG:
> How hard would it be to make it so that when you close the explainer, the "Suggest an update" modal is still visible? This should be a change in the reducer so it might be very straightforward?

Currently none of our navigation from feedback screen back to service detail keeps the modal open, just want to double check that only closing the explainer will keep the modal open. I was recycling the `onClose` handler from the other feedback screens, so I just need to customize for a new handler.

tomy-pg:
@rasmus-storjohann-PG it seems very straight forward, but I remember trying to implement a similar solution with a back button and using the reducer to make this change and having one of the device platforms fighting me on having the modal open.

rasmus-storjohann-PG:
Hitting close on the other feedback screens should take you back to the service detail, because the implication of hitting close is to cancel the flow. But hitting close on the explainer is doesn't have the implication of cancelling, but rather to go back to where you were before, so on the modal. @cherrielam for comment on that. If this is difficult to make work (see @tomy-pg 's comment) then don't worry about it.

src/components/feedback/choose_mode_modal.tsx
```diff
@@ -67,10 +68,14 @@ const OptionButton = (props: { readonly name: JSX.Element, readonly onPress: ()
...
-const SubtitleComponent = (): JSX.Element => (
+const SubtitleComponent = (props: ButtonsComponentProps): JSX.Element => (
     <View style={{ marginHorizontal: 10, marginBottom: 15 }}>
-        <Text style={[textStyles.paragraphSmallStyleLeft, { fontSize: 14 }]}>
+        <Text style={[textStyles.paragraphSmallStyleLeft, { fontSize: 12 }]}>
             <Trans>Your suggestions will be reviewed by our team before making final changes to the service information.</Trans>
+            <Text style={[textStyles.messageLink,  { fontSize: 13 }]} onPress={props.onChooseExplainFeedback}>
+                {' '}
+                <Trans>Learn More</Trans>
+            </Text>
```
tomy-pg, line 78: Please see - https://github.com/pg-irc/pathways-frontend/issues/982#issuecomment-642894645

It is vital that strings match exactly what is shown in mocks / issues as I use these as the source of truth when sending strings for translation. As well as this, the client does not translate strings if it cannot find the exact string in its catalogue. This situation is usually avoided with the `yarn extract-strings` command, but we had to send translations off before this PR had even begun developing. In any case, I recommend copying and pasting strings wherever you can.

`Line 74: Your suggestions will be reviewed by our team before making final changes to the service information.`
should be: `Your suggestions will be reviewed by our team before making changes to the service information.`
**final** should be removed.

`Line 77: Learn More`
should be: `Learn more`
**remove capitalization of second word**

PR #1238, opened 2020-07-31
Feature/issue 1227 include agency name in feedback payload

rasmus-storjohann-PG:


rasmus-storjohann-PG:
Yup, those are the correct QA steps, that should have been in the story I guess.

Issue #996, opened 2020-01-27
Fax on Service Detail Page should not be clickable (i.e. triggers a call on the mobile device)


jparlina:
Users should not be able to click on Fax row/card on the Service Detail Page, which then triggers a call on the mobile device. From the user experience perspective, we should not provide functions/options that are not usable for the users (i.e. can't send a fax from the app).

### Acceptance Criteria 
- [ ]  Users should not be able to click on Fax row/card on the Service Detail Page

### Steps for QA: 
1.  Open a service record that has a fax number (e.g. select the Topic "Contact workers at your local settlement agency > click "Find related services near me" > select "YMCA of Greater Vancouver - Connections Program" from the service listing)
2. On the Service Detail Page, scroll down to the Fax number
3. Click the Fax row - you should not be able to click it and it should not trigger a call on your mobile device

![File_001.png](https://images.zenhubusercontent.com/5ceeccc646b329310b95bccd/f32640ae-96bd-493f-b50f-9f3a09c82afd)

ian-lin-PG:
The card is clickable but no phone call event is triggered.

ian-lin-PG:
QA Passed 1.43.0, Fax number no longer has icon does not trigger the phone app. The opacity still changes but does not trigger any behaviours.
![Screenshot_20200817-102537](https://user-images.githubusercontent.com/52009863/90425134-22080300-e074-11ea-90c1-974deb200792.jpg)

ian-lin-PG:
After confirming with @cherrielam, it is preferred that the touchable opacity of the fax number to be removed entirely to avoid any touch feedback.

tomy-pg:
Staging version 1.44.0 - PASS

PR #1242, opened 2020-08-06
Issue #996 Don't call fax numbers

rasmus-storjohann-PG:


src/components/phone_numbers/phone_numbers_component.tsx
```diff
@@ -48,7 +49,7 @@ const buildPhoneNumber = R.curry((props: Props, phoneNumber: PhoneNumber, index:
             <CardButtonComponent
                 leftContent={renderSinglePhoneNumber(phoneNumber)}
                 rightContent={<ServiceDetailIconComponent name={'phone'} />}
```
tomy-pg, line 51: For UX, I think we should remove the phone icon if it's a fax number. Currently this icon is still clickable for fax numbers, it just does not send users to the call screen like other phone numbers. What do you think @rasmus-storjohann-PG @cherrielam ?
```
rightContent={isFaxNumber ? <EmptyComponent /> : <ServiceDetailIconComponent name={'phone'} />
```

rasmus-storjohann-PG: I like it, I'll make that change

PR #1249, opened 2020-08-17
Issue #996 Create a non clickable fax numbers component

ian-lin-PG:


src/components/phone_numbers/phone_numbers_component.tsx
```diff
@@ -48,11 +48,18 @@ const buildPhoneNumber = R.curry((props: Props, phoneNumber: PhoneNumber, index:
...
+                    leftContent={renderSinglePhoneNumber(phoneNumber)}
+                    rightContent={<EmptyComponent />}
+                    onPress={undefined} />
+                :
+                <CardButtonComponent
+                    leftContent={renderSinglePhoneNumber(phoneNumber)}
+                    rightContent={phoneIcon}
+                    onPress={callPhoneNumber}
+                />
+            }
```
rasmus-storjohann-PG, line 62: I'm not a fan of these long ternary expressions. Can you make a separate component that all it does is to decide which kind of card button component you get. It should still be called with the `<Foo/>` syntax, not `{foo()}`. We can talk about it if that doesn't make sense.

src/components/card_button_component.tsx
```diff
@@ -10,14 +10,23 @@ export interface CardButtonProps {
...
-                    {props.leftContent}
-                </View>
-                <View style={{ flex: .5, alignItems: 'flex-end', paddingRight: 10 }}>
-                    {props.rightContent}
-                </View>
-            </View>
+            <NonClickableCardButtonComponent
+                leftContent={props.leftContent}
+                rightContent={props.rightContent}
+                onPress={undefined}/>
```
rasmus-storjohann-PG, line 16: I figured it was just the fax numbers that should not be clickable, here it looks like this applies more widely...

src/components/phone_numbers/phone_numbers_component.tsx
ian-lin-PG: Made a `PhoneNumberComponent` with `PhoneNumberProps` that switches to fax or card button components.

src/components/card_button_component.tsx
ian-lin-PG: I opted to backtrack on this change cause it's premature optmization. If more fields becomes unclickable we know we change change here.

PR #1255, opened 2020-08-27
Feature/issue 1042 new org detail page

ian-lin-PG:


src/validation/services/types.ts
```diff
@@ -29,6 +29,7 @@ export interface HumanServiceData {
     readonly addresses: ReadonlyArray<Address>;
     readonly website: string;
     readonly email: string;
+    readonly organizationId: string;
```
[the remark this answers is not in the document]
ian-lin-PG: The validation happens at the ajv stage, as long as the struct and object share the same structure it will pass. I changed both in this case to avoid the test from failing. I can add in a specific test for testing `organizationID` or more likely create a test that checks every field.

src/stores/__tests__/services.test.ts
```diff
@@ -178,7 +178,20 @@ describe('services reducer', () => {
...
+            expect(storeService.name).toEqual(aSearchService.name);
+            expect(storeService.description).toEqual(aSearchService.description);
+            expect(storeService.phoneNumbers).toEqual(aSearchService.phoneNumbers);
+            expect(storeService.addresses).toEqual(aSearchService.addresses);
+            expect(storeService.organizationId).toEqual(aSearchService.organizationId);
+            expect(storeService.website).toEqual(aSearchService.website);
+            expect(storeService.email).toEqual(aSearchService.email);
+            expect(storeService.organizationName).toEqual(aSearchService.organizationName);
+            expect(storeService.bookmarked).toEqual(aSearchService.bookmarked);
+            expect(storeService.lastVerifiedDate).toEqual(aSearchService.lastVerifiedDate);
```
rasmus-storjohann-PG: Single responsibility also applies to tests, where it usually means that each test should test just one thing, meaning that ideally a test only have one assert statement. Often it makes sense to have more than one, but this approach of having this many is a code smell. In this case, there is a very simple change that can be made to `expect(storeService).toEqual(aSearchService);`, but in other cases, a deeper rethink of the test can often make for real improvements...

src/components/organizations/organization_detail_component.tsx
```diff
@@ -30,55 +92,162 @@ export const OrganizationDetailComponent = (): JSX.Element => {
...
+    const onScrollEnd = (e: NativeSyntheticEvent<ScrollViewProperties>): void => {
+        // tslint:disable-next-line: no-expression-statement
+        setOrganizationServicesOffset(e.nativeEvent.contentOffset.y);
+    };
+
+    const renderItem = renderServiceItems({
+        ...props,
+        scrollOffset: organizationServicesOffset,
+        saveScrollOffset: props.saveOrganizationServicesOffset,
+    });
```
rasmus-storjohann-PG, line 239: My review comment here wasn't very clear, this is what I had in mind

rasmus-storjohann-PG (approved):
I made some additional tweaks to the PR. It would really help if you figure out how to get linter errors to show up in your IDE.

src/components/organizations/organization_detail_component.tsx
```diff
@@ -30,55 +92,162 @@ export const OrganizationDetailComponent = (): JSX.Element => {
...
+        />
     </Content>
 );
 
-const AboutContactDetails = (): JSX.Element => (
-    <View style={{ paddingHorizontal: values.backgroundTextPadding }}>
-        <Text>TODO</Text>
-    </View>
-);
+const AboutContactDetails = (props: AboutOrganizationProps): JSX.Element => {
```
rasmus-storjohann-PG, line 185: We have a really simple naming convention for props and actions passed in to components, for `FooComponent`, they should be called `FooComponentProps` and `FooComponentActions`. Following conventions like this makes the code much simpler to work with.

src/components/organizations/organization_detail_component.tsx
```diff
@@ -164,7 +166,7 @@ const OrganizationDetailHeader = (props: {
     );
 };
 
-const AboutTab = (props: OrganizationContactDetailsProps): JSX.Element => (
+const AboutTab = (props: AboutOrganizationProps): JSX.Element => (
```
rasmus-storjohann-PG, line 169: Naming convention for props applies here too, I would probably rename this to AboutTabComponent and AboutTabComponentProps

src/components/organizations/organization_detail_connected_component.tsx
```diff
@@ -1,34 +1,34 @@
...
-const mapStateToProps = (store: Store, ownProps: RouterProps): OrganizationDetailProps => {
-    return {
-        history: ownProps.history,
-        organizationServicesOffset: selectOrganizationServicesOffset(store),
-    };
-};
+const mapStateToProps = (store: Store, ownProps: RouterProps): OrganizationDetailProps => ({
+    history: ownProps.history,
+    organizationServicesOffset: selectOrganizationServicesOffset(store),
+});
```
rasmus-storjohann-PG, line 15: Thanks, this is better, try to stay consistent with using this form when possible.

src/stores/user_experience/index.ts
```diff
@@ -52,6 +52,11 @@ export const reducer = (store: UserExperienceStore = buildDefaultStore(), action
                 ...store,
                 bookmarkedServicesOffset: action.payload.offset,
             });
+        case constants.SAVE_ORGANIZATION_SERVICES_OFFSET:
+            return ({
+                ...store,
+                organizationServicesOffset: action.payload.offset,
+            });
```
rasmus-storjohann-PG, line 59: This kind of change should *always* be unit tested.

src/stores/__tests__/services.test.ts
```diff
@@ -180,18 +180,7 @@ describe('services reducer', () => {
...
-            expect(storeService.phoneNumbers).toEqual(aSearchService.phoneNumbers);
-            expect(storeService.addresses).toEqual(aSearchService.addresses);
-            expect(storeService.organizationId).toEqual(aSearchService.organizationId);
-            expect(storeService.website).toEqual(aSearchService.website);
-            expect(storeService.email).toEqual(aSearchService.email);
-            expect(storeService.organizationName).toEqual(aSearchService.organizationName);
-            expect(storeService.bookmarked).toEqual(aSearchService.bookmarked);
-            expect(storeService.lastVerifiedDate).toEqual(aSearchService.lastVerifiedDate);
-
+            expect(storeService).toEqual(aSearchService);
```
rasmus-storjohann-PG, line 183: A bit of advice when changing a test like this: Keep in mind that a common bug that can happen in tests is that tests pass when they ought to fail. A very difficult bug to find. So it's a good habit when making this kind of change to break the production code to make sure that the test works, i.e. that it fails when it ought to. Did you try that?

src/validation/organizations/__tests__/validation.tests.ts
```diff
@@ -0,0 +1,62 @@
+// tslint:disable:no-expression-statement
+import { validateOrganizationResponse } from '..';
+import { aString } from '../../../application/helpers/random_test_values';
+
+describe('Organization response validation', () => {
+    describe('with valid data', () => {
+        it('returns alert', () => {
```
rasmus-storjohann-PG, line 7: alert? Seems like a copy-paste error to me

src/validation/organizations/__tests__/validation.tests.ts
```diff
@@ -0,0 +1,62 @@
...
+            const validOrganizationResponse: ReadonlyArray<any> = [{
+                id: aString(),
+                name: aString(),
+                description: aString(),
+                website: aString(),
+                email: aString(),
```
rasmus-storjohann-PG, line 13: We have generators for these, aWebUrl() or something

src/validation/organizations/__tests__/validation.tests.ts
```diff
@@ -0,0 +1,62 @@
...
+            }];
+            const validator = validateOrganizationResponse(validOrganizationResponse);
+            expect(validator.isValid).toBe(true);
+            expect(validator.validData).toBe(validOrganizationResponse);
+        });
+        it('returns empty array', () => {
```
rasmus-storjohann-PG, line 19: it accepts empty array?

src/validation/organizations/__tests__/validation.tests.ts
tomy-pg, line 7: do you mean that it `returns a valid organization`?

tomy-pg (changes requested):

rasmus-storjohann-PG (changes requested):
Just some small things to fix

src/validation/organizations/__tests__/validation.tests.ts
ian-lin-PG: Currently in the code base we use `aString()` for all test website and emails.

src/validation/organizations/__tests__/validation.tests.ts
ian-lin-PG: This is a defining difference between organizations and services. Currently services could be represented as a list of services (search, topic recommendations) or an individual service (service detail). 

Current issue is only about representing an organization (organization detail), down the road if we do include organization search we might have an empty array. Might be too far ahead for this PR.

src/stores/__tests__/services.test.ts
ian-lin-PG: I added in a check for the number of fields to be expected, this will fail next time if we add or remove fields to the humanServiceData object. This isn't perfect for when a commit adds in and remove same number of fields. I wasn't able to cause the test to fail due to the payload validation on the SaveServiceAction. Other suggestions are welcome.

PR #1258, opened 2020-08-28
Feature/issue 1202 improve keyboard submit ux

tomy-pg:


src/components/feedback/other_remove_service_component.tsx
```diff
@@ -137,10 +140,11 @@ export const OtherRemoveServiceComponent = (props: FeedbackOtherRemoveServicePro
                 onInputChange={setFeedback}
                 placeholder={content.placeholder}
             />
+            <AndroidKeyboardDoneButtonComponent isVisible={keyboardIsVisible}/>
```
rasmus-storjohann-PG, line 143: This seems simpler to me
```ts
<MultilineKeyboardDoneButton isVisible={isAndroid() && keyboardIsVisible} />
```

src/components/services/service_detail_component.tsx
```diff
@@ -252,8 +255,9 @@ export const ServiceDetailComponent = (props: Props): JSX.Element => {
                         />
                     </View>
                 </KeyboardAwareScrollView>
+                <AndroidKeyboardDoneButtonComponent isVisible={keyboardIsVisible}/>
```
rasmus-storjohann-PG, line 258: Same comment as above, just call `isAndroid()` here

src/components/feedback/feedback_component.tsx
```diff
@@ -122,23 +120,19 @@ const TextInputComponent = (props: InputComponentProps): JSX.Element => {
...
-                    multiline={true}
-                    onChangeText={onChangeText}
+                <MultilineTextInputComponent
+                    i18n={i18n}
                     value={props.inputField.value}
-                    textAlignVertical={'top'}
-                    placeholder={i18n._(t`Comment or suggest edits`)}
-                    onFocus={onFocus}
-                    onBlur={onBlur}
+                    numberOfLines={1}
```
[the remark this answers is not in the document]
tomy-pg: This is a multiline text input. The reason why `numberOfLines` is set to 1 is to follow this [mock](https://app.zeplin.io/project/5bc7bfa42e72260acb03aa86/screen/5e0fe741ac79931c869ce48c).

It is my understanding that we want the fields in `Change name or other details` to be multiline so that users have the option to use the return key on their keyboards to format their feedback even though the text input appears to only have 1 line.

src/components/multiline_text_input_for_platform.tsx
```diff
@@ -7,7 +7,7 @@ import { isAndroid } from '../application/helpers/is_android';
 import { useKeyboardIsVisible } from './use_keyboard_is_visible';
 import { EmptyComponent } from './empty_component/empty_component';
 
-export interface MultilineTextInputProps {
+export interface MultilineTextInputForPlatformProps {
```
rasmus-storjohann-PG, line 10: This doesn't seem to need to be exported. If you remove the export, you can rename it to just `Props` which would be nice.

src/components/feedback/other_remove_service_component.tsx
```diff
@@ -155,10 +154,6 @@ export const OtherRemoveServiceComponent = (props: FeedbackOtherRemoveServicePro
...
-    if (isAndroid()) {
-        return <MultilineKeyboardDoneButton isVisible={isVisible} />;
-    }
-
-    return <EmptyComponent />;
-};
\ No newline at end of file
+export const AndroidKeyboardDoneButtonComponent = ({ isVisible }: MultilineKeyboardDoneButtonProps): JSX.Element => (
+    <MultilineKeyboardDoneButton isVisible={isVisible} />
+);
```
rasmus-storjohann-PG, line 159: This component does nothing and can be removed.

tomy-pg: Ah yes this does less than one thing.

tomy-pg:
This PR has been QA'd by Cherrie and the Done button over the keyboard works on her iOS device.

PR #1273, opened 2020-09-09
Feature/issue 1262 feedback back button sends users to wrong screen

tomy-pg:


src/application/__tests__/routing.test.ts
```diff
@@ -118,3 +118,18 @@ describe('the goBack function', () => {
...
+    const serviceDetailPath = routePathDefinition(Routes.ServiceDetail);
+    const otherFeedbackPath = routePathDefinition(Routes.OtherFeedback);
+    const contactInformationPath = routePathDefinition(Routes.ContactInformation);
+
+    it('Clears the feedback paths from the history stack', () => {
+        const initialPathEntries = [servicesPath, serviceDetailPath, otherFeedbackPath, contactInformationPath];
+        const history = createMemoryHistory({ initialEntries: initialPathEntries });
+        clearFeedbackPathsFromHistory(history);
+        const lastPathIndex = history.entries.length - 1;
+        expect(history.entries[lastPathIndex].pathname).toBe(servicesPath);
```
rasmus-storjohann-PG, line 136: This test does not actually excercize the bug though. Keep in mind TDD, start with the test, see that it reproduces the bug and therefore fails, then fix the bug. In this case, you don't just care about the last element in the history, you care going further back, so you need to assert on that.

src/application/routing.ts
```diff
@@ -117,13 +118,31 @@ const previousPathMatchesContactInformationRoute = (memoryHistory: MemoryHistory
...
+    // tslint:disable-next-line: no-let
+    let lastPathIndex = memoryHistory.entries.length - 1;
+
+    while (pathMatchesFeedbackRoute(memoryHistory.entries[lastPathIndex])) {
+        if (lastPathIndex === 0) {
+            break;
+        }
+        memoryHistory.entries.pop();
+        lastPathIndex--;
+    }
```
rasmus-storjohann-PG, line 136: You can really mine Ramda for functions to use here, and it'll be more readable. While(pathMathces(R.lastElement(history.entries) && R.not(R.isEmpty(history.entries)) history.entries.pop()

src/application/routing.ts
```diff
@@ -117,13 +118,31 @@ const previousPathMatchesContactInformationRoute = (memoryHistory: MemoryHistory
...
+};
+
 const pathDoesNotMatchFeedbackRoute = (location: Location<LocationState>): boolean => (
     !pathMatchesAnyRoute(location.pathname, [Routes.OtherFeedback, Routes.ServiceDetail, Routes.ContactInformation])
 );
 
+const pathMatchesFeedbackRoute = (location: Location<LocationState>): boolean => (
+    pathMatchesAnyRoute(location.pathname, [Routes.OtherFeedback, Routes.ServiceDetail, Routes.ContactInformation])
```
rasmus-storjohann-PG, line 160: Hmmm, is ServiceDetail really a feedback route? I think there might be another bug lurking here.

src/application/routing.ts
```diff
@@ -135,6 +136,22 @@ export const clearFeedbackPathsFromHistory = (memoryHistory: MemoryHistory): voi
...
+        return;
+    }
+
+    const previousPathIndex = currentPathIndex - 1;
+    const previousPath = memoryHistory.entries[previousPathIndex].pathname;
+
+    if (currentPath === previousPath) {
+        memoryHistory.entries.pop();
+    }
+};
```
rasmus-storjohann-PG, line 153: I'm gonna go eat and then we can talk about this function ... it's pretty smelly as it stands.

src/application/__tests__/routing.test.ts
rasmus-storjohann-PG: R1, ServiceDetail ->  R1
R1, ServiceDetail, OtherFeedback ->  R1, ServiceDetail
R1, ServiceDetail, OtherFeedback, ContactInformation ->  R1, ServiceDetail, OtherFeedback
R1, ServiceDetail, OtherFeedback, ContactInformation, ServiceDetail ->  R1
R1, ServiceDetail, OtherFeedback, ContactInformation, ServiceDetail, OtherFeedback, ContactInformation, ServiceDetail -> R1

R1, ServiceDetail, OtherFeedback, ContactInformation, ServiceDetail, R2 -> ->  R1
R1, R2, ServiceDetail, OtherFeedback, ContactInformation, ServiceDetail, R3 -> ->  R1, R2

src/application/routing.ts
rasmus-storjohann-PG: `memoryHistory.entries = clearFeedback(memoryHistory.entries)`

PR #1277, opened 2020-09-14
Feature/issue 1276 close button adds path to history

tomy-pg:


rasmus-storjohann-PG:
Changes look good, @ian-lin-PG can you QA?

Issue #1275, opened 2020-09-11
Adding/removing a service bookmark should not scroll screen back to the top


jparlina:
When adding or removing a bookmark on a service, the screen should not scroll to the top. It's confusing for users. 

### Acceptance Criteria 
- [ ]  Select a topic
- [ ]  Click "Find related services near me"
- [ ]  Scroll down the list of services and bookmark a service

### Steps for QA: 
1. Select a topic
2. Click "Find related services near me"
3. Scroll down the list of services and bookmark a service
- [ ] The screen should stay at the service that you just bookmarked, not scroll back up to the top of the list
4. Remove the bookmark
- [ ] The screen should stay at the service that you just bookmarked, not scroll back up to the top of the list

ian-lin-PG:
The bookmarking still occasionally shift the user to a different place on the list. As @tomy-pg have discussed with me, different drag action seem to register the offset or not leading to this inconsistent behaviour. Not closing this issue yet.

rasmus-storjohann-PG:
So the solution is not perfect, but I don't think fixing this is going to be high priority now. Closing this issue, and if this comes up as a bug a new issue should be created.

PR #1286, opened 2020-09-28
Issue #1275 save scroll offset when bookmarking / unbookmarking services

tomy-pg:


src/components/services/render_service_items.tsx
```diff
@@ -22,8 +22,14 @@ export type ServiceItemInfo = ListRenderItemInfo<HumanServiceData>;
 
 export const renderServiceItems = R.curry((props: ServiceItemsProps, itemInfo: ServiceItemInfo): JSX.Element => {
     const service = itemInfo.item;
-    const onBookmark = (): BookmarkServiceAction => props.bookmarkService(service);
-    const onUnbookmark = (): UnbookmarkServiceAction => props.unbookmarkService(service);
+    const onBookmark = (): void => {
+        props.bookmarkService(service);
+        props.saveScrollOffset(props.scrollOffset);
+    };
```
rasmus-storjohann-PG, line 28: If possible, I would prefer the function return types to remain unchanged. Would it not work with
```ts
    const onBookmark = (): BookmarkServiceAction => {
        const result = props.bookmarkService(service);
        props.saveScrollOffset(props.scrollOffset);
        return result;
    };
```
or
```ts
    const onBookmark = (): BookmarkServiceAction => {
        props.saveScrollOffset(props.scrollOffset);
        return props.bookmarkService(service);
    };
```

rasmus-storjohann-PG (changes requested):

src/components/services/render_service_items.tsx
tomy-pg: Fixed.

PR #1301, opened 2020-10-21
Feature/issue 1260 org data

ian-lin-PG:


src/components/organizations/organization_detail_component.tsx
```diff
@@ -91,6 +95,13 @@ export const OrganizationDetailComponent = (props: Props): JSX.Element => {
     //   translation purposes,
     //   see: https://stackoverflow.com/questions/43113859/customise-tabs-of-native-base,
     //   this means we cannot sensibly style active tab text
+    const [organization, setOrganization]: readonly [HumanOrganizationData, (org: HumanOrganizationData)=> void] = useState(testOrganization);
+    const organizationId = props.match.params.organizationId;
+    if(organization.id != organizationId){
+        getOrganization(organizationId).then((res)=> {
+            setOrganization(res.results)});
+    }
```
tomy-pg, line 103: Instead of getting organizations this way, the api call can be done through a `useEffect` with an empty dependency if you want to get the organization only on component mount. Optionally you can use `organization.id` as a dependency if you decide this is needed as well.

src/components/organizations/organization_detail_component.tsx
```diff
@@ -166,36 +178,37 @@ const OrganizationDetailHeader = (props: OrganizationDetailHeaderProps): JSX.Ele
     );
 };
 
-const AboutTabComponent = (props: AnalyticsProps): JSX.Element => (
+const AboutTabComponent = (props: AboutTabProps ): JSX.Element => (
     <Content>
         <MarkdownBodyComponent
-            body={testOrganization.description}
+            body={props.organization.description}
```
tomy-pg, line 184: If you are replacing the test organization with organization state in this commit, I expected to see you using the organization state as opposed to the organization prop.

src/components/organizations/organization_detail_component.tsx
```diff
@@ -38,9 +40,11 @@ export interface OrganizationDetailActions {
     readonly saveOrganizationServicesOffset: (offset: number) => SaveOrganizationServicesOffsetAction;
 }
 
-interface AnalyticsProps {
+
+interface AboutTabProps {
     readonly analyticsLinkPressed: (analyticsLinkProps: AnalyticsLinkProps) => AnalyticsLinkPressedAction;
     readonly currentPathForAnalytics: string;
+    readonly organization: HumanOrganizationData;
```
tomy-pg, line 47: What will you be using this prop for?

src/api/index.ts
```diff
@@ -104,4 +104,14 @@ export async function getAlerts(locale: string): Promise<APIResponse> {
 
 export const createGetAlertUrl = (url: string, locale: string): string => (
     BuildUrl(url, { path: `v1/content/alerts/${locale}/` })
+);
+
+export async function getOrganization(organizationId: string): Promise<APIResponse> {
+    const url = createGetOrganizationtUrl(baseUrl, organizationId);
+    const response = await fetch(url);
+    return createAPIResponse(response);
```
tomy-pg, line 112: Nice. Looking to see `Ajv` validation next. (Similar to what we did with search)

tomy-pg (changes requested):

src/components/organizations/organization_detail_component.tsx
```diff
@@ -92,6 +89,18 @@ export const OrganizationDetailComponent = (props: Props): JSX.Element => {
     //   translation purposes,
     //   see: https://stackoverflow.com/questions/43113859/customise-tabs-of-native-base,
     //   this means we cannot sensibly style active tab text
+    const [organization, setOrganization]: readonly [HumanOrganizationData, (org: HumanOrganizationData)=> void] = useState(undefined);
```
rasmus-storjohann-PG, line 92: Make sure the comment above doesn't get disconnected from whatever it was commenting on

src/components/organizations/organization_detail_component.tsx
```diff
@@ -92,6 +89,18 @@ export const OrganizationDetailComponent = (props: Props): JSX.Element => {
...
+    const organizationId = props.match.params.organizationId;
+
+    useEffect(() => {
+        getOrganization(organizationId).then((res)=> {
+            setOrganization(res.results)});
+    }, []);
```
rasmus-storjohann-PG, line 98: `useEffect()` gotcha is to forget to include things in the dependency list. I think you need to include `getOrganization` and `setOrganization` here.

rasmus-storjohann-PG (changes requested):
Looking good!

src/components/organizations/organization_service_list_component.tsx
```diff
@@ -0,0 +1,71 @@
...
+    });
+
+    return (
+        <View style={{ flexDirection: 'column', backgroundColor: colors.lightGrey, flex: 1 }}>
+            <FlatList
+                style={{ backgroundColor: colors.lightGrey, flex: 1  }}
+                data={props.services}
+                keyExtractor={keyExtractor}
+                renderItem={renderSearchHit({
+                    ...props
```
rasmus-storjohann-PG, line 61: I would prefer that you spelled out the props being passed here rather than using `...`, it's only two I think...

rasmus-storjohann-PG (changes requested):
Lots of good work here. I have some suggestions for improvements. Make sure your linter is working and using the rules from the repo.

src/stores/organization/action.ts
```diff
@@ -0,0 +1,29 @@
+import { HumanOrganizationData } from '../../validation/organizations/types';
```
rasmus-storjohann-PG, line 1: Rename file to plural "actions.ts"

src/validation/organizations/types.ts
```diff
@@ -7,3 +7,12 @@ export interface HumanOrganizationData {
...
     readonly email: string;
 }
+
+export interface OrganizationMap {
+    readonly [organizationId: string]: HumanOrganizationData;
+}
+
+export interface OrganizationStore {
+    readonly organizations: OrganizationMap;
+    readonly organizationTab: number;
```
rasmus-storjohann-PG, line 17: The tab should actually go in the user experience store, not the organization store.

src/validation/search/__tests__/helpers/search_schema.ts
```diff
@@ -20,3 +20,14 @@ export const aGeoLocation = (): any => ({
     lat: aNumber(),
     lng: aNumber(),
 });
+
+export const anSearchService = (): any => ({
```
rasmus-storjohann-PG, line 24: typo: "aSearchService"

src/validation/services/types.ts
```diff
@@ -70,4 +74,5 @@ export interface ServiceMap {
 export interface ServiceStore {
     readonly services: ServiceMap;
     readonly servicesByTopic: ServicesForAllTopics;
+    readonly servicesByOrganization: ServicesForOrganization;
```
rasmus-storjohann-PG, line 77: Inconsistent naming, "by" vs. "for", I prefer "for".

rasmus-storjohann-PG (changes requested):
Hey @ian-lin-PG this is really really good work. I had a lot of fun reviewing this PR. I do suggest some changes, but almost all of them should be pretty trivial. The ones that you think would be time consuming to fix, let's discuss. Well done.

src/components/organizations/organization_detail_component.tsx
ian-lin-PG: I realized the organizationIds will always match since they are both derived from the same location, so I deleted this logic.

src/components/organizations/organization_detail_component.tsx
ian-lin-PG: Removed the logic once we replaced the native-base tab logic.

src/components/organizations/organization_detail_component.tsx
ian-lin-PG: Avoided the useEffect here entirely and have the API call through saga instead.

src/components/loading_screen/loading_service_list_component.tsx
```diff
@@ -15,6 +15,7 @@ export const LoadingServiceListComponent = (props: LoadingServiceListComponentPr
             <LoadingServiceListItemComponent />
             <LoadingServiceListItemComponent />
             <LoadingServiceListItemComponent />
+            <LoadingServiceListItemComponent />
```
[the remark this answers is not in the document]
ian-lin-PG: The loading screen on organization page had white space at the bottom. This is because most other loading screen thus far cover part of the page instead of the entire page. I am open to suggestion such that the loading screen looks less odd.

src/components/services/service_detail_component.tsx
```diff
@@ -168,104 +170,109 @@ export const ServiceDetailComponent = (props: Props): JSX.Element => {
...
-                            setText={setTextForField('name')}
-                            toggleShouldSend={toggleShouldSendForField('name')}
-                            inputField={feedbackInput.name}
-                            label={<Trans>Name</Trans>}
-                            body={props.service.name}
+                (({ i18n }: { readonly i18n: I18n }): JSX.Element =>
+                    <View style={{ flex: 1 }}>
+                        <ServiceDetailHeaderComponent
+                            location={props.location}
+                            history={props.history}
```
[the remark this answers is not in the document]
ian-lin-PG: I think I added the open organization action that starts the cascade of API calls.

src/components/organizations/organization_service_list_component.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+export const OrganizationServiceListComponent = (props: Props): JSX.Element => {
+    const [orgServicesOffset, setOrgServiesOffset]: readonly [number, (n: number) => void] = useState(props.organizationServicesOffset);
+    const flatListRef = useRef<FlatList<HumanServiceData>>();
+
+    useEffect((): void => {
+        if (props.services.length > 0) {
+            flatListRef.current.scrollToOffset({ animated: false, offset: props.organizationServicesOffset });
+        }
+    }, [props.organizationServicesOffset, props.services, flatListRef]);
```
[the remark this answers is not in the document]
ian-lin-PG: We use the same pattern for all the service list offsets now.

src/components/organizations/organization_detail_component.tsx
```diff
@@ -1,252 +1,54 @@
...
-};
-
-const AboutTabComponent = (props: AnalyticsProps): JSX.Element => (
+export const OrganizationDetailComponent = (props: OrganizationDetailProps): JSX.Element => (
     <Content>
         <MarkdownBodyComponent
-            body={testOrganization.description}
+            body={props.organization.description}
             shouldBeExpandable={true}
             // TODO Issue #1080 When organization detail page is online connect the following 2 states to the store/persisted data
```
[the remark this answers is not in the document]
ian-lin-PG: I been thinking about this. Technically only topic detail are in Markdown format which has the link out functionality. For services and organization description that are usually plain test the link alert and hide link alert are never used.

src/sagas/services.ts
```diff
@@ -10,11 +10,19 @@ import { Errors } from '../validation/errors/types';
...
+import { toHumanServiceData } from '../validation/search/to_human_service_data';
+import { OpenOrganizationAction } from '../stores/organization/action';
 
 export function* watchUpdateServicesForTopic(): IterableIterator<ForkEffect> {
     yield takeLatest(constants.LOAD_SERVICES_REQUEST, updateServicesForTopic);
 }
 
+export function* watchUpdateServicesForOrganization(): IterableIterator<ForkEffect> {
+    yield takeLatest(constants.OPEN_ORGANIZATION, updateServicesForOrganization);
+}
```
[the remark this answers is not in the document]
ian-lin-PG: That is true, even if users are unlikely to notice the extra network requests. Checking the organization store before triggering the action might be too much overhead. Currently don't see a beautiful solution...

src/components/organizations/org_tab_swticher.tsx
```diff
@@ -0,0 +1,84 @@
...
+}
+export interface NavigationStateRoute {
+    readonly navigationState: NavigationState<Route>;
+}
+
+type TabBarProps = SceneRendererProps & NavigationStateRoute;
+
+const renderTabBar = (tabBarProps: TabBarProps): JSX.Element => (
+    <TabBar
+        {...tabBarProps}
```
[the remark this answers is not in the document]
ian-lin-PG: `type TabBarProps = SceneRendererProps & NavigationStateRoute;` 
`import { TabView, Route, NavigationState, SceneRendererProps, TabBar } from 'react-native-tab-view';`
The tabBarProps are actually references to the library, I think we only override the library with the few lines underneath.

src/components/search/api/fetch_search_results_from_query.ts
```diff
@@ -50,6 +50,32 @@ export const fetchSearchResultsFromQuery = async (
...
+            body: JSON.stringify({
+                query: organizationId,
+            }),
+        });
+
+        const responseJSON: AlgoliaResponse = await response.json();
+        return responseJSON;
+    } catch (Error) {
+        return undefined;
+    }
```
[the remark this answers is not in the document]
ian-lin-PG: Created a helper to build the request.

src/stores/__tests__/services.test.ts
```diff
@@ -167,6 +168,24 @@ describe('services reducer', () => {
...
+                type: constants.SAVE_SERVICES_BY_ORGANIZATION,
+                payload: {
+                    organizationId: organizationId,
+                    services: [anOrganizationService]
+                },
+            };
+            const store = reducer(theStore, action);
+            const storeService = store.services[anOrganizationService.id];
+            expect(storeService).toEqual(anOrganizationService);
+        });
```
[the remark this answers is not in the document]
ian-lin-PG: The organizationId is tested in `organization.test.ts` file, I do agree it's quite odd to have the test separated like this.

src/components/organizations/org_tab_swticher.tsx
```diff
@@ -0,0 +1,88 @@
...
+type Props = OrgTabSwitcherProps & OrganizationProps & OrganizationActions;
+
+export default function OrgTabSwitcher(props: Props) {
+    const [index, setIndex] = React.useState(0);
+    const routes: TabRoutes = [
+        { key: 'about', title: props.i18n._(t`About`) },
+        { key: 'services', title: props.i18n._(t`Services`) },
+    ];
+
+    const renderScene = ({ route }: { readonly route: Route}): JSX.Element => {
```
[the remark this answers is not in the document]
ian-lin-PG: I took a peek into the documentation, haven't been able to pass in extra parameters into the `renderScene` of the TabView component.

src/stores/__tests__/user_experience.test.ts
```diff
@@ -97,6 +97,21 @@ describe('the user experience reducer', () => {
...
+            const newStore = reducer(oldStore, Actions.saveOrganizationTab(0));
+            expect(newStore.organizationTab).toBe(0);
+        });
+
+        it ('is set to services when save organization tab is set to 1', () => {
+            const oldStore = buildDefaultStore();
+            const newStore = reducer(oldStore, Actions.saveOrganizationTab(1));
+            expect(newStore.organizationTab).toBe(1);
+        });
+    });
```
rasmus-storjohann-PG, line 113: Nice to see the tests. When you get used to writing the tests along with the code rather than afterwards, I promise you that you will enjoy writing them.

src/components/organizations/organization_service_list_component.tsx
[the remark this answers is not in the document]
tomy-pg: This looks good

src/sagas/services.ts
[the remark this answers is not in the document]
tomy-pg: You can use redux-saga's `select()` here if you want. This is not always encouraged to do as it is preferred that saga's be autonomous.

However, I see that in `updateServicesOrganization` you are using this already for `bookmarkedServicesIds`. 
So you can also use a services for organization selector to make the check if the service already exists before making the network request.

src/components/organizations/organization_component.tsx
```diff
@@ -0,0 +1,110 @@
...
+    return (
+        <I18n>
+            {({ i18n }: { readonly i18n: I18n }): JSX.Element => (
+                <View style={{ flex: 1 }}>
+                    <OrganizationHeader
+                        location={props.location}
+                        history={props.history}
+                        openHeaderMenu={props.openHeaderMenu}
+                    />
+                    <View padder style={{ flex: 1 }}>
```
tomy-pg, line 61: This `View` affects how the UX of the services list, especially when the services list is not fully populated. It makes it such that there are two different scrolling views on the page and the service card is background and dividers are not completely extended to the full width of the screen. 

Screen Shot 2020-12-07 at 11 22 31 AM


I think you can remove this `View` and add padding to the different components used here.

src/components/organizations/organization_service_list_component.tsx
```diff
@@ -0,0 +1,59 @@
...
+            renderItem={renderServiceItems({
+                history: props.history,
+                scrollOffset: organizationServicesOffset,
+                bookmarkService: props.bookmarkService,
+                unbookmarkService: props.unbookmarkService,
+                openServiceDetail: props.openServiceDetail,
+                saveScrollOffset: props.saveOrganizationServicesOffset,
+            })}
+            ItemSeparatorComponent={SearchListSeparator}
+            ListHeaderComponent={<View />}
```
tomy-pg, line 56: `Flatlist's` `initialNumToRender` property needs to be declared. The UX for when a user clicks on a service detail and then clicks back to the service list will not work properly without this. 

This is an optimization by React-Native that we have to override since by default, they set the initial batch of items to render to be 20. For our case, if a user navigates to any service detail where its position on the list is beyond the initial 20 and then wants to go back to the service list, their previous scroll-offset will not exist. 

`initialNumToRender={props.organizationServicesOffset ? services.length : 20}`

tomy-pg (changes requested):

src/components/organizations/organization_component.tsx
```diff
@@ -0,0 +1,108 @@
...
+    return (
+        <I18n>
+            {({ i18n }: { readonly i18n: I18n }): JSX.Element => (
+                <View style={{ flex: 1 }}>
+                    <OrganizationHeader
+                        location={props.location}
+                        history={props.history}
+                        openHeaderMenu={props.openHeaderMenu}
+                    />
+                    <DescriptorComponent descriptor={<Trans>ORGANIZATION</Trans>} />
```
tomy-pg, line 61: for the components in this file starting here, more horizontal padding will need to be added since `padder` was removed and the native-base `View` is no longer used. 

As well, watch the alignment of the Title components and the service detail components. See mocks [here](https://app.zeplin.io/project/5bc7bfa42e72260acb03aa86/screen/5d929aa5a78fc061803c87ca)

tomy-pg (changes requested):

tomy-pg (approved):

PR #1315, opened 2020-11-19
Issue pg-irc/pathways-backend#574 Limit geocoder requests to British …

rasmus-storjohann-PG:
…Columbia

### DIFF
--- a/src/api/fetch_lat_long_from_location.ts
+++ b/src/api/fetch_lat_long_from_location.ts
@@ -57,6 +57,7 @@ export const buildGeoCoderUrl = (location: string): string => (
     BuildUrl('https://geocoder.ca', {
         queryParams: {
             locate: location,
+            region: 'BC',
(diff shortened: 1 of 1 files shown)


src/api/fetch_lat_long_from_location.ts
```diff
@@ -57,6 +57,7 @@ export const buildGeoCoderUrl = (location: string): string => (
     BuildUrl('https://geocoder.ca', {
         queryParams: {
             locate: location,
+            region: 'BC',
```
tomy-pg, line 60: From initial QA, I see that Abbotsford is working, but Delta and North Vancouver are not. 
I tried adding `country: 'Canada'`, which fixed North Vancouver but did not fix Delta.
The fix needed for Delta is probably appending ' BC' to `locate`.

tomy-pg (changes requested):

PR #1317, opened 2020-11-25
Feature/issue 574 backend

rasmus-storjohann-PG:


src/api/fetch_lat_long_from_location.ts
```diff
@@ -57,6 +57,7 @@ export const buildGeoCoderUrl = (location: string): string => (
     BuildUrl('https://geocoder.ca', {
         queryParams: {
             locate: location,
+            prov: 'bc',
```
tomy-pg, line 60: Looks good.

From GeoCoder's API documentation:
`The other parameters (addresst, stno, city, etc) have been deprecated but will continue to be supported for backwards compatibility.`

tomy-pg (approved):

Issue #1318, opened 2020-12-08
Expo splash screen appears where our own should be


rasmus-storjohann-PG:
Booting up the app, we now see an ugly low res Expo splash screen. This is a regression very likely caused by a recent expo library upgrade.

### Acceptance Criteria 
- [x]  At boot up, our own splash screen appears.

### Steps for QA: 
1.

rasmus-storjohann-PG:
@cherrielam please give a link to the splash screen we should use (just an image, no buttons, language selectors, etc)

tomy-pg:
This has regressed again in staging release 1.58.0

tomy-pg:
https://forums.expo.io/t/after-expo-upgrade-appear-expo-splash-in-my-app/46048

After some sleuthing, it seems like upgrading `expo-cli` fixes the problem for some people. I upgraded and side-loaded the app and I now see the splash screen and icon fixed. `expo-cli` apparently is not supposed to affect any of this but I along with some others in the expo community have seen this fixed by upgrading. 

@rasmus-storjohann-PG and @ian-lin-PG please run `yarn global add expo-cli` before next release.

ian-lin-PG:
PASS QA 1.62.0

PR #1322, opened 2020-12-10
Issue #1318 replace expo splash with image

tomy-pg:
We can also add an image to the splash screen with the `image` property. We will need a properly sized image for this following this guide: https://docs.expo.io/guides/splash-screens/


rasmus-storjohann-PG:
We have a splash screen image https://github.com/pg-irc/pathways-frontend/blob/develop/assets/images/landing_photo.png but it used to have some text over it, it's all still in the code, see `WelcomeComponent`

rasmus-storjohann-PG:
Hmmm, looks like we need to make an image rather than a component https://docs.expo.io/guides/splash-screens/

PR #1347, opened 2021-01-16
Feature/issue 1323 organization page handle offline

ian-lin-PG:


rasmus-storjohann-PG (changes requested):
Really nice work @ian-lin-PG , we can discuss some of the changes I suggest here, especially around the valid/invalid store.

src/validation/organizations/types.ts
```diff
@@ -1,17 +1,34 @@
...
+}
+
+export interface LoadingOrganization {
+    readonly type: 'ORGANIZATION:LOADING';
+}
+
+export interface ErrorForOrganization {
+    readonly type: 'ORGANIZATION:ERROR';
+    readonly errorMessageType: Errors;
+}
```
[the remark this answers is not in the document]
ian-lin-PG: The validation types.ts doesn't seem to accept references for their type definitions. See also `/Users/ianlin/Desktop/PeaceGeeks/pathways-frontend/src/validation/services/types.ts`. Trying to change the type from strings causes the builder to fail as well.

src/application/constants.ts
```diff
@@ -93,6 +93,10 @@ export const SAVE_ORGANIZATION = 'SAVE_ORGANIZATION';
 export const SAVE_ORGANIZATION_TAB = 'SAVE_ORGANIZATION_TAB';
 export const SAVE_SERVICES_BY_ORGANIZATION = 'SAVE_SERVICES_BY_ORGANIZATION';
 
+export const LOAD_ORGANIZATION_REQUEST = 'LOAD_ORGANIZATION_REQUEST';
+export const LOAD_ORGANIZATION_SUCCESS = 'LOAD_ORGANIZATION_SUCCESS';
+export const LOAD_ORGANIZATION_FAILURE = 'LOAD_ORGANIZATION_FAILURE';
```
rasmus-storjohann-PG, line 98: Yes this is better

src/stores/__tests__/organization.test.ts
```diff
@@ -3,18 +3,29 @@ import * as constants from '../../application/constants';
...
+                type: constants.LOAD_ORGANIZATION_REQUEST,
+            };
+            const store = reducer(theStore, action);
+            expect(store.organizationStatus.type).toEqual(constants.LOADING_ORGANIZATION);
+        });
 
+    });
+
+    describe('when load organization is successful', () => {
         it('updates the store with the new organization in addition to old Organization', () => {
```
rasmus-storjohann-PG, line 25: The name of the test indicates that you will test for the old org being there as well, but there is no assert for that

src/stores/__tests__/organization.test.ts
```diff
@@ -24,6 +35,22 @@ describe('organizations reducer', () => {
             const storeOldOrganization = store.organizations[oldOrganizationBuilder.id];
             expect(storeNewOrganization).toEqual(newOrganization);
             expect(storeOldOrganization).toEqual(oldOrganizationBuilder.build());
```
rasmus-storjohann-PG, line 37: You're testing the behaviour of the unit under test, which is the reducer. Here you are asserting on what went into the reducer as input, that is not relevant to the test.

src/stores/__tests__/organization.test.ts
```diff
@@ -24,6 +35,22 @@ describe('organizations reducer', () => {
...
+            expect(store.organizationStatus.type).toEqual(constants.VALID_ORGANIZATION);
+        });
+    });
+
+    describe('when load organization ends in error', () => {
+        it('updates the store with the new organization in addition to old Organization', () => {
```
rasmus-storjohann-PG, line 43: wrong test name, copy paste error probably

src/stores/__tests__/organization.test.ts
```diff
@@ -24,6 +35,22 @@ describe('organizations reducer', () => {
...
+            const newOrganization = new HumanOrganizationDataBuilder().build();
+            const action: ErrorOrganizationAction = {
+                type: constants.LOAD_ORGANIZATION_FAILURE,
+                payload: {
+                    organizationId: newOrganization.id,
+                    errorMessageType: Errors.Offline,
+                },
+            };
+            const store = reducer(theStore, action);
+            expect(store.organizationStatus.type).toEqual(constants.ERROR_ORGANIZATION);
```
rasmus-storjohann-PG, line 53: Need a test for the error state being added to the store too, I think. Note that if you did actual TDD, you would naturally write all the necessary tests, and you would probably do so in less time.

src/stores/organization/actions.ts
```diff
@@ -4,29 +4,28 @@ import * as constants from '../../application/constants';
...
-export const errorOrganization = (organizationId: string, errorMessageType: Errors) => (
-    helpers.makeAction(constants.ERROR_ORGANIZATION, { organizationId, errorMessageType })
+export const saveOrganization = (organization: HumanOrganizationData) => (
+    helpers.makeAction(constants.LOAD_ORGANIZATION_SUCCESS, { organization })
 );
 
 // tslint:disable-next-line:typedef
-export const loadingOrganization = (organizationId: string) => (
-    helpers.makeAction(constants.LOADING_ORGANIZATION, { organizationId })
+export const errorOrganization = (organizationId: string, errorMessageType: Errors) => (
```
rasmus-storjohann-PG, line 27: Model your names on how you would talk about this to another human. "error organization" does not fit in a normal English sentence. How about something like "loading organization failed" or "organization loading failure", better? Same with the other names.

rasmus-storjohann-PG (changes requested):
Just minor issues with naming and such, coming together nicely

Issue #1344, opened 2021-01-14
Build client UX for Review this service


tomy-pg:
[Mocks are in Zeplin](https://zpl.io/aMy3Gdo), also available on [Figma (under FOR PRODUCTION - Service Feedback)](https://www.figma.com/file/0Fhv64lm9QeljUKOvaJ128/AA-Feedback-2.0?node-id=529%3A611)

Here is a **[Figma prototype](https://www.figma.com/proto/0Fhv64lm9QeljUKOvaJ128/AA-Feedback-2.0?node-id=529%3A2476&viewport=452%2C476%2C0.2726389169692993&scaling=scale-down)** to see the flow/interaction between the pages

There is **additional copy** added for "Learn more" > Services listed on Arrival Advisor page. "How is PeaceGeeks and bc211 using the information collected?" section is being updated to the following:

> **What happens when I “suggest an update” on the information listed about a service?** 
> 
> When you suggest an update on service information, PeaceGeeks collects and shares with bc211 your suggestions, as well as the name of the service you’re commenting on. We record the date and time that the suggestion was received. If you choose to provide your email address, we collect that so that PeaceGeeks and bc211 may contact you regarding your suggested change. If you work at the service provider organization, you will have the option to share your full name, job title and organization name. It is useful for us to know that the suggestion comes from someone who is familiar with the organization’s work. See also PeaceGeeks’ general [privacy policy](link) and bc211’s [privacy page](link).
> 
> **How does PeaceGeeks use my feedback on services?**
> 
> You can give feedback about a service by selecting the smiley faces at the bottom of a service page and providing an optional comment about your experience. When you submit this feedback, PeaceGeeks collects and reviews it, and then shares your feedback with the service provider organization. We also record the date and time the feedback was received. See also PeaceGeeks’ general [privacy policy](link).

### Acceptance Criteria 
- [x] `How was your experience with [this service]` (smileys) can be seen in the service detail screen
- [x] After clicking on one of the smileys, the user is navigated to a new screen where they can optionally leave a comment
- [x] When clicking on the nth smiley (n=0..3), after navigating to the new screen, the smiley is selected, i.e. the app remembers which smiley was clicked in the first screen
- [x] Clicking on the 'X' returns the user back to their previous page
- [x] Their rating is not sent to the database if they click on the 'X'
- [x] `How was your experience with [this service]` no longer appears in any service' service detail screen that they have previously reviewed.

### Steps for QA: 
[...]

PR #1349, opened 2021-01-19
Feature/issue #1344 build ux for review this service

tomy-pg:


src/components/reviews/ratings_component.tsx
```diff
@@ -0,0 +1,76 @@
...
+            R.map((icon: RatingIcon): JSX.Element => {
+                return (
+                    <RatingIconComponent
+                        key={icon.name}
+                        isChosen={icon.value === props.rating}
+                        icon={icon}
+                        onPress={(): void => props.chooseRating(icon.value)}
+                    />
+                );
+            }, props.iconsList)}
```
rasmus-storjohann-PG, line 58: Looking to create some consistency on how we write functions that just do a return, let's stick with the shorter form using `()` and no return:
```ts
            R.map((icon: RatingIcon): JSX.Element => (
                    <RatingIconComponent
                        key={icon.name}
                        isChosen={icon.value === props.rating}
                        icon={icon}
                        onPress={(): void => props.chooseRating(icon.value)}
                    />
                ), props.iconsList)}
```

src/components/reviews/service_review_component.tsx
```diff
@@ -0,0 +1,139 @@
...
+            <SubmitFeedbackButton
+                isVisible={!keyboardIsVisible}
+                disabled={false}
+                onPress={onSubmitButtonPress}
+            />
+            <DiscardChangesModal
+                isVisible={props.showDiscardChangesModal}
+                onKeepEditingPress={props.closeDiscardChangesModal}
+                onDiscardPress={onDiscardPress}
+            />
```
rasmus-storjohann-PG, line 75: The copy on this modal should ideally be adapted to the review function. Not bad enough to break the string freeze, but probably create an issue for it so we don't forget. Currently is says "Discard the changes you suggested?" which is appropriate for feedback but not review.

src/components/reviews/service_review_component.tsx
```diff
@@ -0,0 +1,139 @@
...
+                extraHeight={100}
+                enableOnAndroid={true}
+                extraScrollHeight={100}
+                style={{ padding: 20 }}
+            >
+                <ServiceNameComponent name={props.serviceName} onPress={memoryHistory.goBack} />
+                <RatingQuestionComponent />
+                <RatingsComponent rating={props.rating} chooseRating={props.chooseRating}/>
+                <CommentComponent comment={comment} setComment={setComment}/>
+            </KeyboardAwareScrollView>
```
rasmus-storjohann-PG, line 64: Slight UX glitch where it doesn't look like it can scroll all the way down ... I think we've already seen this elsewhere and it can't be fixed, IIRC.

rasmus-storjohann-PG (changes requested):
Really good work, and quick! You have over-learned the thing about small commits though, there are too many!

src/components/reviews/service_review_component.tsx
tomy-pg: Issue has been created, see: https://github.com/pg-irc/pathways-frontend/issues/1352

src/components/reviews/service_review_component.tsx
tomy-pg: I can give this another look over. I recall this bug from the work we did with `Feedback` inputs.

src/application/routing.ts
```diff
@@ -121,6 +124,22 @@ const pathMatchesFeedbackRoute = (location: Location<LocationState>): boolean =>
...
+export const backToServiceDetailOnServiceReviewDiscard = (memoryHistory: MemoryHistory): void => {
+    popServiceReviewPathFromHistory(memoryHistory);
+    const positionOfCurrentPathInHistoryStack = 0;
+    memoryHistory.go(positionOfCurrentPathInHistoryStack);
+};
+
+export const popServiceReviewPathFromHistory = (memoryHistory: MemoryHistory): void => {
+    while (R.findLast(pathMatchesServiceReviewRoute, memoryHistory.entries)) {
+        memoryHistory.entries.pop();
+    }
```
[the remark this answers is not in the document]
tomy-pg: Great find. Fixed this for `Feedback` as well.

src/application/routing.ts
```diff
@@ -115,7 +115,7 @@ export const backToServiceDetailOnFeedbackSubmit = (memoryHistory: MemoryHistory
 };
 
 export const popFeedbackPathsFromHistory = (memoryHistory: MemoryHistory): void => {
-    while (R.findLast(pathMatchesFeedbackRoute, memoryHistory.entries)) {
+    while (pathMatchesFeedbackRoute(R.last(memoryHistory.entries))) {
```
rasmus-storjohann-PG, line 118: yup

PR #1354, opened 2021-01-26
Feature/issue 1345 wire-up airtable for service reviews

tomy-pg:
Please see https://github.com/PeaceGeeksSociety/pathways-deploy/pull/5 for env variable changes.


src/sagas/reviews.ts
```diff
@@ -0,0 +1,93 @@
...
+const buildRequestUrl = (): string => {
+    if (!AIRTABLE_BASE_ID) {
+        throw new Error('AIRTABLE_BASE_ID is missing.');
+    }
+
+    if (!AIRTABLE_TABLE_SERVICES_REVIEWS_ID) {
+        throw new Error('AIRTABLE_TABLE_SERVICES_REVIEWS_ID is missing.');
+    }
+
+    return buildUrl('https://api.airtable.com', { path: `v0/${AIRTABLE_BASE_ID}/${AIRTABLE_TABLE_SERVICES_REVIEWS_ID}` });
```
rasmus-storjohann-PG, line 63: Just be aware of the issues raised here https://stackoverflow.com/a/3369494

src/sagas/reviews.ts
```diff
@@ -0,0 +1,93 @@
...
+};
+
+const buildRequestInit = (serviceReviewJSON: string): RequestInit => {
```
rasmus-storjohann-PG, line 66: `RequestInit` is a bit of a weird name. It's an abbreviation for one. How about ReviewPostRequest?

src/selectors/locale/select_locale_code.ts
```diff
@@ -0,0 +1,8 @@
+import { Store } from '../../stores';
+import { pickLocaleStore } from './pick_locale_store';
+import { toSelectorLocale } from './to_selector_locale';
+
+export const selectLocaleCode = (appStore: Store): string => {
+    const store = pickLocaleStore(appStore);
+    return toSelectorLocale(store).code;
```
rasmus-storjohann-PG, line 7: You can just do `return store.code` here I think

src/sagas/reviews.ts
```diff
@@ -30,6 +30,7 @@ function* submitServiceReview(action: SubmitServiceReviewAction): IterableIterat
         console.warn(error?.message || 'Error saving review');
     } finally {
         yield put(setIsSendingReview(false));
+        yield put(clearReview());
```
rasmus-storjohann-PG: Do you really need these two actions? Redux is great for expressing situations where one action should have more than one consequence in different parts of the store. It feels like that's what's really going on here, that at the end of this flow, you want to dispatch one action that will cause two distinct changes to the store. The code here should just worry about the actions to emit, the reducers are then responsible for responding to those actions, thereby separating those concerns.

src/components/reviews/service_review_component.tsx
```diff
@@ -49,7 +49,10 @@ export const ServiceReviewComponent = (props: Props): JSX.Element => {
 
     const onSubmitButtonPress = (): void => {
         props.submitServiceReview(props.serviceId, comment);
+        // TODO rename this to backToServiceDetailFromServiceReview
+        backToServiceDetailOnServiceReviewDiscard(memoryHistory);
```
rasmus-storjohann-PG: I would prefer that you do the rename in the same commit. I would not prefer that if the function you're renaming was used in lots and lots of places, since that would make the other changes in the same commit hard to see, but I don't believe that is the case here.

src/components/services/service_detail_component.tsx
```diff
@@ -282,7 +283,7 @@ export const ServiceDetailComponent = (props: Props): JSX.Element => {
                             onPress={onSubmitPress}
                         />
                         <ThankYouMessageOrEmptyComponent
-                            isSendingFeedback={props.isSendingFeedback}
+                            isSendingFeedbackOrReview={props.isSendingFeedback || props.isSendingReview}
```
rasmus-storjohann-PG, line 286: So the question here is, should this component know about the fact that there are two different things that might be in the process of being sent? If the component should not know, when who should? Would it make sense to have a selector that knows about the details of the store (like all selectors do) and permit the connected component and this code to just worry about "something" being sent?

src/sagas/reviews.ts
```diff
@@ -27,6 +28,7 @@ function* submitServiceReview(action: SubmitServiceReviewAction): IterableIterat
         const response = yield call(sendToAirtable, serviceReviewJSON);
         yield call(handleResponse, response);
     } catch (error) {
+        Sentry.captureException(error);
```
rasmus-storjohann-PG, line 31: good call, I haven't had Sentry on my horizon enough

rasmus-storjohann-PG (changes requested):
Nice work, just a few comments and requested changes

rasmus-storjohann-PG:
I've reviewed the deploy PR and it's ready to go in. Suggested a name change, but it's optional, not a big deal.

src/sagas/reviews.ts
tomy-pg: I am happy to make this change. I originally named it after the interface the function returns.

src/components/reviews/service_review_component.tsx
tomy-pg: Okay that makes sense!

src/sagas/reviews.ts
tomy-pg: Okay, I am happy to do this. I originally named this function based on the interface being returned.

PR #1357, opened 2021-02-01
Feature/issue 1348 rate this service and banner

tomy-pg:


src/components/services/service_detail_component.tsx
```diff
@@ -531,4 +531,8 @@ const getAddressesString = (addresses: ReadonlyArray<Address>): string => (
 
 const buildPhonesString = (phones: ReadonlyArray<PhoneNumber>): string => (
     phones.map((phone: PhoneNumber): string => `${phone.type}: ${phone.phone_number}`).join('\n')
+);
+
+const showDividerComponentOrEmpty = (isVisible: boolean): JSX.Element => (
+    isVisible ? <DividerComponent /> : <EmptyComponent />
```
rasmus-storjohann-PG, line 537: `DividerComponent` is ours, you can add an `isVisible` prop on it that defaults to true so you don't have to touch all it's call sites.

src/components/bookmarks/service_bookmarks_component.tsx
```diff
@@ -15,11 +15,14 @@ import { setServicesOffsetThrottled } from '../set_services_offset_throttled';
 import { ServiceBanner } from './service_banner';
 import { EmptyComponent } from '../empty_component/empty_component';
 import { scrollToOffsetWithTimeout } from '../scroll_to_offset_with_timeout';
+import { ThankYouMessageOrEmptyComponent } from '../services/thank_you_message_or_empty_component';
```
rasmus-storjohann-PG, line 18: It's pretty natural for a component to have an isVisible property without that having to be represented in the name of the component, so I would prefer the simpler name of `ThankYouMessage` here.

src/components/scroll_to_offset_with_timeout.ts
```diff
@@ -0,0 +1,10 @@
+// tslint:disable: no-expression-statement
+import { FlatList } from 'react-native';
+
+export const scrollToOffsetWithTimeout = (flatListRef: React.MutableRefObject<FlatList>, offset: number): void => {
```
rasmus-storjohann-PG, line 4: How about if we consider the fact that this function uses a timeout to be an implementational detail that the caller has no need to know about? In that case, this function should just be called `scrollToOffset()`, which is nicer

src/components/search/search_results_component.tsx
```diff
@@ -84,11 +86,13 @@ const renderComponentWithResults = (props: Props): JSX.Element => {
...
         }
-    }, [props.searchTerm, props.searchLocation]);
+    }, [props.searchTerm, props.searchLocation, flatListRef]);
 
     useEffect((): void => {
         if (props.searchResults.length > 0) {
-            flatListRef.current.getNode().scrollToOffset({ animated: false, offset: props.searchOffset });
+            setTimeout((): void => {
+                flatListRef.current.getNode().scrollToOffset({ animated: false, offset: props.searchOffset });
+              }, 0);
```
rasmus-storjohann-PG, line 95: We should use the helper function that you created here as well, no?

src/components/search/search_results_component.tsx
```diff
@@ -155,6 +164,11 @@ const renderSearchHit = R.curry((props: SearchHitProps, itemInfo: ListRenderItem
 
     const onBookmark = (): BookmarkServiceAction => props.bookmarkService(service);
     const onUnbookmark = (): UnbookmarkServiceAction => props.unbookmarkService(service);
+    const onPressServiceReview = (): void => {
+        props.saveService(service);
```
rasmus-storjohann-PG, line 168: Here I find myself thinking, what do you mean, save service? Does it end up being persisted in the device, or just the id tucked away somewhere? This function can probably just be renamed to fix this problem.

rasmus-storjohann-PG (changes requested):
Really good work, readable code.

src/components/bookmarks/service_bookmarks_component.tsx
rasmus-storjohann-PG: So the improved DividerComponent also does not need to be renamed

src/components/reviews/service_review_component.tsx
```diff
@@ -142,4 +144,21 @@ const CommentComponent = (props: CommentProps): JSX.Element => (
...
         }
     </I18n>
-);
\ No newline at end of file
+);
+
+const ExplainReviewUsageComponent = ({ history, serviceId }: { readonly history: History, readonly serviceId: Id }): JSX.Element => {
+    const onPress = (): void => {
+        goToRouteWithParameter(Routes.ExplainFeedback, serviceId, history)();
+    };
```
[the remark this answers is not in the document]
tomy-pg: I don't mind that `onPress` returns a function. This immediately invoked function isn't my favourite either `goToRouteWithParameter(Routes.ExplainFeedback, serviceId, history)();` but we use it in a lot of places .

src/components/search/search_results_component.tsx
tomy-pg: This `flatListRef` is different than the one used in the helper function.
This one refers to the animated flatlist: 
`const AnimatedFlatList = Animated.createAnimatedComponent(FlatList);` so the way in which scrollToOffset is retrieved is different since we have to call `getNode()` in this ref.

src/components/search/search_results_component.tsx
tomy-pg: Sounds good. This function is actually tied to a redux action so this name change is going to go through the different layers of react and redux

src/components/services/thank_you_message_or_empty_component.tsx
```diff
@@ -4,12 +4,12 @@ import { t } from '@lingui/macro';
 import { EmptyComponent } from '../empty_component/empty_component';
 import { showToast } from '../../application/toast';
 
-interface ThankYouMessageOrEmptyProps {
+interface ThankYouMessage {
```
rasmus-storjohann-PG, line 7: `ThankYouMessageProps`

rasmus-storjohann-PG (changes requested):
Just the one typo, otherwise this is good to go

PR #1367, opened 2021-02-08
Feature/issue 1350 topics scroll offset

tomy-pg:


src/components/topics/task_list_component.tsx
```diff
@@ -70,14 +70,24 @@ export class TaskListComponent extends React.PureComponent<Props, State> {
     }
 
     componentDidUpdate(previousProps: Props): void {
-        if (this.flatListRef && hasContentChanged(previousProps, this.props)) {
+        console.log(this.props.scrollOffset)
+        if (this.flatListRef && hasTopicDetailPageChanged(previousProps, this.props)) {
```
rasmus-storjohann-PG, line 73: Legibility will be improved if you do an early return on `this.flatListRef` being null.

How about if you had a local variable like so
```ts
const foo = {
                ...this.state,
                data: this.props.tasks,
                scrollOffset: this.state.scrollOffset,
            };
```
and then you could have this in one call
```ts
            return this.setState(
                            foo, 
                            (): void => scrollToOffsetWithTimeout(this.flatListRef, this.state.scrollOffset),
                    );
```
and this in the other
```ts
            return this.setState(
                     { ...foo, scrollOffset: 0}, 
                    (): void => scrollToOffsetWithTimeout(this.flatListRef, this.state.scrollOffset),
            );
```

src/components/topics/task_list_component.tsx
```diff
@@ -135,9 +145,13 @@ const scrollToOffsetWithTimeout = (flatListRef: FlatListRef, offset: number): vo
...
-const hasContentChanged = (previousProps: Props, props: Props): boolean => {
-    return previousProps.headerContentIdentifier !== props.headerContentIdentifier || tasksHaveChanged(previousProps, props);
-};
+const hasTopicDetailPageChanged = (previousProps: Props, props: Props): boolean => (
+    previousProps.headerContentIdentifier !== props.headerContentIdentifier
+);
+
+const hasTopicListDataChanged = (previousProps: Props, props: Props): boolean => (
+    tasksHaveChanged(previousProps, props)
+)
```
rasmus-storjohann-PG: Do you need this function? it's effectively just an alias.

rasmus-storjohann-PG (changes requested):

src/components/topics/task_list_component.tsx
tomy-pg: Makes sense

src/components/topics/task_list_component.tsx
tomy-pg: The earlier return for `this.flatListRef` makes a lot of sense. 

As for the local variable, the values inputted into this `setState` changes in the last commit to clean it up a bit.

rasmus-storjohann-PG (approved):
I like it!

Issue #1356, opened 2021-01-28
UX inconsistency: Button disabled when it should be enabled


rasmus-storjohann-PG:
Open any service on the app. Click "Suggest an update".  Click "This service no longer exists". Observe that ux says "Provide any additional information (optional)". Also observe that the Submit button is disabled until additional information field is not empty, implying that the additional feedback is required.

### Acceptance Criteria 
- [x]  The submit button is always enabled

ian-lin-PG:
PASS QA 1.62.0

PR #1370, opened 2021-02-09
Issue #1356 enable submit button when on remove this service feedback…

tomy-pg:
… screen


rasmus-storjohann-PG (approved):
The logic is a little bit hard to follow, some renames might help but I don't have specific suggestions and I'm fine with this going in as it is.

PR #1371, opened 2021-02-09
Feature/issue 1360 enable my location on search by default

ian-lin-PG:


src/components/search/search_input_component.tsx
```diff
@@ -321,11 +321,16 @@ const openAppSettings = (): void => {
...
-    Permissions.askAsync(Permissions.LOCATION).
-        then((permissionResponse: Permissions.PermissionResponse): void => {
-            if (permissionResponse.status === Permissions.PermissionStatus.GRANTED) {
-                setSearchLocationInput(MY_LOCATION);
-            }
-        });
+const askLocationPermission = async (setSearchLocationInput: (location: string) => void,
+    saveSearchLocation: (location: string) => void, i18n: I18n): Promise<void> => {
+    const status = await getPermission();
+    if (status === Permissions.PermissionStatus.UNDETERMINED) {
```
tomy-pg, line 327: This condition is too restrictive. It should include `PermissionStatus.DENIED`. If a user initially denies location permission, but later decides they would like to allow it, `Permissions.askAsync` (function body in this conditional) will not be executed since their status at the time will be `DENIED`.

tomy-pg: You may want to also consider an early return on `PermissionStatus.GRANTED` to improve readability as well.

tomy-pg (changes requested):

src/components/search/search_input_component.tsx
```diff
@@ -61,14 +59,16 @@ export const SearchInputComponent = (props: Props): JSX.Element => {
     }
 
     return (
-        <View style = {{paddingTop: 7}}>
+        <View style={{ paddingTop: 7 }}>
```
tomy-pg, line 62: I appreciate this!

src/components/search/search_input_component.tsx
ian-lin-PG: The code is restrictive by design to prevent the location input field is resetting constantly otherwise. If `PermissionStatus.Granted` is used, what would be the expected behaviour. I wouldn't want to set the location input field if user has ran a search using a custom location not my location. Hm....

src/components/search/search_input_component.tsx
```diff
@@ -236,15 +240,23 @@ const SearchButton = (props: {
     readonly searchTermInput: string,
     readonly searchLocationInput: string,
     readonly i18n: I18n,
+    readonly setSearchLocationInput: (s: string) => void;
+    readonly saveSearchLocation: (s: string) => void;
     readonly onSearchRequest: (searchTerm: string, location: string) => void,
 }): JSX.Element => {
     const location = toLocationForQuery(props.searchLocationInput, props.i18n);
+    askLocationPermission(props.setSearchLocationInput, props.saveSearchLocation, props.i18n);
```
tomy-pg, line 248: I am not a huge fan of how this is triggered. React recommends side-effects to be done with their useEffect hook. We can also create a custom hook for this or trigger it with a callback.

tomy-pg (changes requested):

src/components/search/search_input_component.tsx
rasmus-storjohann-PG: Under the SRP, components are responsible for showing state on the screen, meaning that components should not contain complex logic directly. The best place for complex logic are reducers, sagas and hooks. Does it make sense to move this logic to one of those?

src/components/search/search_input_component.tsx
```diff
@@ -236,15 +240,23 @@ const SearchButton = (props: {
...
         <TouchableOpacity
             style={props.searchTermInput.length === 0 ? [applicationStyles.searchButton, applicationStyles.disabled] : applicationStyles.searchButton}
             disabled={props.searchTermInput.length === 0}
             onPress={(): void => {
-                props.onSearchRequest(props.searchTermInput, location);
+                if (location === '') {
+                    const defaultLocation = 'Vancouver';
+                    props.setSearchLocationInput(defaultLocation);
+                    props.saveSearchLocation(defaultLocation);
+                    props.onSearchRequest(props.searchTermInput, defaultLocation);
```
rasmus-storjohann-PG, line 258: Here the component has to worry about a lot of stuff that has nothing to do with presenting the state of the app on the screen. It should not need to know or care that there are three different things going on under the hood when setting the location. It should instead be able to delegate this to a redux action. The first two are already redux actions, so does it make sense to change the reducers to have the same effect? The last triggers a saga I think? It can also be hooked up to an action. Note also that the code here needs to be consistent with what happens in the non-default case ... if it is as messy as this, they both need to be cleaned up ... if it is cleaner than this, then the same cleaner approach can be used here.

rasmus-storjohann-PG (changes requested):
Still got some way to go on this ... make sure to have a good understanding of what the non-default code path looks like so that you can match it in the new default behaviour.

src/components/search/search_input_component.tsx
ian-lin-PG: The code block here has been removed after reviewing the UX with cherrie and melanie.

src/components/search/use_ask_location_permission.ts
```diff
@@ -0,0 +1,23 @@
...
+                const permissionResponse = await Permissions.askAsync(Permissions.LOCATION);
+                if (permissionResponse.status === Permissions.PermissionStatus.GRANTED) {
+                    setLocationInput(i18n._(MY_LOCATION_MESSAGE_DESCRIPTOR));
+                    saveLocation(i18n._(MY_LOCATION_MESSAGE_DESCRIPTOR));
+                }
+            }
+        };
+        askLocationPermission();
+    }, []);
+};
```
rasmus-storjohann-PG, line 23: - [ ] Make local variables for the predicates to explain what's going on
- [ ] Inline `await getPermission();` to make it clear what's the difference between the two calls to `ermissions.askAsync(Permissions.LOCATION);`

rasmus-storjohann-PG (changes requested):

src/components/search/use_ask_location_permission.ts
```diff
@@ -0,0 +1,23 @@
+// tslint:disable: no-expression-statement
+import * as Permissions from 'expo-permissions';
+import { useEffect } from 'react';
+import { MY_LOCATION_MESSAGE_DESCRIPTOR } from '../partial_localization/to_location_for_query';
+import { getPermission } from './search_input_component';
+
+type LocationSetter = (location: string) => void;
+
+export const useAskLocationPermission = (setLocationInput: LocationSetter, saveLocation: LocationSetter, i18n: I18n): void => {
```
rasmus-storjohann-PG, line 9: One more change will make this clearer, the signature should just take a single, zero-argument callback:
```ts
export const useAskLocationPermission = (setLocationToMyLocation: () => void): void => {`
```
then you'll have just "if permission is granted, set location to my location", and all the low level i18n stuff will be in the caller.

tomy-pg: I like this suggestion a lot.

Issue #1364, opened 2021-02-05
iOS AR specific: Service review screen requires padding


tomy-pg:
See screenshot below:
![Simulator Screen Shot - iPhone X - 2021-02-05 at 11.04.06.png](https://images.zenhubusercontent.com/5c6c8cff9d62b128eab447ff/96e46d5e-ff83-425c-812e-8a88986a8ae4)

### Acceptance Criteria 
- [ ]  

### Steps for QA: 
1.

ian-lin-PG:
PASS QA 1.62.0
![Screenshot_20210219-111412](https://user-images.githubusercontent.com/52009863/108550741-d3962400-72a3-11eb-8c72-15bf255d5898.jpg)

PR #1378, opened 2021-02-12
Issue #1364 ensure padding exists for all locales in service screen

tomy-pg:


rasmus-storjohann-PG:
Looks good, go ahead and merge after QA

PR #1380, opened 2021-02-12
Feature/issue 1379 search input bars do not clear

tomy-pg:


src/components/search/search_input_component.tsx
```diff
@@ -140,11 +143,13 @@ const ExpandedInput = (props: ExpandedInputProps): JSX.Element => {
     const searchLocationPlaceholder = t`Enter city, address, or postal code`;
     const clearTermInput = (): void => {
         props.setSearchTermInput('');
+        props.saveSearchTerm('');
         props.searchTermInputRef.current.focus();
     };
 
     const clearLocationInput = (): void => {
         props.setSearchLocationInput('');
+        props.saveSearchLocation('');
```
rasmus-storjohann-PG, line 152: Same issue as in @ian-lin-PG 's PR I think. Do we ever need to save one and not the other? Should they just be part of a single action? If you do make that change, make sure you do it in a separate commit and that you can review that one commit and understand everything that changes in it ... and obviously keeping the compiler and the test suite happy

tomy-pg: With the current design we need to use both of these functions to display the correct UX unless unless we want to send a `saveSearchLocation` action for every character inputted in the search bar:
```
<TextInput 
 onChangeText={(text: string): void => {
      // props.setSearchTermInput(text);
     //props.saveSearchLocation(text)
 }}
/>
```

tomy-pg: I am with you in that I would love for this all to be handled with a single action and state. Using local state and store state for the inputs feels awkward.

tomy-pg: It is recommended by React to use local state to handle inputs. The reason we are saving it to the store is to maintain the state of the inputs when users navigate to and from the search screen. We also use the stores search inputs for when we query Algolia.

rasmus-storjohann-PG: OK, I wasn't too clear in my head on whether this was being stored to local state or store, my bad. So I guess it's really just a bit of a DRY problem that has a simple solution just affecting this one file.

ian-lin-PG (approved):

PR #1445, opened 2021-03-17
Feature/issue 1440 Android device locale Arabic crash

tomy-pg:


rasmus-storjohann-PG (changes requested):

src/sagas/locale.ts
```diff
@@ -74,7 +76,7 @@ const isAppRTL = (deviceLocale: string, appLocale: string): boolean => {
 };
 
 const isAndroidAppRTL = (deviceLocale: string, appLocale: string): boolean => (
-     isSetToRTL(appLocale) || isSetToRTL(deviceLocale) && isSetToRTL(appLocale)
+    isRTL(appLocale) || deviceAndAppAreRTL(deviceLocale, appLocale)
```
rasmus-storjohann-PG: Boolean arithmetic my friend: A || (A && B) = A, since if B is false, A && B is false, in which case A || (A && B) is A || false which is just the same as A. If B is true, A && B is the same as A, and A || A is also the same as A. So the device locale has no effect on the computed value here.

src/sagas/locale.ts
```diff
@@ -63,20 +61,9 @@ export function* loadCurrentLocale(): IterableIterator<CallEffect | PutEffect<ac
...
     if (isAndroid()) {
-        return isAndroidAppRTL(deviceLocale, appLocale);
+        return isRTL(appLocale);
     }
     return I18nManager.isRTL;
-};
-
-const isAndroidAppRTL = (deviceLocale: string, appLocale: string): boolean => (
-    isRTL(appLocale) || deviceAndAppAreRTL(deviceLocale, appLocale)
-);
```
rasmus-storjohann-PG: OK, you figured it out

src/locale/effects.ts
```diff
@@ -19,12 +19,12 @@ export function setTextDirection(localeCode: string): void {
     });
 }
 
-export const isRTL = (localeCode: string): boolean => (localeCode === 'ar');
-
 export function reload(): void {
     Updates.reloadAsync();
 }
 
+export const isRTL = (locale: string): boolean => locale.toLocaleLowerCase().startsWith('ar');
```
rasmus-storjohann-PG, line 26: Is this change needed now with the simplification? I like the old way better if it's all the same ...  having nightmares about supporting Armenian

rasmus-storjohann-PG (changes requested):
Nice, simple solution in the end. Just a couple of more suggestions

src/sagas/locale.ts
tomy-pg: Eventually....

src/locale/effects.ts
tomy-pg: I reverted back to the old way!

PR #1446, opened 2021-03-18
Feature/issue 1431 province selection

ian-lin-PG:


src/components/welcome/welcome_component.tsx
```diff
@@ -66,8 +70,9 @@ export function WelcomeComponent(props: Props): JSX.Element {
                         <Picker
                             mode='dropdown'
                             placeholder='Select province'
+                            selectedValue={props.currentProvince.code}
                             placeholderStyle={[getBoldFontStylesForOS(), { color: colors.teal }]}
-                            onValueChange={handleChange}
+                            onValueChange={props.setProvince}
```
rasmus-storjohann-PG, line 75: "set" province or "save" province - please be consistent. I think "save" is a better antonym to "load"

src/province/types.ts
```diff
@@ -0,0 +1,16 @@
...
+    readonly [provinceCode: string]: string;
+}
+
+export interface ReactI18n {
+    _(str: string): string;
+}
+
+export interface ReactI18nRenderProp {
+    readonly i18n: ReactI18n;
+}
```
rasmus-storjohann-PG, line 16: How come we have not needed these before? I'm worried they're duplicates of existing types. They certainly don't belong in a file that is about "province" stuff

src/province/types.ts
```diff
@@ -0,0 +1,16 @@
+export interface Province {
+    readonly code: string;
+    readonly fallback: string;
+}
```
rasmus-storjohann-PG, line 4: Types like this are either used in the store or created by selectors. This interface should sit in a file that is located either within the store or selector folder, in this case, it is used in both store and selector, so it should be located in the store folder. I would like to avoid having lots of specific folders at the top level under `src/`. I'm not a fan of the locale code, and that includes the current `src/locale` folder, so let's not use it as a pattern to follow.

src/stores/index.ts
```diff
@@ -19,6 +20,7 @@ import * as reviews from './reviews';
 export interface Store {
     readonly fonts: fonts.FontsStore;
     readonly locale: locale.LocaleStore;
+    readonly province: province.ProvinceStore;
```
rasmus-storjohann-PG, line 23: I think it's overkill to have a top level store just for this. This data can be part of the `UserExperience` store. Make sure that it's persisted as well...

src/stores/province/actions.ts
```diff
@@ -0,0 +1,17 @@
+import * as constants from '../../application/constants';
+import * as helpers from '../helpers/make_action';
+
+export type LoadProvinceAction = Readonly<ReturnType<typeof loadProvince>>;
+export type SaveProvinceAction = Readonly<ReturnType<typeof saveProvince>>;
+
+// tslint:disable-next-line:typedef
+export const loadProvince = () => {
+    return helpers.makeAction(constants.LOAD_CURRENT_PROVINCE);
+};
```
rasmus-storjohann-PG, line 10: Is this needed?

src/application/styles.ts
```diff
@@ -489,6 +489,18 @@ export const applicationStyles = StyleSheet.create({
         fontFamily: getNormalFontFamily(),
         ...getTextAlignForLanguage(),
     },
+    item: {
```
rasmus-storjohann-PG, line 492: Can we get a more descriptive name for this style?

src/components/welcome/welcome_component.tsx
```diff
@@ -1,25 +1,29 @@
...
 import { LocaleInfo, Locale } from '../../locale';
 import { SaveLocaleRequestAction } from '../../stores/locale/actions';
 import { Routes, goToRouteWithoutParameter } from '../../application/routing';
-import { applicationStyles, colors, textStyles } from '../../application/styles';
+import { applicationStyles, colors, textStyles, getBoldFontStylesForOS } from '../../application/styles';
 import { arrivalAdvisorLogo, landingPhoto, peacegeeksLogo } from '../../application/images';
 import { History } from 'history';
 import { needsTextDirectionChange } from '../../locale/effects';
+import { Province } from '../../province/types';
+import { SaveProvinceAction } from '../../stores/province/actions';
```
rasmus-storjohann-PG, line 14: How about getting both of these from `../../stores/user_profile.ts`, would that make sense?

src/province/types.ts
rasmus-storjohann-PG: See my other comment, I suggest adding this to `/src/stores/user_profile.ts`

src/stores/province/index.ts
```diff
@@ -0,0 +1,24 @@
...
+    readonly code: string;
+    readonly fallback: string;
+}
+
+export const buildDefaultStore = (): ProvinceStore => ({
+    code: undefined,
+    fallback: undefined,
+});
+
+export const reducer = (store: ProvinceStore = buildDefaultStore(), action?: actions.ProvinceAction): ProvinceStore => {
```
rasmus-storjohann-PG, line 14: This is overkill too, don't need a new reducer for this

rasmus-storjohann-PG (changes requested):
Nice work, I suggested some simplifications and reorganizations.

rasmus-storjohann-PG (changes requested):
Still a couple of things to fix

src/components/welcome/welcome_component.tsx
```diff
@@ -107,6 +128,23 @@ export function WelcomeComponent(props: Props): JSX.Element {
     );
 }
 
+const StartButton = (props: Props): JSX.Element => {
+    if (props.currentRegion && props.currentLocale.code) {
```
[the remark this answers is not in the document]
tomy-pg: > I would also suggest that the locale isn't actually changed until the start button has been hit

I hear you on this and is definitely something that needs to be addressed but I think this takes us away from the 'thin thread' we are aiming for, for the deadline. This functionality is tied to the `saga/locale.ts` which also covers how the language picker works in the hamburger menu. Making this change will require us to either make the same change to the hamburger menu (which currently does not have a start button) or to make a separate saga for just the welcome screen language picker.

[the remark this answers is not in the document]
rasmus-storjohann-PG: Valid points, @tomy-pg , I'm open with just making it work for now and merge with some known issues, including UX glitches that better separation of concerns can fix.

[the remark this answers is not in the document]
tomy-pg: Created an issue for this here: #1450

src/stores/__tests__/user_profiles.test.ts
```diff
@@ -1,115 +1,109 @@
...
+
+        test('is loaded from persisted data', () => {
+            const savedRegion = RegionCode.BC;
+            const oldStore = new UserProfileBuilder().withRegion(savedRegion).build();
+            const dataRegion = new PersistedDataBuilder().
+                withRegion(savedRegion).
+                build();
+            const actionCustomizeLatLong = DataPersistence.loadSuccess(dataRegion);
+            const newStore = reducer(oldStore, actionCustomizeLatLong);
+            expect(newStore.region).toBe(savedRegion);
```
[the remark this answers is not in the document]
rasmus-storjohann-PG: You didn't address this comment

tomy-pg (approved):

PR #1448, opened 2021-03-19
Issue 615 Add MB data to the app

rasmus-storjohann-PG:
…data type

### DIFF
--- a/src/fixtures/types/topics.ts
+++ b/src/fixtures/types/topics.ts
@@ -5,6 +5,7 @@ export type Id = string;
 
 export interface Topic {
     readonly id: Id;
+    readonly region: string;
     readonly chapter: string;
     readonly title: LocalizedText;
     readonly description: LocalizedText;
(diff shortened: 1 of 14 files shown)


src/fixtures/types/topics.ts
```diff
@@ -5,6 +5,7 @@ export type Id = string;
 
 export interface Topic {
     readonly id: Id;
+    readonly region: string;
```
tomy-pg, line 8: Are you keeping `region` as a string as opposed to an enum to maintain flexibility for future expansion?

tomy-pg (commented):
Looks good, just one question.

src/fixtures/types/topics.ts
rasmus-storjohann-PG: Yeah, either would work, just going with the simpler solution for now. I don't expect there will be a lot of code in the app that will look at this field, just a selector that filters topics in this field, so the added typesafety of an enum wouldn't make a significant difference.

PR #1452, opened 2021-03-29
Feature/issue 618 region specific algolia search

tomy-pg:


src/components/search/search_component.tsx
```diff
@@ -80,8 +82,12 @@ export const SearchComponent = (props: Props): JSX.Element => {
         // tslint:disable-next-line: no-let
         let geocoderLatLong = props.searchLatLong;
         try {
-            const vancouverLatLong = { lat: 49.282729, lng: -123.120738 };
-            geocoderLatLong = location === '' ? vancouverLatLong : await fetchLatLongFromLocation(location, props.customLatLong);
```
rasmus-storjohann-PG, line 84: Break out a helper function to map from region to default lat/long

rasmus-storjohann-PG (changes requested):

rasmus-storjohann-PG (changes requested):

src/api/fetch_lat_long_from_location.ts
```diff
@@ -53,11 +54,11 @@ const fetchLatLongFromAddress = async (location: string): Promise<LatLong> => {
     }
 };
 
-export const buildGeoCoderUrl = (location: string): string => (
+export const buildGeoCoderUrl = (location: string, region: RegionCode): string => (
     BuildUrl('https://geocoder.ca', {
         queryParams: {
             locate: location,
-            prov: 'bc',
+            prov: region,
```
rasmus-storjohann-PG, line 61: Here the assumption is that when a value of RegionCode (an enum, not a string) is converted to a string, it ends up lowercase? It would be nice to have code in here that ensures that that actually happens. Something to explicitly convert from the enum to a string and then `lower()`, I guess?

rasmus-storjohann-PG (changes requested):

src/api/fetch_lat_long_from_location.ts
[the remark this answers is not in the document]
rasmus-storjohann-PG: Never mind, I see the enum defines it well enough

PR #1476, opened 2021-05-06
Feature/issue 1473 usereducer in welcome screen

ian-lin-PG:


src/application/locales.ts
```diff
@@ -40,6 +41,26 @@ export interface ReactI18nRenderProp {
...
+                { code: 'ar', label: 'عربى' },
+                { code: 'fr', label: 'Français' },
+                { code: 'ko', label: '한국어' },
+                { code: 'pa', label: 'ਪੰਜਾਬੀ' },
+                { code: 'tl', label: 'Tagalog' },
+                { code: 'zh_CN', label: '简体中文' },
+                { code: 'zh_TW', label: '繁體中文' },
+            ];
+        default:
+            return [];
```
rasmus-storjohann-PG, line 60: Throw exception here instead, it is not valid to have a different region.

src/validation/region/types.ts
```diff
@@ -6,4 +10,23 @@ export interface Region {
 export enum RegionCode {
     BC = 'bc',
     MB = 'mb',
-}
\ No newline at end of file
+}
+
+export interface RegionLocaleMap {
```
rasmus-storjohann-PG, line 15: Naming: How is this a map?

src/components/welcome/welcome_component.tsx
```diff
@@ -1,40 +1,66 @@
...
     const arrivalAdvisorLogoSize = Dimensions.get('screen').width / 2.15;
 
+    const reducer = (regionState: RegionLocaleMap, action: SelectRegionLocaleAction): RegionLocaleMap => {
+        switch (action.type) {
+            case constants.SELECT_REGION:
+                return {
+                    ...regionState,
+                    region: action.payload.region,
+                    locale: undefined,
+                    availableLocales: getAvailableLocales(action.payload.region),
```
rasmus-storjohann-PG, line 44: Even better if the reducer rather than setting the locale to undefined, could be smart about not changing it in cases where the same locale is supported in both the old and new regions

src/components/welcome/welcome_component.tsx
```diff
@@ -1,40 +1,66 @@
...
+                return {
+                    ...regionState,
+                    locale: action.payload.locale,
+                };
+            default:
+                return {
+                    ...regionState,
+                    region: undefined,
+                    availableLocales: [],
+                };
```
rasmus-storjohann-PG, line 56: Under what conditions do you anticipate to hit this default? Is this how to intend to initiate the state? If so, a `buildDefaultState()` helper function would make it more clear what's going on.

src/components/welcome/welcome_component.tsx
```diff
@@ -1,40 +1,66 @@
...
+        }
+    };
+
+    const [selectedRegionState, dispatch]: readonly [RegionLocaleMap, Dispatch<SelectRegionLocaleAction>] = useReducer(
+        reducer, { region: undefined, locale: undefined, availableLocales: [] },
```
rasmus-storjohann-PG, line 61: Is this the best value to use for the default state, or would it be better to use values passed in through the props?

src/components/welcome/welcome_component.tsx
```diff
@@ -89,67 +122,63 @@ export function WelcomeComponent(props: Props): JSX.Element {
     );
 }
 
-const RegionPicker = (props: Props): JSX.Element => {
-
-    const handleRegionChange = (regionCode: RegionCode): void => {
-        props.saveRegion(regionCode);
-        props.resetLocale();
-    };
+const RegionPicker = (state: RegionLocaleMap, dispatch: Dispatch<SelectRegionAction>): JSX.Element => {
```
rasmus-storjohann-PG, line 125: Here's a way to build a habit to pay closer attention to naming: See that you have this "map" type `RegionLocaleMap`, but in a lot of different places where you use it, you use "state" rather than "map". This indicates that "state" is a better name for the type than "map". See if you can make it a habit to pay attention to names in that way

src/components/welcome/welcome_component.tsx
```diff
@@ -89,67 +122,63 @@ export function WelcomeComponent(props: Props): JSX.Element {
...
-                placeholder={placeholder }
-                selectedValue={props.localeIsSet ? props.currentLocale : undefined}
+                placeholder={placeholder}
+                selectedValue={state.locale}
                 placeholderStyle={[getBoldFontStylesForOS(), { color: colors.teal }]}
-                onValueChange={handleLocaleChange}
+                onValueChange={(locale: string): void => dispatch({ type: constants.SELECT_LOCALE, payload: { locale } })}
                 style={applicationStyles.picker}
-                enabled={props.currentRegion ? true : false}
+                enabled={state.region ? true : false}
```
rasmus-storjohann-PG, line 158: You can also use `!!` to convert a value to the corresponding boolean value: Negate once to convert any truthy value to false and any falsy value to true, negate again to convert any truthy value to trye and falsy value to false.

src/components/welcome/welcome_component.tsx
```diff
@@ -63,11 +89,18 @@ export function WelcomeComponent(props: Props): JSX.Element {
...
                 <View>
-                    <StartButton {...props} />
+                    <StartButton
+                        selectedRegionState={selectedRegionState}
+                        currentLocale={props.currentLocale}
+                        currentRegion={props.currentRegion}
+                        showOnboarding={props.showOnboarding}
+                        history={props.history}
+                        saveLocale={props.saveLocale}
+                        saveRegion={props.saveRegion} />
```
rasmus-storjohann-PG, line 103: I think you can remove a lot of code if you create your `onStartButtonPress` function in this scope and pass it down to the start button component. Most of the `StartButtonProps` properties would disappear

src/validation/region/types.ts
```diff
@@ -6,4 +10,23 @@ export interface Region {
...
+
+// tslint:disable-next-line:typedef
+export const selectRegion = (region: RegionCode) => (
+    helpers.makeAction(constants.SELECT_REGION, { region })
+);
+
+// tslint:disable-next-line:typedef
+export const selectLocale = (locale: string) => (
+    helpers.makeAction(constants.SELECT_LOCALE, { locale })
+);
```
rasmus-storjohann-PG, line 32: These action generator functions are written just the way I thought they should, but I am a little concerned with the file that they're in. useReducer is *similar* to Redux, but it's not the same thing exactly. It is very likely that people will be confused, thinking that these are actual Redux actions. A way to reduce that confusion is to put these two functions, and the constants they use as well, inside the component file. That way people will see that they are scoped to that one file and therefore *clearly* not actual Redux actions.

src/validation/region/types.ts
```diff
@@ -1,3 +1,7 @@
+import { LocaleWithLabel } from "../../application/locales";
+import * as constants from '../../application/constants';
+import * as helpers from '../../stores/helpers/make_action';
+
 export interface Region {
     readonly code: RegionCode;
     readonly fallback: RegionCode;
```
rasmus-storjohann-PG, line 7: What is this fallback thing? I think you may have been inspired by a bad pattern in the locale code, which I just removed ;-)

rasmus-storjohann-PG (changes requested):

rasmus-storjohann-PG:
Still no unit tests on the reducer ... this is the whole point of using useReduce

src/components/welcome/index.ts
```diff
@@ -0,0 +1,67 @@
+// tslint:disable: no-this readonly-keyword no-expression-statement no-class
```
rasmus-storjohann-PG, line 1: You can give a bit more thought to the name of this file. I would say that this file contains different kinds of things right now, so it should be split up into two or three smaller files that will be easier to give meaningful names.

rasmus-storjohann-PG (changes requested):
I added some commits to the PR, just because this PR gave a good opportunity for me to demonstrate how to write test code, which will become very handy in upcoming on-boarding. There are some changes I would like you to make as well.

src/components/welcome/welcome_component.tsx
```diff
@@ -122,16 +148,21 @@ export function WelcomeComponent(props: Props): JSX.Element {
...
+    readonly dispatch: Dispatch<SelectRegionLocaleAction>;
+}
+
+const RegionPicker = (props: PickerProps): JSX.Element => {
 
     return (
         <Item style={applicationStyles.pickerItem}>
             <Picker
                 mode='dropdown'
                 placeholder='Select province'
```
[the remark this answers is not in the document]
ian-lin-PG: English since we decided not to update the app as the locale is changed within the welcome screen.

src/components/welcome/welcome_component.tsx
```diff
@@ -122,16 +148,21 @@ export function WelcomeComponent(props: Props): JSX.Element {
...
-                selectedValue={state.region}
+                selectedValue={props.state.region}
                 placeholderStyle={[getBoldFontStylesForOS(), { color: colors.teal }]}
-                onValueChange={(region: RegionCode): void => dispatch({ type: constants.SELECT_REGION, payload: { region } })}
+                onValueChange={(region: RegionCode): void => props.dispatch({ type: constants.SELECT_REGION, payload: { region } })}
```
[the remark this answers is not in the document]
ian-lin-PG: The `dispatch` is from the `useReducer` hook. I have yet to find a way to compile without passing it down.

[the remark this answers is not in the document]
rasmus-storjohann-PG: No what I meant was this
```ts
onValueChange={(region: RegionCode): void => props.dispatch(selectRegion(region) })}
```

PR #1484, opened 2021-05-20
Feature/issue 1481 client upgrade

ian-lin-PG:


src/components/search/search_input_component.tsx
```diff
@@ -309,26 +309,28 @@ const MyLocationButton = (props: {
...
+        case 'always':
+            setSearchLocationInput(i18n._(MY_LOCATION_MESSAGE_DESCRIPTOR));
+            break;
+        case 'fine':
+            setSearchLocationInput(i18n._(MY_LOCATION_MESSAGE_DESCRIPTOR));
+            break;
+        case 'coarse':
+            setSearchLocationInput(i18n._(MY_LOCATION_MESSAGE_DESCRIPTOR));
+            break;
+        case 'none':
```
rasmus-storjohann-PG: Shouldn't this be instead
```ts
if (status === 'none') {
    openAppSettings();
} else {
    setSearchLocationInput(i18n._(MY_LOCATION_MESSAGE_DESCRIPTOR));
}
```

src/components/search/use_ask_location_permission.ts
```diff
@@ -1,15 +1,18 @@
...
+import { isAndroid } from '../../application/helpers/is_android';
+import { requestLocationPermission } from '../../application/helpers/request_location_permission';
 
 export const useAskLocationPermission = (setLocationToMyLocation: () => void): void => {
     useEffect((): void => {
         const askLocationPermission = async (): Promise<void> => {
-            const getPermissionResponse = await Permissions.getAsync(Permissions.LOCATION);
-            const haveNeverAskedBefore = getPermissionResponse.status === Permissions.PermissionStatus.UNDETERMINED;
+            const getPermissionResponse = await Location.getPermissionsAsync();
+            const status = isAndroid() ? getPermissionResponse.android.scope : getPermissionResponse.ios.scope;
```
rasmus-storjohann-PG: This is repeated from above, can be DRYed up

src/components/search/search_input_component.tsx
```diff
@@ -311,19 +311,10 @@ const MyLocationButton = (props: {
...
-            setSearchLocationInput(i18n._(MY_LOCATION_MESSAGE_DESCRIPTOR));
-            break;
-        case 'fine':
-            setSearchLocationInput(i18n._(MY_LOCATION_MESSAGE_DESCRIPTOR));
-            break;
-        case 'coarse':
-            setSearchLocationInput(i18n._(MY_LOCATION_MESSAGE_DESCRIPTOR));
-            break;
-        case 'none':
+        case PermissionStatus.DENIED:
```
rasmus-storjohann-PG, line 317: If/else still the preferred form here

brandon-PG (approved):
Looks good to me

rasmus-storjohann-PG (changes requested):
Really nice work, just some minor changes to suggest

rasmus-storjohann-PG:
@ian-lin-PG one more thing, can you update to use eslint and https://github.com/jonaskello/eslint-plugin-functional while you're at it?

src/components/search/search_input_component.tsx
ian-lin-PG: I eventually installed the expo-location package separately which has the more recognizable Permission.Status that we are more used to.

src/components/search/use_ask_location_permission.ts
ian-lin-PG: Updated package as described as above.

Issue #1490, opened 2021-05-31
About Arrival Advisor page updated per province


cherrielam:
We would like to update the "About Arrival Advisor" page copy and show a slightly different version in each province.

See mocks:
- [About Arrival Advisor page in BC](https://zpl.io/aXjG7Wx)
- [About Arrival Advisor page in MB](https://zpl.io/2pRx8LJ)

~~NOTE: The "About Arrival Advisor" modal page should no longer be a modal; same with "Disclaimer" - all of these pages should be updated to reflect the page design in the mocks above. See related note in #1489~~ per conversation with @brandon-PG we decided not to change the page style to modal since it seems more complicated than it's worth.

### Acceptance Criteria 
- [x] On the "About Arrival Advisor", copy is updated and logos are added. Changing the province shows the correct province's content.
~~- [ ] The "About Arrival Advisor" modal page, and other similar modal pages (eg. "Disclaimer") are updated to match the page design of the mock (no longer a modal).~~
- [ ] The text for Arrival Advisor version number, Service version number, and AAStagingVersion number current shows in the system font. It should show up in Avenir.

### Steps for QA: 
1.

cherrielam:
It's updated! Only issue is the font - can we make sure the circled text below shows up in Avenir?

![IMG_8702](https://user-images.githubusercontent.com/21990736/124329565-7e66fa80-db40-11eb-99a3-2438e45c5cb2.PNG)

ian-lin-PG:
Screenshot after change
![Simulator Screen Shot - iPhone 12 Pro Max - 2021-07-07 at 14 13 28](https://user-images.githubusercontent.com/52009863/124829678-bf3e8500-df2d-11eb-8eed-7143a8722128.png)

cherrielam:
@ian-lin-PG @brandon-PG can you also send an Android screenshot? Just to check that it's behaving in Android as well

brandon-PG:
![Screenshot_1625765996](https://user-images.githubusercontent.com/84804716/124967154-f1f18780-dfd8-11eb-9ea3-041b2099f403.png)
on Android

cherrielam:
Looks great! Thanks

rasmus-storjohann-PG:
Waiting for OK from MBStart on copy, do not QA until they've gotten back to us

alysha-PG:
MB Start has approved copy, font looks good.
[...]

Issue #1489, opened 2021-05-31
Update hamburger menu with logos + Creators page


cherrielam:
We would like to add logos onto the hamburger right side menu. The logos should link to a "Creators" page.

See also #1521 

Mocks:

- [Right menu drawer - MB](https://zpl.io/a3rx4Nl), clicking the logos takes you to [MB version of the "Creators" page](https://zpl.io/aRRon3z)
- [Right menu drawer - BC](https://zpl.io/29Gl7BW), clicking the logos takes you to [BC version of "Creators" page](https://zpl.io/VY5WqwJ)

Note: In the app, we currently have a few kinds of pages: the pop-up modal for "About Arrival Advisor" or "Disclaimer" (accessed via right hamburger menu) and the page for "Services listed on Arrival Advisor" (accessed when you click "Suggest an Update" on a Service and then click "Learn more"). All of those pages should be updated to match the 'Creators' page style.

### Acceptance Criteria 
- [ ]  Right menu drawer shows the correct logos per province, and links to the correct province's "Creators" page
- [ ] The pop-up modal pages (eg. About Arrival Advisor and Disclaimer) are updated to match the design of the "Creators" page
- [ ] The text pages with close button (eg. "Services listed on Arrival Advisor") are updated to match the design of the "Creators" page 

### Steps for QA: 
1.

deniza-pg:
The disclaimer is still BC.

The right menu drawer shows the correct logos per province, and links to the correct province's "Creators" page: There is link to Manitoba 211 but not Manitoba start

brandon-PG:
Changing locales to Arabic will not restart the app with right to left reading. Instead the layout of the app remains the same as the previous language.

rasmus-storjohann-PG:
This may be iOS specific...

brandon-PG:
Is an iOS restart issue https://forums.expo.io/t/updates-reloadasync-doesnt-work-on-standalone-ios-sdk-41/52349

deniza-pg:
The logos don't go to creator's page.

brandon-PG:
The creators page was made a modal for ease of development

ian-lin-PG:
Screenshots of creator's page:
![Simulator Screen Shot - iPhone 12 Pro Max - 2021-08-13 at 11 00 08](https://user-images.githubusercontent.com/52009863/129400692-8e29c54b-63fe-4bcd-90ce-1e2d8f2ddddb.png)
[...]

PR #1492, opened 2021-06-02
Issue #1490 Update About Arrival Advisor page based on region

brandon-PG:
Will have to rebase after #1489 is merged.


src/application/images.ts
```diff
@@ -20,3 +20,7 @@ export const emptySearch = require('../../assets/images/aa_empty_search.png');
 export const noMatchingSearchResults = require('../../assets/images/aa_no_matching_search_results.png');
 export const invalidSearchLocation = require('../../assets/images/aa_invalid_search_location.png');
 export const emptyBookmarks = require('../../assets/images/aa_empty_bookmarks.png');
+export const welcomeBcLogo = require('../../assets/images/sponsor_logo_lock_up_welcome_bc.png');
```
rasmus-storjohann-PG, line 23: `sponsor_logo_lock_up_welcome_bc` is weird, please rename the file to something more straightforward

src/components/about/about_component.tsx
```diff
@@ -79,12 +70,70 @@ export const AboutComponent: React.StatelessComponent<Props> = (props: Props): J
...
+                        <Trans>Information in Arrival Advisor</Trans>
+                    </Text>
+                    <RegionSpecificInformation {...props} />
                 </Content>
             </View>
         </Modal>
     );
 };
 
+const RegionSpecificInformation = (props: { readonly region: RegionCode }): JSX.Element => {
```
rasmus-storjohann-PG, line 83: This function does three things: The logic for which region to show info for, then it contains all the detail for what to show for one and the other region. SRP: Break this into three functions that each do one of those things.

rasmus-storjohann-PG (changes requested):
Just some minor changes requested

Issue #1487, opened 2021-05-27
New first onboarding screen for MB & BC


cherrielam:
We would like to **replace** the current first slide of the onboarding flow to a new one that features the logos for our MB and BC partners. 

See Zeplin mocks for [MB](https://zpl.io/bzdk3oG) and [BC](https://zpl.io/aMwBogJ)

The old version of the first onboarding slide is no longer used

![Screen Shot 2021-05-27 at 10.12.56 AM.png](https://images.zenhubusercontent.com/5823860ee40e5714b1700133/270171d5-7710-40c5-87f8-b36040605099)

### Acceptance Criteria 
- [ ]  First slide of the onboarding in each province shows the logos of that province. 
- [ ]  Copy and design is updated according to Zeplin mocks
- [ ] In AA BC, user sees the BC copy and logos
- [ ] In AA MB, user sees the MB copy and logos

### Steps for QA: 
1.

PR #1493, opened 2021-06-04
Issue #1487 Change onboarding screen based on region

brandon-PG:


src/components/onboarding/onboarding_component.tsx
```diff
@@ -70,17 +80,57 @@ const ONBOARDING_DATA: ReadonlyArray<OnboardingData> = [
       title: t`Answer a few optional questions to get tailored recommendations for your needs.`,
     },
 ];
+const BC_ONBOARDING_TITLE: Readonly<string> = t`Trusted information for newcomers from Newcomers’ Guide to British Columbia and BC211`;
+const MB_ONBOARDING_TITLE: Readonly<string> = t`Trusted information for newcomers from Manitoba Start and 211 Manitoba`;
+
+const OnboardingImage = (props: {readonly item: OnboardingData, readonly region: RegionCode}): JSX.Element => {
+    if (props.item.id === 1 && props.region === RegionCode.MB) {
+        return <MBOnboardingImages/>;
+    } else if (props.item.id === 1 && props.region === RegionCode.BC) {
```
rasmus-storjohann-PG, line 89: You don't need `else` after `return`

src/components/onboarding/onboarding_component.tsx
```diff
@@ -48,11 +56,13 @@ interface OnboardingData {
     readonly title: string;
 }
 
+type Props = OnboardingComponentProps & OnboardingComponentActions & OwnProps;
+
 const ONBOARDING_DATA: ReadonlyArray<OnboardingData> = [
```
rasmus-storjohann-PG, line 61: // one thing that takes the props and gives back the image, as an JSX.Element
// one thing that takes the props and gives back the text
// maybe common logic in a shared helper function for the MB/BC stuff

src/components/onboarding/onboarding_component.tsx
```diff
@@ -134,7 +184,16 @@ function computeSwiperIndex(index: number, isRTL: boolean, platform: PlatformOST
...
+function getOnboardingText(item: OnboardingData, region: RegionCode): string {
+    if (item.id === 1 && region === RegionCode.MB) {
+        return MB_ONBOARDING_TITLE;
+    } else if (item.id === 1 && region === RegionCode.BC) {
+        return BC_ONBOARDING_TITLE;
+    }
+    return item.title;
+}
+
+export const OnboardingComponent = (props: Props): JSX.Element => {
```
rasmus-storjohann-PG, line 196: If you look at the changes for the commit, this changed line is hard to review because the commit is too big and it's not easy to see what it changed from and why

rasmus-storjohann-PG (changes requested):

PR #1501, opened 2021-06-10
Feature/issue 1399 Add alternate name to service page

brandon-PG:


src/validation/search/__tests__/to_human_service_data.tests.ts
```diff
@@ -13,6 +13,7 @@ describe('Adapting search result to service data', () => {
             const result = toHumanServiceData({
                 type: 'SearchServiceData',
                 service_name: aString(),
+                alternate_name: aString(),
                 service_id: aString(),
                 service_description: aString(),
                 last_verified_date: aDate(),
```
ian-lin-PG, line 19: It may help to make a human service data builder to clean up the code. Since we are adding a new field, it may make sense to write a unit test for checking `altername_name` as well. It's quite common for this field to be undefined, so check the expected output in that case as well.

src/components/services/service_detail_component.tsx
```diff
@@ -327,6 +330,16 @@ const Name = (props: NameProps): JSX.Element => (
     </>
 );
 
+const AlternateName = (props: {readonly alternateName: string, readonly isFeedbackInputEnabled: boolean}): JSX.Element => {
+    if (props.alternateName && !props.isFeedbackInputEnabled) {
+        return <AlternateNameText alternateName={props.alternateName}/>;
+    }
+    return <></>;
```
ian-lin-PG, line 336: Is this an empty component, if so we have <EmptyComponent /> that we use for this purpose.

src/components/services/service_detail_component.tsx
```diff
@@ -327,6 +330,16 @@ const Name = (props: NameProps): JSX.Element => (
...
+};
+
+const AlternateNameText = (props: {readonly alternateName: string}): JSX.Element => (
+    <Text style={[markdownStyles.body, {paddingTop: 4}]}><Trans>(aka {props.alternateName})</Trans></Text>
```
ian-lin-PG, line 340: `<Trans>` will translate the text in between the tag, where translations exists. Generally we have the tag around hard coded app strings that we can control, we don't have translations for service data so we can remove the tag in this case.

src/components/services/service_detail_component.tsx
```diff
@@ -221,7 +221,10 @@ export const ServiceDetailComponent = (props: Props): JSX.Element => {
                                     isFeedbackInputEnabled={isFeedbackInputEnabled}
                                     nonFeedbackComponent={<Name name={props.service.name} />}
                                 />
-                                <DividerComponent />
+                                <AlternateName
+                                    alternateName={props.service.alternateName}
+                                    isFeedbackInputEnabled={isFeedbackInputEnabled} />
+                                <EmptyDividerComponent />
```
rasmus-storjohann-PG, line 227: This should be a feedback component, following the pattern of the other pieces. Ian can explain in more detail what Feedback means here.

src/validation/search/__tests__/to_human_service_data.tests.ts
```diff
@@ -300,6 +309,7 @@ describe('Adapting search result to service data', () => {
...
```
rasmus-storjohann-PG, line 315: These changes are fine, but the fact that you have to make all these single line changes is a code smell, it indicates that these tests are to some degree redundant and that they all test the same thing. This breaks with DRY and SRP. If you have time, you might just read this code in detail and try to envisage how it might be changed to avoid this amount of repetition, it would be a very good exercise (but hard!)

rasmus-storjohann-PG (changes requested):
A pretty bit change is needed in this PR

src/components/services/service_detail_component.tsx
```diff
@@ -326,6 +333,44 @@ const Name = (props: NameProps): JSX.Element => (
...
+                <AlternateNameTopMargin isFeedbackInputEnabled={props.isFeedbackInputEnabled} />
+                <FeedbackComponent
+                    setText={props.setFeedbackInputForField('alternateName')}
+                    toggleShouldSend={props.toggleShouldSendForField('alternateName')}
+                    inputField={props.feedback.alternateName}
+                    label={<Trans>Alternate Name</Trans>}
+                    body={props.alternateName}
+                    isFeedbackInputEnabled={props.isFeedbackInputEnabled}
+                    nonFeedbackComponent={<AlternateNameText alternateName={props.alternateName} />}
+                />
```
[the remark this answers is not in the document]
rasmus-storjohann-PG: Looks like the `Description` field is a good one to follow as a pattern for the new field you're adding. I think the change we need is just a bit of refactoring.

src/components/services/service_detail_component.tsx
```diff
@@ -326,6 +333,44 @@ const Name = (props: NameProps): JSX.Element => (
     </>
 );
 
+interface AlternateNameProps {
+    readonly toggleShouldSendForField: (fieldName: keyof ServiceFeedback) => (field: FeedbackField) => void;
+    readonly setFeedbackInputForField: (fieldName: keyof ServiceFeedback) => (field: FeedbackField) => void;
+    readonly feedback: ServiceFeedback;
+    readonly alternateName: string;
+    readonly isFeedbackInputEnabled: boolean;
```
rasmus-storjohann-PG, line 341: This feels non-DRY to me ... this will also be resolved with the refactoring, I expect this props type to go away

src/components/services/service_detail_component.tsx
```diff
@@ -326,6 +333,44 @@ const Name = (props: NameProps): JSX.Element => (
...
+}
+
+const AlternateName = (props: AlternateNameProps): JSX.Element => {
+    if (props.alternateName) {
```
rasmus-storjohann-PG, line 345: Oh yeah, we should also talk about our convention of preferring early returns. I should update our style guide with that....

src/components/services/service_detail_component.tsx
```diff
@@ -549,6 +594,10 @@ const SuggestAnUpdateText = (): JSX.Element => (
     </View>
 );
 
+const EmptyDividerComponent = (): JSX.Element => (
```
rasmus-storjohann-PG, line 597: I'm a little surprised that this is needed in this PR, but I'm open to be convinced

rasmus-storjohann-PG (changes requested):
Still a bit of work left here, but we're getting close. This is quite a big change...

PR #1503, opened 2021-06-15
Feature/issue 1451 new hamburger

ian-lin-PG:


src/application/locales.ts
```diff
@@ -61,6 +61,27 @@ export const getAvailableLocales = (region: RegionCode): ReadonlyArray<LocaleWit
...
+        case 'tl':
+            return 'Tagalog';
+        case 'zh_CN':
+            return '简体中文';
+        case 'zh_TW':
+            return '繁體中文';
+        default:
+            return 'English';
+    }
+};
```
rasmus-storjohann-PG, line 83: This is very clean and tidy, which I like, but I am confused as to why it is needed, because we clearly already has this mapping defined in the system, see `getAvailableLocales()`. The locale management code is complex, I've already spent a lot of time cleaning this up, I don't want to throw another piece in there that makes it even more complicated.

src/components/header_menu/header_menu_component.tsx
```diff
@@ -76,67 +75,36 @@ export interface SectionListItemInfo {
...
-                    type='FontAwesome'
-                    style={{ fontSize: 22, color: colors.teal, marginHorizontal: -2, marginVertical: -2 }}
-                />
+        <View style={{ backgroundColor: colors.white, marginHorizontal: 10 }}>
+            <MenuSectionTitle title={<Trans>SELECT YOUR PROVINCE</Trans>} />
+            <View key={props.currentLocale} style={[applicationStyles.listItem, { justifyContent: 'space-between' }]}>
+                <Text style={[textStyles.headlineH4StyleBlackLeft, { fontWeight: 'bold' }]}>{localeCodeToLabel(props.currentLocale)}</Text>
+                <TouchableOpacity onPress={props.openLanguageDrawer}>
+                    <Text style={[textStyles.headlineH4StyleBlackLeft, { fontWeight: 'bold', color: colors.teal }]}><Trans>Change</Trans></Text>
+                </TouchableOpacity>
```
rasmus-storjohann-PG, line 106: Is there some duplication from the code above?

src/components/header_menu/language_drawer_component.tsx
```diff
@@ -0,0 +1,123 @@
...
+};
+
+const createLocaleItem = R.curry((setLocale: (code: LocaleCode, flipOrientation: boolean) => void,
+    updateToken: () => void,
+    locale: LocaleWithLabel): LocaleListItem => (
+    {
+        ...locale,
+        onPress: (): void => {
+            setLocale(locale.code, I18nManager.isRTL !== isRTL(locale.code));
+            updateToken();
```
rasmus-storjohann-PG, line 92: This is exactly the wrong way to use actions with sagas. Sagas are designed to watch the actions being emitted and respond with async calls. (In this case, we want a new push notification token post request to go to the server so that we know what locale to use when sending notifications to this user.) What is wrong here is that we have UX code that knows about this and has to make this extra call to update token. This is bad because (1) the UX code should not know or care about this, and (2) sagas, when used correctly, are exactly the right solution to this: Have the saga listen for the `setLocale` action, and respond by updating the token. Then the whole `updateToken()` action and all related prop drilling disappears and we are all happier for it.

I know that this is not new code in this PR, so you don't need to change it now ... but we want to fix this soon, and I think it should actually be pretty simple to fix, since it should mostly consist of removing code.

src/components/header_menu/language_drawer_component.tsx
```diff
@@ -0,0 +1,123 @@
...
+        },
+    }
+),
+);
+
+const LocaleItem = (sectionListLocaleItem: SectionListItemInfo): JSX.Element => {
+    return (
```
rasmus-storjohann-PG, line 99: https://docs.google.com/document/d/1C8A7-vdiMDFSIuR7rID8tuSM1n6SxHKmNaOLXuFcWLI/edit#heading=h.udq7bye9nrkh

src/components/main/main_component.tsx
```diff
@@ -97,38 +103,56 @@ export const MainComponent = (props: Props): JSX.Element => {
...
-                    </ScrollContext.Provider>
-                </Container>
+                <Drawer
+                    side='right'
+                    onClose={props.closeRegionDrawer}
+                    open={props.isRegionDrawerVisible}
+                    content={
+                        <RegionDrawerConnectedComponent />
+                    }
+                >
```
rasmus-storjohann-PG, line 122: Fix white space

src/selectors/user_experience/select_is_language_drawer_visible.ts
```diff
@@ -0,0 +1,6 @@
+import { Store } from '../../stores';
+import { LanguageDrawer } from '../../stores/user_experience';
+
+export const selectIsLanguageDrawerVisible = (appStore: Store): boolean => (
+    appStore.userExperience.languageDrawer !== LanguageDrawer.LanguageDrawerIsClosed
```
rasmus-storjohann-PG, line 5: Sometimes it's the little things that really helps readability and code quality, would you not prefer to read this?
```ts
appStore.userExperience.languageDrawer === LanguageDrawer.LanguageDrawerIsOpen
```
Negations slows us down. Also note the inconsistent naming "Visible" vs "Open"/"Closed".

src/selectors/user_experience/select_is_region_drawer_visible.ts
```diff
@@ -0,0 +1,6 @@
+import { Store } from '../../stores';
+import { RegionDrawer } from '../../stores/user_experience';
+
+export const selectIsRegionDrawerVisible = (appStore: Store): boolean => (
+    appStore.userExperience.regionDrawer !== RegionDrawer.RegionDrawerIsClosed
```
rasmus-storjohann-PG, line 5: "visible" vs "open"/"closed" here too

src/validation/region/types.ts
```diff
@@ -15,3 +15,12 @@ export interface RegionLocaleState {
     readonly locale: string;
     readonly availableLocales: ReadonlyArray<LocaleWithLabel>;
 }
+
+export const regionCodeToLabel = (region: RegionCode): string => {
+    switch (region) {
+        case RegionCode.MB:
+            return 'Manitoba';
+        default:
+            return 'British Columbia';
```
rasmus-storjohann-PG, line 24: Defensive programming suggests a different appriach to a switch statement like this: Map MB to Manitoba, BC to British Columbia, and throw an exception in the default. That way if we add a region and forget to update this code, we know about it right away.

rasmus-storjohann-PG (changes requested):
@ian-lin-PG excellent work on the hamburger menu. I've made a few suggestions, but they don't all have to go in now, as long as we raise another PR against this same story and fix remaining issues there right after we've pushed to staging.

src/application/locales.ts
ian-lin-PG: We should either have `LocaleCode` or `LocaleWithLabel`. The difficulty with `LocaleWithLabel` is that it's harder to set than simply `LocaleCode`. So `LocaleCode` is saved in the store, we can certainly map it inside the connected component.

That's why I only have 'RegionCode` and chose to have a similar getLabel function there.

src/components/main/main_component.tsx
ian-lin-PG: The tree drawers are nested together, not sure which white space?

src/components/header_menu/region_drawer_connected_component.tsx
```diff
@@ -0,0 +1,26 @@
...
+const mapStateToProps = (store: Store): RegionDrawerProps => ({
+    currentLocale: selectLocale(store),
+    currentRegion: pickRegion(store),
+});
+
+type Actions = CloseRegionDrawerAction | SaveRegionAction | OpenLanguageDrawerAction | SaveLocaleRequestAction;
+const mapDispatchToProps = (dispatch: Dispatch<Actions>): RegionDrawerActions => ({
+    closeRegionDrawer: (): CloseRegionDrawerAction => dispatch(closeRegionDrawer()),
+    saveLocale: (localeCode: string, flipOrientation: boolean): SaveLocaleRequestAction => dispatch(saveLocaleRequest(localeCode, flipOrientation)),
+    saveRegion: (regionCode: RegionCode): SaveRegionAction => (
```
brandon-PG, line 21: Can be 1 line

src/application/styles.ts
```diff
@@ -85,6 +85,10 @@ export const getTextAlignForLanguage = (): TextStyle => (
     { textAlign: I18nManager.isRTL ? 'right' : 'left' }
 );
 
+export const getViewBackgroundColorForPlatform = (): string => (
```
brandon-PG, line 88: The teal background on Android causes this:
Screenshot_1624402357

brandon-PG (changes requested):

src/components/header_menu/header_menu_component.tsx
rasmus-storjohann-PG: Nice solution to this issue

src/components/header_menu/language_drawer_component.tsx
```diff
@@ -0,0 +1,128 @@
...
+            <LanguageAvailabilityDisclaimer region={props.currentRegion} />
+            <LocaleSection {...props} />
+        </View>
+    </View>
+);
+
+const LanguageAvailabilityDisclaimer = (props: { readonly region: RegionCode }): JSX.Element => {
+    const translatedDisclaimer = props.region === RegionCode.MB ?
+        <Trans>Information about Manitoba is available in the following languages:</Trans> :
+        <Trans>Information about British Columbia is available in the following languages:</Trans>;
```
rasmus-storjohann-PG, line 71: You should use placeholders for region here. Also, change the text to "Information about {newRegion} is only available in {languages}, do you want to access information about {newRegion} in English?"

PR #1504, opened 2021-06-15
Feature/issue 1482 welcome screen redesign

brandon-PG:


src/components/welcome/welcome_component.tsx
```diff
@@ -76,6 +77,10 @@ export function WelcomeComponent(props: Props): JSX.Element {
                         dispatch={dispatch}
                     />
                 </Form>
+                <RegionPickerN
```
rasmus-storjohann-PG: Why the N at the end? No need to fix, I see you change it later...

src/application/styles.ts
```diff
@@ -491,15 +491,17 @@ export const applicationStyles = StyleSheet.create({
     },
```
rasmus-storjohann-PG, line 498: Fix commit message

src/components/welcome/welcome_component.tsx
```diff
@@ -146,17 +150,14 @@ export interface StartButtonProps {
 }
 
 const StartButton = (props: StartButtonProps): JSX.Element => {
-    if (props.state) {
+    if (props.state.region && props.state.locale) {
```
rasmus-storjohann-PG, line 130: Doing it this way, the very important logic of which of these components are visible at any time is hidden inside each component. It will be much clearer if the caller that uses these components, contains the logic of whether or not each should be visible. The same applies to `LocalePicker` above. This would also make it easier to pull this logic into helper functions, making SRP happy.

src/components/welcome/welcome_component.tsx
```diff
@@ -99,15 +99,7 @@ const LocalePicker = (props: PickerProps): JSX.Element => {
...
-            <View style={{
-                justifyContent: 'center',
-                marginVertical: 16,
-                borderColor: colors.teal,
-                borderRadius: 50,
-                borderWidth: 2,
-                height: 64,
-                width: 232,
-            }}>
+            <View style={applicationStyles.pickerContainer}>
```
rasmus-storjohann-PG, line 102: nice, like the consistent naming, and the small commits

src/components/welcome/welcome_component.tsx
```diff
@@ -76,6 +77,10 @@ export function WelcomeComponent(props: Props): JSX.Element {
                         dispatch={dispatch}
                     />
                 </Form>
+                <RegionPickerN
+                    state={selectedRegionState}
```
rasmus-storjohann-PG, line 77: Fix the commit message

src/components/welcome/welcome_component.tsx
```diff
@@ -50,64 +49,40 @@ export function WelcomeComponent(props: Props): JSX.Element {
...
-                    source={arrivalAdvisorLogo}
-                    resizeMode={'contain'}
+                    source={welcomeHeader}
+                    resizeMode='contain'
                     style={{
-                        width: arrivalAdvisorLogoSize,
-                        height: arrivalAdvisorLogoSize,
-                        marginBottom: 20,
+                        width: Dimensions.get('window').width,
+                        height: Dimensions.get('window').width * 34 / 36,
```
rasmus-storjohann-PG, line 63: That math looks a little dodgy. Won't keeping the aspect ratio fixed just work?

src/components/welcome/welcome_component.tsx
```diff
@@ -116,48 +91,33 @@ export interface PickerProps {
...
+        <View>
+            <Text style={textStyles.headlineH3StyleBlackCenter}><Trans>Select language</Trans></Text>
+            <View style={applicationStyles.pickerContainer}>
+                <Item style={applicationStyles.pickerItem}>
+                    <Picker
+                        mode='dropdown'
+                        placeholder={placeholder}
+                        selectedValue={props.state.locale}
+                        placeholderStyle={[getBoldFontStylesForOS(), { color: colors.teal }]}
+                        onValueChange={(locale: string): void => props.dispatch({ type: constants.SELECT_LOCALE, payload: { locale } })}
```
rasmus-storjohann-PG, line 109: Not really SRP compliant, a helper function would be better.

rasmus-storjohann-PG (changes requested):

Issue #1508, opened 2021-06-17
IOS notification crashes app upon cold start


brandon-PG:
When an IOS user clicks on a notification with the app not in the background, the app with crash on start.

### Acceptance Criteria 
- [ ] The app should open when the notification is clicked  

### Steps for QA: 
0. Ensure that the app is not currently running on the phone
1. Send a notification to an IOS phone
2. Click on the notification when it appears
3. Verify the app opens correctly

ian-lin-PG:
PASS QA 1.74.0

PR #1509, opened 2021-06-18
Issue #1508 Fix crash when handling notification json

brandon-PG:


src/components/main/notification.ts
```diff
@@ -23,7 +23,9 @@ export const notificationListener = R.curry((history: History, notification: any
 });
 
 // tslint:disable-next-line:no-any
-const getValidRouteFromNotificationPayload = (data: any): string | undefined => {
+const getValidRouteFromNotificationPayload = (notification: any): string | undefined => {
+    // tslint:disable-next-line:no-string-literal
+    const data = notification['notification']['request']['content']['data'];
```
rasmus-storjohann-PG, line 28:
```ts
    const data = notification.notification.request.content.data;
```
or
```ts
    const data = notification?.notification?.request?.content?.data;
```

rasmus-storjohann-PG: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

Issue #1498, opened 2021-06-08
Upgrade ajv


ian-lin-PG:
The ajv API for validation and errors have changed, which means we need to change all the calls referencing the library. The error is now an object instead of a string, and all tests relying on the error message will now fail. Even after fixing all the library calls, user data tests all fails on unknown field `address` even though this field is not present in the user data schema or persisted data builder.

### Acceptance Criteria 
- [ ]  ajv package upgrade passes all tests

### Steps for QA: 
1.

rasmus-storjohann-PG:
Unit tests pass, no further QA needed

PR #1512, opened 2021-06-22
Issue #1498 Upgrade ajv and fix broken tests

brandon-PG:


src/validation/content/index.ts
```diff
@@ -5,7 +5,7 @@ import { alertArray } from './schema';
 const Ajv = require('ajv');
 
 export const validateAlertResponse = (data: ReadonlyArray<any>): ValidationResult<Alert> => {
-    const ajv = new Ajv();
+    const ajv = new Ajv({strict: false});
```
ian-lin-PG, line 8: Just a tad curious with the difference between `{strict: true}` and `{strict: false}`

brandon-PG: https://ajv.js.org/strict-mode.html
Strict mode did not exist until ajv version 7 and is now turned on by default.

rasmus-storjohann-PG: OK, so it's strict with respect to the way the schemas are defined. It does not make it more or less strict with respect to validating the incoming data against those schemas. So non-strict can be OK ... still feels like a bit of tech debt...

PR #1513, opened 2021-06-24
Feature/issue 1511 manitoba string changes

brandon-PG:


rasmus-storjohann-PG (changes requested):
Need some updates to this PR. Consider whether you want to change some of the if/elses to switch statements, just to see how they are a more natural control flow mechanism for enums.

Issue #1527, opened 2021-07-06
Android back button does not close region/locale drawers


brandon-PG:
Tapping the physical back button will not close the region/locale drawer.

### Acceptance Criteria 
- [ ] Tapping  the physical back button once should exit the drawer

### Steps for QA: 
1.  Open the region drawer in the hamburger menu
2. Tap the physical back button on an android device
3. Verify the drawer closes
4. Repeat for the language drawer

rasmus-storjohann-PG:
This passes QA

PR #1528, opened 2021-07-06
Issue #1527 Add support for physical back button in region and language drawer

brandon-PG:


ian-lin-PG:
@brandon-PG I just noticed a related bug, the back button when the app first start up will cause a crash. I found out after clicking one extra back button after closing the drawers. This also shows up on staging app. Though it's not specific to this issue, it does touch on similar code, I think it can fixed within this PR.

ian-lin-PG:
here is a screenshot of the showstopper
Screenshot_20210707-141150

brandon-PG:
> @brandon-PG I just noticed a related bug, the back button when the app first start up will cause a crash. I found out after clicking one extra back button after closing the drawers. This also shows up on staging app. Though it's not specific to this issue, it does touch on similar code, I think it can fixed within this PR.

This crash is fixed

ian-lin-PG:
I can confirm pressing the back button at the beginning closes the app instead of crashing the app.

ian-lin-PG (approved):

Issue #1473, opened 2021-04-28
Language and province selection in welcome screen


rasmus-storjohann-PG:
The welcome screen has a selector for region and a selector for locale and an OK button for the user to click to proceed with the currently selected region and locale.

Current behaviour is that the app resets to the new locale each time the user changes the selected locale.

Expected behaviour is that the app does not change the locale until the OK button is clicked.

This will involve keeping more state in the component (as opposed to in redux or locale store), which will be managed using useReducer()

cherrielam:
Once a language selection is made, the selection text is no longer shown in Avenir. Is it possible to make it show in Avenir?
![IMG_8698](https://user-images.githubusercontent.com/21990736/124327052-58d7f200-db3c-11eb-8b64-a1e636cf0264.PNG)

Also, I just realized you can't go "back" after selecting the province. Would it be possible/easy to add a back button? Could be a new issue.
![IMG_8698](https://user-images.githubusercontent.com/21990736/124327161-815fec00-db3c-11eb-9034-efe0734d1935.PNG)

ian-lin-PG:
Screenshot after the change
![Simulator Screen Shot - iPhone 12 Pro Max - 2021-07-07 at 13 48 56](https://user-images.githubusercontent.com/52009863/124826947-48ec5380-df2a-11eb-8d4e-1be5d780b0bc.png)

cherrielam:
Looks good

alysha-PG:
picker in iOS is now in avenir, looks good.

PR #1529, opened 2021-07-06
Issue #1473 Change styling of locale picker item

brandon-PG:
This change will only appear on iOS. Android does not allow for styling of the picker.


ian-lin-PG:
Screenshot after the change.
Simulator Screen Shot - iPhone 12 Pro Max - 2021-07-07 at 13 48 56

Issue #1524, opened 2021-07-02
BC Services showing when switching to MB


cherrielam:
When you run a search in BC, and then switch to MB, the BC services stay in the search page. However if you run the same search, it will not work as those services aren't available in MB. This is a confusing experience for the user. A design should be proposed to address this.

### Acceptance Criteria 
- [ ]  Clear service search results from the redux store on the change region action. This can just be done in the service search reducer in response to the set region action.

### Steps for QA: 
1.

PR #1533, opened 2021-07-14
Issue #1524 Clear search results when region is changed

brandon-PG:


ian-lin-PG (approved):

src/stores/search.ts
```diff
@@ -125,6 +127,7 @@ export const reducer = (store: SearchStore = buildDefaultStore(), action?: Searc
                 searchLocation: action.payload.searchLocation,
             });
         case constants.CLEAR_ALL_USER_DATA:
+        case constants.SAVE_CURRENT_REGION:
```
rasmus-storjohann-PG, line 130: This reducer has unit tests, you need to add a test to confirm this behavour. It's important to make sure it works, and also the unit tests end up documenting what the code is supposed to do.

rasmus-storjohann-PG (changes requested):

PR #1538, opened 2021-07-19
Feature/issue 1389 increase suggest update font weight

brandon-PG:


rasmus-storjohann-PG:
On IOS: 
image

rasmus-storjohann-PG:
On Android: 
image

cherriecyl:
Looks good!

Issue #1517, opened 2021-06-29
Make RTL text work


alysha-PG:
Arabic is read right-to-left, and in the app it is aligned to the left. UX would be improved if it could be right-aligned. Will be the same case with Farsi if added in the future.

### Acceptance Criteria 
- [x]  Arabic text is right-aligned

### Steps for QA: 
1.  Change the language from English to Arabic
2. Verify app restarts and app is RTL

rasmus-storjohann-PG:
We need a screenshot. @alysha-PG

alysha-PG:
Now I am having the reverse problem of the issue. Arabic text is now showing up right-aligned (image 1) but then when I switch back to English, it's still right-aligned (image 2).
Not sure if this is an issue elsewhere.

![WhatsApp Image 2021-07-07 at 2 51 33 PM](https://user-images.githubusercontent.com/85197659/124833354-ee0b2a00-df32-11eb-9c3c-1ff3986d092f.jpeg)
![WhatsApp Image 2021-07-07 at 2 51 34 PM](https://user-images.githubusercontent.com/85197659/124833360-efd4ed80-df32-11eb-8bcb-b4a0bfcd43bb.jpeg)

brandon-PG:
On iOS in staging, Updates.reloadAsync does nothing.

alysha-PG:
App switches between correctly right-aligned Arabic and correctly left-aligned English.

PR #1548, opened 2021-07-22
Issue #1517 Revert expo sdk to version 40

brandon-PG:


rasmus-storjohann-PG:
I tested this on Adroid, boots up fine

rasmus-storjohann-PG:
Tested the PR, it works on a real android and a simulated iPhone. Maybe test it on a real iPhone and then it can be merged

PR #1549, opened 2021-07-23
Feature/issue 1523 make app work with mb topics

rasmus-storjohann-PG:


ian-lin-PG:
Can confirm that switching region on topic page will not crash the app and also removes the topics from the old region. This works in conjunction with the #1551 service equivalent.

brandon-PG (approved):

PR #1551, opened 2021-07-26
Feature/issue 1523 bookmark services in different regions

rasmus-storjohann-PG:


ian-lin-PG:
Switching region after bookmarked topics doesn't crash and removes the service that is not in the same region. The booking topic still crashes.

src/selectors/services/filter_bookmarked_services.ts
```diff
@@ -0,0 +1,12 @@
+import R from 'ramda';
```
brandon-PG, line 1: Same issue as the other pr

brandon-PG (changes requested):

src/selectors/services/filter_bookmarked_services.ts
rasmus-storjohann-PG: Fixed @brandon-PG

brandon-PG (approved):

Issue #1552, opened 2021-07-26
Crash when changing regions on a topic screen


brandon-PG:
### Acceptance Criteria 
- [ ]  After changing regions the user is sent back to the recommended topics screen

### Steps for QA: 
1.  Navigate to a topic with the region selected to British Columbia
2. Change regions to Manitoba
3. Verify the recommended topics screen is visible

PR #1560, opened 2021-07-28
Issue #1552 Switch to recommended topics when changing region

brandon-PG:


src/components/header_menu/region_drawer_component.tsx
```diff
@@ -12,6 +12,12 @@ import * as R from 'ramda';
 import { LocaleCode, ReactI18n, ReactI18nRenderProp } from '../../application/locales';
 import { SaveLocaleRequestAction } from '../../stores/locale/actions';
 import { isRTL } from '../../application/locale_effects';
+import { goToRouteWithoutParameter, Routes } from '../../application/routing';
+import { History } from 'history';
+
+export interface OwnProps {
+    readonly history: History;
+}
```
rasmus-storjohann-PG, line 20: Can the history property just be added to the `RegionDrawerProps`?

src/components/header_menu/region_drawer_component.tsx
```diff
@@ -102,11 +108,17 @@ const OtherRegion = (props: Props & { readonly otherRegion: RegionCode }): JSX.E
...
+            resetMemoryHistory(props.history);
+            goToRouteWithoutParameter(Routes.RecommendedTopics, props.history);
             props.saveRegion(props.otherRegion);
         }
         props.closeRegionDrawer();
     };
 
+    const resetMemoryHistory = (history: History): void => {
+        history.go(0 - history.length);
+    };
```
rasmus-storjohann-PG, line 120: This feels like a system-wide helper function, so it should be in a file of its own located in `src/application/helpers/`

src/components/main/main_component.tsx
```diff
@@ -125,7 +125,7 @@ export const MainComponent = (props: Props): JSX.Element => {
                     side='right'
                     onClose={props.closeRegionDrawer}
                     open={props.isRegionDrawerVisible}
-                    content={<RegionDrawerConnectedComponent />}
+                    content={<RegionDrawerConnectedComponent history={memoryHistory} />}
```
rasmus-storjohann-PG, line 128: Huh, so you pass it to the connected component and it is passed on through, interesting, would not have thought of doing it that way

rasmus-storjohann-PG (changes requested):
Nice work. Really happy with the solution, just see the two minor tweaks I suggest.

src/components/header_menu/region_drawer_component.tsx
rasmus-storjohann-PG: I think I would rename it to `clearRoutingHistory()`

src/components/header_menu/region_drawer_component.tsx
brandon-PG: It cannot be added to `RegionDrawerProps` or it will interfere with the `mapStateToProps` in the connected component.

PR #1561, opened 2021-07-29
PeaceGeeksSociety/content#810 Allow MB Start path to be passed as arg…

rasmus-storjohann-PG:
…ument


bin/prepare_deploy.sh
```diff
@@ -21,6 +21,10 @@ while (( "$#" )); do
     then
         BC211PATH=$2
         shift 2
+    elif [ "$1" == "--mbwinpath" ]
+    then
+        MBIWNPATH=$2
```
ian-lin-PG, line 26: `MBIWNPATH` is a typo.

Issue #1399, opened 2021-02-22
UX Polish: Add Alternate Name to Service Detail


cherrielam:
We would like the AlternateName for a Service to the Service Detail page. This is because users can search for services using the AlternateName but the AlternateName itself is nowhere to be seen in the search result detail page. 

See mock here: https://zpl.io/b6Gp4Gj

You will notice that the separating lines have been removed in the top section of the Service Detail page to make the content cleaner and more easily readable for the user. 

### Acceptance Criteria 
- [ ]  AlternateName is added to the Service Detail as per https://zpl.io/b6Gp4Gj

### Steps for QA: 
1.

cherrielam:
Can we capitalize the "A" in "Alternate name"? (In the parentheses)

PR #1575, opened 2021-08-05
Issue #1399 Capitalize the A in Alternate

brandon-PG:


ian-lin-PG (approved):

rasmus-storjohann-PG:
@ian-lin-PG this PR will mess with the translations potentially ... is it safe to merge, you think? We may need to keep track of this and fix things manually to make it translate correctly.

PR #1580, opened 2021-08-10
Design MB UI cleanup

cherrielam:
Here is a PR of my UI tweaks for our Manitoba release!


src/components/welcome/welcome_component.tsx
```diff
@@ -127,7 +129,7 @@ const LocalePicker = (props: PickerProps): JSX.Element => {
                         textStyle={textStyles.pickerTealText}
                         onValueChange={(locale: string): void => onLocaleChange(props.dispatch, locale)}
                         style={applicationStyles.picker}
-                        iosIcon={<Icon name='keyboard-arrow-down' type='MaterialIcons' />}
+                        iosIcon={<Icon name='keyboard-arrow-down' type='MaterialIcons' style={{ color: colors.teal }} />}
```
rasmus-storjohann-PG, line 132: Should this be main colour from Brandon's recent change

cherrielam: no because the theme is not applied to the welcome pages :) as discussed by @brandon-PG

PR #1586, opened 2021-08-12
Feature/issue 1581 ux tweaks

rasmus-storjohann-PG:


src/components/welcome/welcome_component.tsx
```diff
@@ -127,7 +129,7 @@ const LocalePicker = (props: PickerProps): JSX.Element => {
                         textStyle={textStyles.pickerTealText}
                         onValueChange={(locale: string): void => onLocaleChange(props.dispatch, locale)}
                         style={applicationStyles.picker}
-                        iosIcon={<Icon name='keyboard-arrow-down' type='MaterialIcons' />}
+                        iosIcon={<Icon name='keyboard-arrow-down' type='MaterialIcons' style={{ color: colors.teal }} />}
```
rasmus-storjohann-PG, line 132: @cherrielam should this be teal or `props.theme.mainColour` so it becomes green in MB?

cherriecyl: no I don't think so because the welcome screen locale picker does not have the theme applied (theme is only passed after the welcome flow)

cherriecyl: oh sorry I mean it should be teal

PR #1594, opened 2021-08-20
Feature/issue 1566 welcome back button

brandon-PG:


src/components/welcome/welcome_component.tsx
```diff
@@ -171,22 +173,28 @@ const StartButton = (props: StartButtonProps): JSX.Element => {
...
+                <Icon name={I18nManager.isRTL ? 'chevron-right' : 'chevron-left'}
+                    type='FontAwesome'
+                    style={{ color: colors.teal, fontSize: 16, paddingEnd: 8 }}
+                />
+                <Text style={[textStyles.whiteTealButton, { fontSize: 16 }]}><Trans>Back</Trans></Text>
+            </TouchableOpacity>
+        );
+    }
+    return <EmptyComponent />;
+};
```
rasmus-storjohann-PG: Prefer early return, not sure if I've given you the background on the reasoning behind it? Anyway, it's more readable to write "if not visible, return empty." followed by the meat of the function. We can discuss if you like. It's all in Uncle Bob's Clean Code too.

rasmus-storjohann-PG (changes requested):
Just one small tweak needed to the code, the UX looks good

PR #1596, opened 2021-08-24
Issue #1595 add french language options to manitoba

ian-lin-PG:


brandon-PG:
"Language not supported" alert appears when switching from bc to mb with french as the selected language.

Issue #1590, opened 2021-08-16
Integrate translated UX strings back into app


alysha-PG:
Translated UX strings can be found here:

https://drive.google.com/drive/u/2/folders/1d1ZDMh3SIsV8QUN9onvbcwPcKV0pzvsY

- [x] BC UX strings (multilingual)

- [x] MB UX strings (French)

rasmus-storjohann-PG:
![image](https://user-images.githubusercontent.com/32307781/131388325-0959ebbb-975d-476c-929d-5d7a479a5ce9.png)

rasmus-storjohann-PG:
![image](https://user-images.githubusercontent.com/32307781/131397798-5f87e7c9-b8d8-403b-bc3c-13830e994260.png)

deniza-pg:
![IMG_2156](https://user-images.githubusercontent.com/63252138/131402009-9a2f1103-e267-4512-afa2-3745e66b8e6f.PNG)
![IMG_2155](https://user-images.githubusercontent.com/63252138/131402013-c8d96cb1-75be-45fc-acbe-d6fd89bbde79.PNG)
![IMG_2154](https://user-images.githubusercontent.com/63252138/131402018-a9f9cf64-c8bc-48d4-8856-4a1b92ea7a76.PNG)
![IMG_2153](https://user-images.githubusercontent.com/63252138/131402019-ce514521-5e54-4dbf-8c60-5d80b3f99329.PNG)
![IMG_2150](https://user-images.githubusercontent.com/63252138/131402022-6743b2d4-cece-4a41-8503-cdeb4faa95c8.PNG)
![IMG_2149](https://user-images.githubusercontent.com/63252138/131402025-6b9d03ba-2368-43a8-83bb-7627438a7767.PNG)
![IMG_2148](https://user-images.githubusercontent.com/63252138/131402026-b4c78109-6ed1-4013-9aee-edf7c9e239a2.PNG)

![IMG_2157](https://user-images.githubusercontent.com/63252138/131402000-18c7d678-1ffd-48a4-8909-9be0ffb7b120.PNG)

rasmus-storjohann-PG:
![image](https://user-images.githubusercontent.com/32307781/133337535-4931349b-5430-4549-a704-a70b4a2d7ffa.png)

PR #1612, opened 2021-09-14
Issue #1590 Fix question string

ian-lin-PG:


src/fixtures/hard_coded/questionnaire.ts
```diff
@@ -18,13 +18,13 @@ export const buildQuestionnaireFixture = (): ValidQuestionnaireStore => {
         questionDocumentsBc: {
             id: 'questionDocumentsBc',
             showInRegions: [RegionCode.BC],
-            text: 'Do you have any of the following important Canadian documents?',
+            text: 'Do you have any of the following important Canadian documents? Select all that apply.',
```
rasmus-storjohann-PG, line 21: will we have translation for these? Or do we just google translate it?

rasmus-storjohann-PG (changes requested):

src/fixtures/hard_coded/questionnaire.ts
rasmus-storjohann-PG: Oh I see that the new string matches what's in Weblate. I will merge now.

Issue #1615, opened 2021-09-17
Related topics only open after two clicks (all languages)


Farouk-PG:
Reproduce the issue: Go to related topics and click on any related topic with any selected language.
Observation: you need to clicks on any related topic icon for the click to take effect (hear the sound of clicks in the video below)
Expected: Only one click to transition

https://user-images.githubusercontent.com/89798867/133832922-a5b5d770-6ea4-4755-b582-4dfb1201d48a.MP4

rasmus-storjohann-PG:
Fix this after the MB release is out.

Farouk-PG:
It was fixed, but it affected the bookmark action, so we prioritised bookmark action over double click until we figure a better solution that fixes both

rasmus-storjohann-PG:
Neither Alysha or myself could reproduce, need to get a sense of how widespread this issue is.

Farouk-PG:
I just checked and it looks fixed in this release which I am not sure how but great!

Farouk-PG:
Sorry I said it looks fixed but it is still there on my staging app on my phone, but I trust it is a react-native ios issue only.

PR #1624, opened 2021-09-23
Issue #1615 Fix buttons two click error

Farouk-PG:


brandon-PG:
Make sure you commit messages are in the format of  ``` Issue #<Issue Number> <Commit message>```
In this case the message would be ```Issue #1615 Fix buttons two click error```

Farouk-PG:
Updated!

PR #1628, opened 2021-09-29
Feature/issue 1614 topic services with algolia

brandon-PG:


src/sagas/services.ts
```diff
@@ -49,19 +47,14 @@ export function* updateServicesForTopic(action: actions.BuildServicesRequestActi
             }));
         }
 
-        const apiResponse: APIResponse = yield call(searchServices, topicId, deviceLocationResponse);
-        if (errors.isBadResponseError(apiResponse)) {
-            return yield put(actions.buildServicesError(topicId, Errors.BadServerResponse));
-        }
+        const algoliaResponse: AlgoliaResponse = yield call(fetchServicesForTopic, topicId, deviceLocationResponse);
```
rasmus-storjohann-PG, line 50: Here too I think you're missing some error handling that was here before

rasmus-storjohann-PG (changes requested):

Issue #1535, opened 2021-07-15
Clicking Done on questionnaire should set active question to the first question


rasmus-storjohann-PG:
Open the questionnaire
Answer every question, clicking Next for each
On the last question, answer and click Done to finish the questionnaire
Observe the "you have new recommendations" alert
Observe that you're taking to the home screen
Click to open the questionnaire again
Observe that you're taken to the last question in the questionnaire
Expected is that you're taken to the *first* question, since you completed it before.

### Acceptance Criteria 
- [x]  Change behaviour on clicking "Done" on the questionnare, so that next time, the questionnaire opens on the first question

rasmus-storjohann-PG:
The action to close the questionnaire needs to take the current region as part of its payload, so that the questionnaire reducer can set active question to the first question for the given region.

rasmus-storjohann-PG:
Alternatively, closing the questionnaire could set the active question to `undefined`. `buildDefaultStore()` should then also set active question to `undefined`. The selector pulling out the questionnaire will then need to get test for active question being undefined, pull the region from the store and initialize the active question to the first question for the region.

alysha-PG:
questionnaire opens on first question

PR #1630, opened 2021-10-01
Issue #1535 Reset to first question when closing questionnaire

brandon-PG:


src/stores/questionnaire/reduce_valid_store.ts
```diff
@@ -41,10 +42,11 @@ export const reduceValidStore = (store: ValidQuestionnaireStore, action?: Questi
     }
 };
 
-const setOldAnswersToCurrentAnswers = (store: ValidQuestionnaireStore): ValidQuestionnaireStore => (
+const setOldAnswersToCurrentAnswers = (store: ValidQuestionnaireStore, region: RegionCode): ValidQuestionnaireStore => (
     new ValidQuestionnaireStore({
         ...store,
         oldAnswers: store.answers,
+        activeQuestion: getFirstQuestionId(region),
```
rasmus-storjohann-PG, line 49: Looks good I think, if you can change the logic in `getFirstQuestionId()` to instead of hardcoding question ids, to use the better approach of `const regionalQuestions = R.filter(regionMatches(region), questionsList);` from `toSelectorQuestion()` then we're a bit more solid. Instead of `R.filter()` you would I guess use `R.find()` to find the first question in the questionnaire with the right region.

rasmus-storjohann-PG (changes requested):

src/stores/questionnaire/reduce_valid_store.ts
rasmus-storjohann-PG: I tried out your PR, doesn't look like the problem is fixed. We can talk about the solution.

PR #1672, opened 2021-10-12
Feature/issue 1651 Put organization data in store rather than API call

brandon-PG:


src/stores/organization/index.ts
```diff
@@ -30,6 +31,14 @@ export const reducer = (store: OrganizationStore = buildDefaultStore(), action?:
                     errorMessageType: action.payload.errorMessageType,
                 },
             };
+        case constants.OPEN_SERVICE_DETAIL:
```
rasmus-storjohann-PG, line 34: I think it would be more natural to do this on the service data request success, so the data doesn't need to do this trip to the UX layer before it gets added to the store.

src/sagas/organization.ts
```diff
@@ -17,23 +15,12 @@ export function* watchUOpenOrganization(): IterableIterator<ForkEffect> {
...
-        if (!isOnline) {
-            return yield put(buildOrganizationFailure(organizationId, Errors.Offline));
-        }
-
-        const apiResponse: APIResponse = yield call(getOrganization, organizationId);
-        const organization: HumanOrganizationData = apiResponse.results;
-        yield put(buildOrganizationSuccess(organization));
-    } catch (error) {
-        return yield put(buildOrganizationFailure(organizationId, Errors.Exception));
+    const existingOrganization: HumanOrganizationData = yield select(selectOrganizationById, organizationId);
```
rasmus-storjohann-PG, line 18: Doing this in a saga is a bit of tech-debt, since it could just be done in a selector. Tech debt can just be a fact of life though ...

src/validation/search/__tests__/to_human_service_data.tests.ts
```diff
@@ -338,9 +320,7 @@ describe('Adapting search result to service data', () => {
                 service_name: aString(),
                 alternate_name: aString(),
                 service_id: aString(),
-                service_description: {
-                    en: aString(),
-                },
+                service_description: aDescription(),
```
rasmus-storjohann-PG, line 323: Nice

src/components/organizations/organization_detail_component.tsx
```diff
@@ -19,7 +19,7 @@ interface OrganizationDetailProps {
 export const OrganizationDetailComponent = (props: OrganizationDetailProps): JSX.Element => (
     <Content padder>
         <MarkdownBodyComponent
-            body={props.organization.description}
+            body={props.organization.description.en}
             shouldBeExpandable={true}
             // TODO Issue #1080 When organization detail page is online connect the following 2 states to the store/persisted data
             showLinkAlerts={true}
```
rasmus-storjohann-PG: If the comment is no longer valid, or the issue it makes reference to, get rid of both.

src/components/services/get_service_description.ts
```diff
@@ -1,7 +1,7 @@
 import { LocaleCode } from '../../application/locales';
 import { Description } from '../../validation/services/types';
 
-export const getServiceDescription = (description: Description, locale: LocaleCode): string => {
+export const getDescription = (description: Description, locale: LocaleCode): string => {
     if (locale === 'fr' && description.fr) {
         return description.fr;
```
rasmus-storjohann-PG, line 6: Filename should match function name

src/components/services/service_detail_component.tsx
```diff
@@ -65,7 +65,7 @@ import { Rating } from '../../stores/reviews';
 import { OffsetHook, useOffset } from '../use_offset';
 import { UxTheme } from '../../application/ux_themes';
 import { BackButtonComponent } from '../header_button/back_button_component';
-import { getDescription } from './get_service_description';
+import { getDescription } from './get_description';
```
rasmus-storjohann-PG: Thanks, ignore my earlier comment

src/components/organizations/organization_detail_component.tsx
```diff
@@ -8,30 +8,28 @@ import { MarkdownBodyComponent } from '../content_layout/markdown_body_component
...
             theme={props.theme}
             hideLinkAlerts={console.log} />
         <DividerComponent />
-        <OrganizationContactDetailsComponent
-            organization={props.organization}
-            analyticsLinkPressed={props.analyticsLinkPressed}
-            currentPathForAnalytics={props.currentPathForAnalytics}
-            theme={props.theme}
-        />
+        <OrganizationContactDetailsComponent {...props} />
```
rasmus-storjohann-PG, line 32: We ran into problems with this form in the past and ended up preferring the more explicit form that was there before.

src/validation/search/schema.ts
```diff
@@ -55,11 +42,24 @@ export const serviceDescription = {
...
+    'required': ['id', 'name', 'website', 'email', 'service_count', 'description'],
+};
+
 export const serviceSearchItem = {
     'type': 'object',
     'properties': {
         'service_name': { 'type': 'string' },
-        'service_description': serviceDescription,
+        'service_description': description,
         'service_id': { 'type': 'string' },
```
rasmus-storjohann-PG, line 63: I agree with the more general term so you can use it more widely. Possibly "localized string" might take you one step further and let you use the same code for title and other things too.

PR #1714, opened 2021-11-15
Issue #1686 Use the same store value for both searches

brandon-PG:


src/components/help/help_connected_component.tsx
```diff
@@ -10,22 +10,24 @@ import {
...
 import { selectHelpTopicForRegion } from '../../selectors/topics/select_help_topic_for_region';
 import { pickRegion } from '../../selectors/region/pick_region';
 import { selectUxTheme } from '../../selectors/user_experience/select_ux_theme';
+import { selectSearchLocation } from '../../selectors/search/select_search_location';
+import { selectSearchLatLong } from '../../selectors/search/select_search_lat_long';
 
 const mapStateToProps = (store: Store, ownProps: RouterProps): HelpComponentProps => ({
     topicId: selectHelpTopicForRegion(store),
     history: ownProps.history,
-    manualUserLocation: selectManualUserLocation(store),
```
rasmus-storjohann-PG, line 25: It looks like this selector, `selectManualUserLocation()`, can be removed, and presumably `manualUserLocation` store property as well....

rasmus-storjohann-PG (changes requested):

Issue #1684, opened 2021-10-15
“Brought to you by” section links appear on the left side in Arabic


Farouk-PG:
clicking on “Brought to you by” section in Arabic, links for PeaceGeeks, welcomebc, and bc211 appear on the left side. Expected to have them under the texts on the right side.

### Acceptance Criteria 
- [ ]  links are from right side under text

![IMG_5303](https://user-images.githubusercontent.com/89798867/137547670-6f894aee-f498-47de-a83d-90b81aebcbaa.PNG)

alysha-PG:
links are now right-aligning!

PR #1718, opened 2021-11-16
Issue #1684 fix links alignment in Arabic

Farouk-PG:


src/components/creator/creator_component.tsx
```diff
@@ -39,7 +39,7 @@ export const CreatorComponent: React.StatelessComponent<Props> = (props: Props):
                                 PeaceGeeks creates digital tools to support greater peace, connection and safety for those who have been displaced.
                             </Trans>
                         </Text>
-                        <Link href={pgUrl} style={[textStyles.link, { fontSize: 14 }]}>peacegeeks.org</Link>
+                        <Link href={pgUrl} style={[textStyles.link, { fontSize: 14, alignSelf: "right", writingDirection: "ltr" }]}>peacegeeks.org</Link>
```
rasmus-storjohann-PG, line 42: Expect you'll need to use `I18nManager.isRTL` here so that it looks correct in non-RTL languages too. There are several examples in the code base for you to follo

rasmus-storjohann-PG (changes requested):
I think this is not done yet, let me know if I'm wrong.

PR #1728, opened 2021-11-19
Feature/issue 1694 region specific error

brandon-PG:
Needs https://github.com/pg-irc/ui-strings/pull/118 for functionality


src/components/error_screens/error_screen_switcher_component.tsx
```diff
@@ -171,22 +171,25 @@ const renderNoMatchingResultsSuggestions = (): JSX.Element => (
...
-            {bulletPoint} <Trans>City - "{getCity(region)}"</Trans>
+            {bulletPoint} <Trans>City - "{city}"</Trans>
         </Text>
         <Text style={textStyles.paragraphStyleBrown}>
-            {bulletPoint} <Trans>Street - "{getStreet(region)}"</Trans>
+            {bulletPoint} <Trans>Street - "{street}"</Trans>
         </Text>
         <Text style={textStyles.paragraphStyleBrown}>
-            {bulletPoint} <Trans>Postal Code - "{getPostalCode(region)}" </Trans>
+            {bulletPoint} <Trans>Postal Code - "{postalCode}"</Trans>
```
rasmus-storjohann-PG, line 189: I don't think this will work, the <Trans> thing works by lingui scanning the source code for the <Trans> elements and pulling out the strings. With this code, the string will be computed at run time, so too late for the translation to happen
I think we should aim for the regional dependent parts of these strings to not need translation, i.e. they should be outside <Trans> elements

rasmus-storjohann-PG: So this should be something like 
```tsx
{bulletPoint} <Trans>Postal Code</Trans> - "{postalCode}"
```

rasmus-storjohann-PG (changes requested):

src/components/error_screens/error_screen_switcher_component.tsx
brandon-PG: It works with [pg-irc/ui-strings#118 ](https://github.com/pg-irc/ui-strings/pull/118) in the ui strings repo. I could use the change you suggested but its the same result

src/components/error_screens/error_screen_switcher_component.tsx
```diff
@@ -178,10 +180,17 @@ const renderInvalidSearchLocationSuggestions = (): JSX.Element => (
...
         </Text>
     </View>
 );
+
+const getStreet = (region: RegionCode): string => {
+    if (region === RegionCode.MB) {
+        return '271 Portage Ave Winnipeg';
+    }
+    return '1356 Kingsway Vancouver';
+};
```
rasmus-storjohann-PG, line 199: I know this will work, but it's complicated, in particular it is possible that this will confuse translation work needlessly, because we can make it simpler by chaning the return type to `JSX.Element` and just do `return <Trans>Street - 271 Portage Ave Winnipeg</Trans>` for the MB case and `return <Trans>1356 Kingsway Vancouver</Trans>` for the BC case.

src/components/organizations/organization_component.tsx
```diff
@@ -33,6 +34,7 @@ export interface OrganizationProps {
     readonly bookmarkedServicesIds: ReadonlyArray<Id>;
     readonly theme: UxTheme;
     readonly locale: LocaleCode;
+    readonly region: RegionCode;
```
rasmus-storjohann-PG, line 37: With the PR that was just merged, there is now a `useRegion()` hook that would let remove some of this prop drilling, you can use it if you like

src/components/error_screens/error_screen_switcher_component.tsx
```diff
@@ -194,3 +194,17 @@ const getStreet = (region: RegionCode): string => {
...
+    }
+    return 'Vancouver';
+};
+
+const getPostalCode = (region: RegionCode): string => {
+    if (region === RegionCode.MB) {
+        return 'R3B 2A8';
+    }
+    return 'V5Y 1V4';
+};
```
rasmus-storjohann-PG, line 213: Same idea as for the street, just return a JSX.Element with the entire translated string

src/components/error_screens/error_screen_switcher_component.tsx
rasmus-storjohann-PG: Actually, this is nice work, keep it the way it is, ignore my comments above.

rasmus-storjohann-PG (changes requested):

Issue #1716, opened 2021-11-16
Replace old BC211 site links with updated links


alysha-PG:
Recently, BC211 updated their website to bc.211.ca . Now, links that went to a specific entry in the past bring people to this page: https://bc.211.ca/redirect

We need to update anywhere in the app where it links to "https://bc211.ca" into "https://bc.211.ca"

We discovered this from old issues that linked to BC 211 entries, but need to find out where inside the app it links to the redirect page and update the link. So far, we know it shows up on the Creators page.

rasmus-storjohann-PG:
Need to update links from our tasks content as well

alysha-PG:
Links confirmed changed on creators page and directories topic.

In services, bc211 entry still shows the old website - we can either manually change this now or expect their updated service entry to contain the right link in the future.

rasmus-storjohann-PG:
@Farouk-PG and @brandon-PG can you tell me what if anything remains to be done on this issue?

Farouk-PG:
I haven’t revisited this since we discussed it and decided the solution needs to be after updates from BC211 and also different than the one I proposed.  I remember that Brandon had a pr regarding BC211 but not sure if it tackled this issue. I will let him speak on that.

Farouk-PG:
Closing as this has been either fixed or still waiting for updates from BC211. In both cases, a new issue can be created if still needed.

PR #1735, opened 2021-11-23
Issue #1716 Change bc211 links to new url

brandon-PG:


rasmus-storjohann-PG (commented):
There should also be a PR in the content repo, changing the URLs in the 'how-to-use-directories-and-the-internet' chapter and possibly elsewhere in the newcomers' guide.

PR #1750, opened 2021-12-10
Feature/issue 1744 mb211 banner

brandon-PG:


brandon-PG:
Requires https://github.com/pg-irc/ui-strings/pull/122 to function

src/components/search/search_results_component.tsx
```diff
@@ -287,6 +282,13 @@ const ListHeaderComponent = (props: { readonly region: RegionCode, readonly them
     );
 };
 
+const RegionSpecificHeaderText = (props: { readonly region: RegionCode }): JSX.Element => {
+    if (props.region === RegionCode.MB) {
+        return <Trans>Search is currently only available in English and French. For support in other languages, please</Trans>;
+    }
+    return <Trans>Search is currently only available in English. For support in other languages, please</Trans>;
```
rasmus-storjohann-PG, line 289: This is really tricky, we're building a string out of two pieces, "...for support in other lanugages" + "please call 211...", where the strings are both translated and region dependent. You are assuming that the "please call" part of the sentence is always at the end. If there is a language where it has to come in the middle, i.e. there is a non-region-dependent piece at the end of the sentence, we won't be able to handle the translation correctly. I don't know if we will run into that. It seems pretty unlikely and I'm willing to take the chance that we;ll be all right, but the real solition to this will be quite a bit more complicated than what is here now, especially with it being clickable to make the phone call too.

Issue #1756, opened 2021-12-17
Missing Translation: BC/French/Welcome to BC/Public Transportation topic


alysha-PG:
I expected to go to BC/French/Bienvenue en CB/Public Transportation topic and see the topic title written in French. Instead it is in English.

<img src="https://images.zenhubusercontent.com/60b823128e6f424e93549168/dc564cdc-a88d-4e89-842a-45e880abc0a1" width=300 align=centre><br>

### Acceptance Criteria 
- [x]  Change "Public Transportation" to French "Transport public".

### Steps for QA: 
1.  Go to topics
2. Click on "Bienvenue en CB"
3. Look for "Transport public" and make sure it doesn't still say "Public Transportation"

alysha-PG:
fixed.

rasmus-storjohann-PG:
This issue is still there, testing version 1.89.0 on Andoid, MB release candidate

rasmus-storjohann-PG:
Reversing my re-opening of this issue, the problem is tracked here https://github.com/pg-irc/pathways-frontend/issues/1757

Issue #1765, opened 2022-01-03
BC Teal in MB


alysha-PG:
I expect all of MB to have green colour scheme, instead there's 1 place where it's still teal - topics > find services near me> , when location is off, the button to 'turn on location'.

<img src="https://images.zenhubusercontent.com/60b823128e6f424e93549168/6a7de14f-2c11-4ad4-97dc-1010f118271c" width=300 align=centre><br>

### Steps for QA: 
1.  Turn off location
2. Go to any topic
3. Click 'Search for services near me"
4. Observe that the button to turn on location is Green.

Farouk-PG:
Closing as fixed!

PR #1772, opened 2022-01-11
Issue #1756 Add theme to error screen

brandon-PG:


rasmus-storjohann-PG:
Very possible that I could have done a better communicating this change, but are you aware of this new thing I made called `useTheme()`? It uses context to make theme available basically as a global. Check it out

rasmus-storjohann-PG:
Wrong issue number in commit message, should be #1765

PR #1774, opened 2022-01-11
Feature/issue 1771 new unit tests for persisted storage format changes

rasmus-storjohann-PG:


brandon-PG:
Looks good but what is with the empty file?

Issue #1773, opened 2022-01-11
Change taxonomy id to "questionnaire" for all taxonomy terms used in the questionnaire


rasmus-storjohann-PG:
Currently, if a topic is tagged with `timeInCanada:under1Year` and `immigrant_type:refugee_claimant`, and in the questionnaire you select an answer with `immigrant_type:refugee_claimant`, the topic does not get recommended, due to the boolean logic implied by the taxonomy system. Because these terms belong to different taxonomies, boolean AND is applied between them. 

Expected is that if *either* `timeInCanada:under1Year` or `immigrant_type:refugee_claimant` is selected in the questionnaire, the topic should be recommended. 

The solution is to use `questionnaire:under1Year` and `questionnaire:refugee_claimant` as the taxonomy terms for all topics and all answers in the questionnaire. With this change, the terms will both belong to the same taxonomy, and boolean OR will be applied, fixing the usability issue.

- [ ] For the duration of working on this bug, copy the content files for MB and BC into the frontend repo so that changes to the questionnaire code and the content can be made at the same time
- [ ] Work in small commits to make the review process easier. Make corresponding changes to the questionnaire code and the content together and commit them together.
- [ ] When this issue has been completed and merged, the content files can be moved back to their proper location in the content repo.

Farouk-PG:
Duplicate taxonomy fixtures:

"french_level": {
    		"fluent": {},
    		"intermediate": {}
    	},

with:

 	"english_level": {
    		"none": {},
    		"beginner": {},
    		"intermediate": {},
    		"fluent": {}
    	},
-
  	"timeInCanada": {
    		"under1Year": {},
    		"notYet": {},
    		"over1Year": {}
    	},

with:

	"timeinCanada": {
    		"notYet": {}
    	},
-

 	"healthcare": {
    		"needDentist": {}
    	},

with:

	"healthCare": {
    		"disability": {},
    		"needDentist": {},
    		"needEyeExam": {},
    		"information": {},
    		"emergency": {},
    		"bc_services_card": {},
    		"needMedication": {},
    		"costs": {},
    		"MSP": {},
    		"MBHealth": {},
    		"MBHealthCard": {}

Farouk-PG:
Note: There appears to not be taxonomies for “french_level” tag

PR #1775, opened 2022-01-12
Issue #1773 update taxonomyId in fixture

brandon-PG:


src/fixtures/hard_coded/questionnaire.ts
```diff
@@ -202,7 +202,7 @@ export const buildQuestionnaireFixture = (): ValidQuestionnaireStore => {
             isInverted: true,
             taxonomyTerms: [
                 {
-                    taxonomyId: 'driving',
+                    taxonomyId: 'questionnaire',
                     taxonomyTermId: 'driversLicence',
                 },
             ],
```
rasmus-storjohann-PG, line 208: Now that taxonomyId is `questionnaire` for every element in this data, we can make some nice simplifications. We can remove the `taxonomyId` property from these terms, and update the code to hard-code that value. We then end up with 
```ts
            taxonomyTerms: [
                {
                    taxonomyTermId: 'driversLicence',
                },
            ],
```
which is also redundant, so it can be simplified even further to
```ts
            taxonomyTerms: [ 'driversLicence' ],
```
IMPORTANT don;t make this change until the changes to the content repo have been merged ... if we find a mistake in the content repo, we'll need to update this file again, which is harder after this refactoring is done.

rasmus-storjohann-PG (changes requested):

src/fixtures/hard_coded/questionnaire.ts
rasmus-storjohann-PG: Also important, this will have ripple effects to the mongo code generating the fixture

rasmus-storjohann-PG: In conclusion, let's hold off on making these changes ... we may make broader changes to make the questionnaire system scale better that will make these issues irrelevant.

Issue #1762, opened 2021-12-23
Press to call asking about external link


alysha-PG:
When I press call 311, I expect it to trigger a prompt on my phone which asks if I want to call. Instead, the prompt that comes up is asking if I want to open an external link, and when I press yes, then the prompt to call comes up. 

### Acceptance Criteria 
- [ ]  when I press to call 311 or 211 or any number, the prompt to call comes up, not the external link question.

### Steps for QA: 
1.

Farouk-PG:
Closing as fixed

PR #1776, opened 2022-01-13
Issue #1762 Remove external call alert for phone numbers

brandon-PG:


src/components/markdown/markdown_component.tsx
```diff
@@ -16,6 +16,22 @@ interface Props {
 
 export const MarkdownComponent = (props: Props): JSX.Element => {
     const theme = useTheme();
+
+    const isTelephoneNumberHref = (href: string): boolean => {
+        const telHref = /^tel:/;
+        return telHref.test(href);
```
rasmus-storjohann-PG, line 22: There is also https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith

src/components/markdown/markdown_component.tsx
```diff
@@ -16,6 +16,22 @@ interface Props {
...
+
+    // tslint:disable-next-line: no-any
+    const linkOnPress = (node: any, i18n: I18n, hideLinkAlerts: () => void): void | Promise<void> => {
+        const href = node.attributes.href;
+        if (isTelephoneNumberHref(href)) {
+            return Linking.openURL(href);
+        }
+
+        return alertOnLinkClicked(node, i18n, hideLinkAlerts);
+    };
```
rasmus-storjohann-PG, line 33: Clean Code talks about keeping all the code in a given function at the same "level". Here you have high level code like `alertOnLinkClicked(node, i18n, hideLinkAlerts);` mixed with low level code like `const href = node.attributes.href;` that get's into the nitty gritty details. Simplify by changing the call to `if (isTelephoneNumberHref(node)) {...`

src/components/markdown/markdown_component.tsx
```diff
@@ -16,6 +16,22 @@ interface Props {
...
 export const MarkdownComponent = (props: Props): JSX.Element => {
     const theme = useTheme();
+
+    const isTelephoneNumberHref = (href: string): boolean => {
+        const telHref = /^tel:/;
+        return telHref.test(href);
+    };
+
+    // tslint:disable-next-line: no-any
+    const linkOnPress = (node: any, i18n: I18n, hideLinkAlerts: () => void): void | Promise<void> => {
```
rasmus-storjohann-PG, line 26: Can you use an `await` to simplify the return type to just be `void`?

rasmus-storjohann-PG (changes requested):

PR #1782, opened 2022-01-24
Feature/issue 1516 switch locale

Farouk-PG:


rasmus-storjohann-PG:
I will put up a PR at some point with my suggested solution to this.

PR #1792, opened 2022-01-31
Issue 1569 bookmark banner revisions

brandon-PG:


src/components/bookmarks/bookmarks_connected_component.tsx
```diff
@@ -21,6 +23,7 @@ const mapStateToProps = (store: Store): BookmarksProps => ({
     isSendingReview: selectIsSendingReview(store),
     theme: selectUxTheme(store),
     locale: selectLocale(store),
+    showBookmarkBanner: selectBookmarkBanner(store) === BookmarkBanner.BookmarkBannerIsShown,
```
brandon-PG, line 26: Should I create a new selector with this logic? This selector is used in select_user_data_for_local_persistence.ts

src/stores/user_profile.ts
```diff
@@ -102,6 +125,16 @@ export const reducer = (store: UserProfileStore = buildDefaultStore(), action?:
                 ...store,
                 showLinkAlerts: false,
             });
+        case constants.HIDE_BOOKMARK_BANNER:
+            return ({
+                ...store,
+                bookmarkBanner: BookmarkBanner.BookmarkBannerIsHidden,
+            });
+        case constants.FINISH_WELCOME_COMPONENT:
```
rasmus-storjohann-PG, line 133: Are there other reducers that respond to `SAVE_CURRENT_REGION` and therefore also need to respond to `FINISH_WELCOME_COMPONENT`?

rasmus-storjohann-PG (changes requested):
Really nice work, please make note of the refactoring here to get a better sense of the correct way to use Redux. Some minor changes still needed

src/stores/locale/index.ts
```diff
@@ -24,6 +24,8 @@ export const reducer = (store: LocaleStore = buildDefaultStore(), action?: actio
             return { ...store, loading: false, code: action.payload.localeCode };
         case constants.LOAD_CURRENT_LOCALE_FAILURE:
             return { ...store, loading: false };
+        case constants.FINISH_WELCOME_FLOW:
+            // intended fall through
```
rasmus-storjohann-PG, line 28: This is the part I don't understand. The finish welcome flow has the region, the locale and the flip orientation properties. I only see the region property being used. Where is the code that takes the locale and flip orientation of a finish welcome flow action and applies it to the store? This should line up with the corresponding action triggered by changing the language in the hamburger menu

src/stores/__tests__/locale.test.ts
```diff
@@ -130,6 +146,14 @@ describe('the reducer', () => {
         expect(theNewStore.loading).toBe(true);
     });
 
+    it('when called with FINISH_WELCOME_FLOW should return store with loading flag set', () => {
+        const theStore = new LocaleStoreBuilder().build();
+        const regionCode: RegionCode = undefined;
+        const theAction = finishWelcomeFlow(regionCode, aLocaleCode, aBoolean());
+        const theNewStore = locale.reducer(theStore, theAction);
+        expect(theNewStore.loading).toBe(true);
+    });
```
rasmus-storjohann-PG, line 155: Missing tests to make sure the reducer also takes the new locale and flip orientation flags into account

src/stores/__tests__/user_profiles.test.ts
```diff
@@ -193,4 +201,42 @@ describe('user profile reducer', () => {
...
+
+    describe('the bookmark banner flag', () => {
+
+        test('is set by hide bookmark banner action', () => {
+            const oldStore = new UserProfileBuilder().withBookmarkBanner(BookmarkBanner.BookmarkBannerIsShown).build();
+            const newStore = reducer(oldStore, hideBookmarkBanner());
+            expect(newStore.bookmarkBanner).toBe(BookmarkBanner.BookmarkBannerIsHidden);
+        });
+
+        test('is set to shown by save region action when in intial state', () => {
```
rasmus-storjohann-PG, line 207: test name is not right

src/stores/__tests__/user_profiles.test.ts
```diff
@@ -193,4 +201,42 @@ describe('user profile reducer', () => {
             expect(newStore.showLinkAlerts).toBe(true);
         });
     });
+
+    describe('the bookmark banner flag', () => {
```
rasmus-storjohann-PG, line 205: The tests in this section should tell the story of how this flag is set. This means that the tests need to be in a particular order, so that it is clear that all the needed tests are there. You're very close, but not quite there and it makes the tests hard to read. Also I like to use `it()` over `test()` for readability, so you get this:
- it('is set to RegionHasNotBeenChanged by default')
- it('is set to BookmarkBannerIsShown by a save region action')
- it('is set to BookmarkBannerIsHidden by a hide banner action')
- it('is not changed by a save region action when it is already hidden')

src/stores/locale/index.ts
brandon-PG: src/sagas/locale.ts are where the locales are applied

rasmus-storjohann-PG: Right, OK, that makes sense now

rasmus-storjohann-PG: So I was looking to make sure that all the fields in the new action were actually used in a reducer somewhere, which is how most actions work, but because locale is weird, those fields are actually consumed by a saga, not a reducer, so it makes sense.

PR #1798, opened 2022-02-04
Feature/issue 1789 onboarding flow

brandon-PG:


rasmus-storjohann-PG:
Last page of onboarding copy: "Take our optional questionnaire to see recommended topics for your needs". The questionnaire is no longer optional. @alysha-PG , thoughts? CTA button on the last onboarding page is "Start", is that the best label for that button?

PR #1817, opened 2022-03-07
Feature/issue 1799 add algolia analytics

brandon-PG:


src/components/search/api/fetch_search_results_from_query.ts
```diff
@@ -36,13 +37,14 @@ export const fetchSearchResultsFromQuery = async (
...
         aroundRadius: latLong ? 25000 : '',
+        clickAnalytics: true,
     });
 
     try {
         const response = await fetch(url, buildAlgoliaRequest(algoliaQuery));
         const responseJSON: AlgoliaResponse = await response.json();
         setNumberOfPages(responseJSON.nbPages);
-        return validateServiceSearchResponse(responseJSON.hits);
+        return validateServiceSearchResponse(addQueryIDProperty(responseJSON.hits, responseJSON.queryID));
```
rasmus-storjohann-PG, line 47: Hmmm, let's talk about data validation

src/components/search/api/fetch_search_results_from_query.ts
```diff
@@ -167,6 +169,12 @@ const addTopicAlgoliaParameter = (parameters: TopicAlgoliaParams, location: Mayb
     };
 };
 
+// tslint:disable-next-line: no-any
+const addQueryIDProperty = (hits: ReadonlyArray<any>, queryID: string): ReadonlyArray<any> => (
+    // tslint:disable-next-line: typedef
+    hits.map(serviceObject => ({ ...serviceObject, queryID }))
```
rasmus-storjohann-PG, line 175: If we do validation correctly, we can remove these comments

src/components/search/search_results_component.tsx
```diff
@@ -169,6 +170,7 @@ const renderSearchHit = R.curry((props: SearchHitProps, itemInfo: ListRenderItem
     const onPress = (): void => {
         props.saveServiceToMap(service);
         props.openServiceDetail(service);
+        sendSearchAnalytics(service.objectID, item.queryID, itemInfo.index);
```
rasmus-storjohann-PG, line 173: Async operations like sending data to a server should make you think "saga". Sagas respond to actions, you already have an action triggered by `openServiceDetail()` that already contains the info you need, so a saga will be simpler to write

src/components/search/send_search_analytics.ts
```diff
@@ -0,0 +1,37 @@
+import { ALGOLIA_APPLICATION_ID, ALGOLIA_SEARCH_API_KEY, ALGOLIA_SERVICES_INDEX } from 'react-native-dotenv';
+
+// tslint:disable-next-line: no-any
+export const sendSearchAnalytics = async (objectID: string, queryID: string, position: number): Promise<any> => {
+    const url = 'https://insights.algolia.io/1/events';
+    const query =  JSON.stringify(getQuery(objectID, queryID, position));
+    const response = await fetch(url, buildAlgoliaRequest(query));
+    return response;
+
+};
```
rasmus-storjohann-PG, line 10: Turn this into a saga, it should be straightforward I think.

src/validation/search/__tests__/to_human_service_data.tests.ts
```diff
@@ -17,6 +17,8 @@ describe('Adapting search result to service data', () => {
                 service_id: aString(),
                 service_description: aDescription(),
                 last_verified_date: aDate(),
+                objectID: aString(),
+                queryID: aString(),
```
rasmus-storjohann-PG, line 21: So when you modify all these tests to make them pass, you should also think "do I need to add new tests to cover these new properties I'm adding?"

src/validation/search/schema.ts
```diff
@@ -66,6 +66,7 @@ export const serviceSearchItem = {
         'phone_numbers': phoneNumbers,
         '_geoloc': geoLocation,
         'email': { 'type': 'string' },
+        'objectID': { 'type': 'string' },
```
rasmus-storjohann-PG, line 69: No queryId here?

src/validation/search/__tests__/to_human_service_data.tests.ts
```diff
@@ -322,6 +340,8 @@ describe('Adapting search result to service data', () => {
...
```
rasmus-storjohann-PG, line 344: Really annoying: inCamelCase, each upper case letter indicates the start of a new word. So `objectID` is wrong, `objectId` is correct. With F2 in VS Code you should be able to rename these quickly.

rasmus-storjohann-PG (changes requested):

src/validation/search/types.ts
```diff
@@ -38,7 +38,7 @@ export interface SearchServiceData {
     readonly objectID: string;
     readonly phone_numbers?: ReadonlyArray<PhoneNumber>;
     readonly _geoloc: AlgoliaLatLong;
-    readonly queryID: string;
+    readonly queryID?: string;
```
rasmus-storjohann-PG, line 41: This is a bit of a short cut. With the field optional, you need to confirm that the field is populated whenever you access it throughout the code base. If you had two types, one with and one without the field, you add a bit of complexity here, but you can remove the checks everywhere else in the code base. I'm not saying that the latter is always the correct way to go. If you only care about the field in a small number of places, checking the field is a small price to pay. But keep the consern in mind. Typically, if you go with the current solution and later discover that this causes bugs or slows you down in making changes to the code, then refactoring this to use the latter approach can be a good thing.

src/components/search/search_results_component.tsx
```diff
@@ -161,16 +160,15 @@ interface SearchHitProps {
...
 }
 
 const renderSearchHit = R.curry((props: SearchHitProps, itemInfo: ListRenderItemInfo<SearchServiceData>): JSX.Element => {
     const item: SearchServiceData = itemInfo.item;
     const service: HumanServiceData = toHumanServiceData(item, props.bookmarkedServicesIds, props.reviewedServicesIds);
     const onPress = (): void => {
         props.saveServiceToMap(service);
-        props.openServiceDetail(service);
-        sendSearchAnalytics(service.objectID, item.queryID, itemInfo.index);
+        props.openServiceDetail(service, item.queryID, itemInfo.index);
```
rasmus-storjohann-PG, line 171: This is a good change: The idea of redux is to create *one* action from an event, and make sure the action contains *all* the information needed for reducers and sagas to do their work.

src/sagas/services.ts
```diff
@@ -25,6 +26,10 @@ export function* watchUpdateServicesForOrganization(): IterableIterator<ForkEffe
     yield takeLatest(constants.OPEN_ORGANIZATION, updateServicesForOrganization);
 }
 
+export function* watchOpenServiceDetail(): IterableIterator<ForkEffect> {
+    yield takeLatest(constants.OPEN_SERVICE_DETAIL, sendSearchAnalytics);
+}
```
rasmus-storjohann-PG, line 31: This should be pulled into its own file, one watch statement per saga file. Arguably there should be a better naming convention for the saga files to make this mapping clear...

src/components/bookmarks/bookmarks_connected_component.tsx
```diff
@@ -41,7 +41,7 @@ const mapDispatchToProps = (dispatch: Dispatch<Actions>): BookmarkActions => ({
     unbookmarkTopic: (topicId: Id): UnbookmarkTopicAction => dispatch(unbookmarkTopic(topicId)),
     bookmarkService: (service: HumanServiceData): BookmarkServiceAction => dispatch(bookmarkService(service)),
     unbookmarkService: (service: HumanServiceData): UnbookmarkServiceAction => dispatch(unbookmarkService(service)),
-    openServiceDetail: (service: HumanServiceData): OpenServiceAction => dispatch(openServiceDetail(service)),
+    openServiceDetail: (service: HumanServiceData): OpenServiceAction => dispatch(openServiceDetail(service, undefined, undefined)),
```
rasmus-storjohann-PG, line 44: Not quite done here, I think?

src/components/services/service_list_connected_component.tsx
```diff
@@ -49,7 +49,7 @@ const mapDispatchToProps = (dispatch: Dispatch<Action>): ServiceListActions => (
     unbookmarkService: (service: HumanServiceData): actions.UnbookmarkServiceAction => dispatch(actions.unbookmarkService(service)),
     hidePartialLocalizationMessage: (): HidePartialLocalizationMessageAction => dispatch(hidePartialLocalizationMessage()),
     openHeaderMenu: (): OpenHeaderMenuAction => dispatch(openHeaderMenu()),
-    openServiceDetail: (service: HumanServiceData): actions.OpenServiceAction => dispatch(actions.openServiceDetail(service)),
+    openServiceDetail: (service: HumanServiceData): actions.OpenServiceAction => dispatch(actions.openServiceDetail(service, undefined, undefined)),
```
rasmus-storjohann-PG, line 52: These `undefined` fields should be fixable, right? All service records come from Algolia, so they should all have the necessary ids on them to populate these fields....

rasmus-storjohann-PG (changes requested):

PR #1824, opened 2022-03-17
Feature/issue 1134 add review modal

brandon-PG:


src/stores/__tests__/user_profiles.test.ts
```diff
@@ -208,4 +208,19 @@ describe('user profile reducer', () => {
             expect(newStore.bookmarkBanner).toBe(BookmarkBanner.RegionHasNotBeenChanged);
         });
     });
+
+    describe('the timesAppStarted number', () => {
+
+        test('is increased by application started action', () => {
+            const oldStore = new UserProfileBuilder().build();
+            const newStore = reducer(oldStore, applicationStarted());
+            expect(newStore.timesAppStarted).toBe(oldStore.timesAppStarted + 1);
```
rasmus-storjohann-PG, line 217: JS doesn't have an integer type, so all numbers are floating point, which are tricky to compare for equality

src/selectors/user_profile/select_times_app_started.ts
```diff
@@ -0,0 +1,5 @@
+import { Store } from '../../stores';
+
+export const selectTimesAppStarted = (appStore: Store): number => {
+    return appStore.userProfile.timesAppStarted;
+};
```
rasmus-storjohann-PG, line 5: This is not our coding style for one line functions

src/stores/__tests__/user_profiles.test.ts
```diff
@@ -222,5 +222,15 @@ describe('user profile reducer', () => {
...
+
+        it('is loaded from persisted data', () => {
+            const oldStore = new UserProfileBuilder().build();
+            const persistedDataBuilder = new PersistedDataBuilder().
+                withTimesAppStarted(aNumber()).
+                build();
+            const presistanceAction = DataPersistence.loadSuccess(persistedDataBuilder);
+            const newStore = reducer(oldStore, presistanceAction);
+            expect(newStore.timesAppStarted).toBe(persistedDataBuilder.timesAppStarted);
+        });
```
rasmus-storjohann-PG, line 234: Nice

src/components/main/main_connected_component.ts
```diff
@@ -55,6 +56,7 @@ const mapStateToProps = (store: Store, ownProps: RouterProps): Props => ({
     isCreatorModalVisible: selectIsCreatorModalVisible(store),
     feedbackScreen: selectFeedbackScreen(store),
     theme: selectUxTheme(store),
+    showReviewModal: selectTimesAppStarted(store) === 2,
```
rasmus-storjohann-PG, line 59: I don't want that kind of low level detail in a connected component file, it should be high level "connecting" only. Make a selector that computes whether to show the modal or not. This selector can replace the existing `selectTimesAppStarted()`

src/stores/__tests__/user_profiles.test.ts
```diff
@@ -217,10 +217,10 @@ describe('user profile reducer', () => {
             expect(newStore.timesAppStarted).toBe(oldStore.timesAppStarted + 1);
         });
 
-        test('is set to 1 by clear all user data action', () => {
+        test('is set to 0 by clear all user data action', () => {
             const oldStore = new UserProfileBuilder().build();
             const newStore = reducer(oldStore, clearAllUserData());
-            expect(newStore.timesAppStarted).toBe(1);
+            expect(newStore.timesAppStarted).toBe(0);
```
rasmus-storjohann-PG, line 223: that makes more sense

src/components/main/main_component.tsx
```diff
@@ -30,6 +30,7 @@ import { RegionDrawerConnectedComponent } from '../header_menu/region_drawer_con
 import { BackHandler, SafeAreaView } from 'react-native';
 import { UxTheme } from '../../application/ux_themes';
 import { ApplicationStartedAction } from '../../stores/user_profile';
+import { hasAction, requestReview } from 'expo-store-review';
```
rasmus-storjohann-PG, line 33: Odd names for these functions, you can make better names at the point of import : `import { foo as bar, bz as fuz } from 'whatever';`

src/components/main/main_component.tsx
```diff
@@ -116,6 +120,17 @@ export const MainComponent = (props: Props): JSX.Element => {
         return (): void => subscription.remove();
     }, []);
 
+    useEffect((): void => {
+        sendReviewRequest();
+        props.applicationStarted();
+    }, []);
```
rasmus-storjohann-PG, line 126: Do you think this is consistent with the single responsibility principle? Let's talk

rasmus-storjohann-PG (changes requested):
Very nice work

src/components/main/main_connected_component.ts
brandon-PG: selectTimesAppStarted in needed in `select_user_data_for_local_persistence.ts` when saving the value to persistent data. I either create another selector or keep this logic in the connected component.

PR #1889, opened 2022-05-16
Feature/issue 1882 rename explore to chapters

rasmus-storjohann-PG:


brandon-PG:
the topic path strings in 'routing.test.ts' line 406-414 must be changed the reflect the new topic detail path. `/task/:topicId` -> `/topics/:topicId`

Issue #1850, opened 2022-04-15
When app is in Arabic and all memory is cleared, it ends up in English in LeftToRight mode


rasmus-storjohann-PG:
Start the app. Change the language to Arabic. Clear all data off the app. Observe the app onboarding flow. Expect the app to be in English in LTR mode. Observe the app is in English in RTL mode.

If the app is in RTL mode and all memory is cleared from the app, it needs to reboot to be in the correct LTR mode. A simple solution is to always reboot the app when the memory is cleared.

### Acceptance Criteria 
- [ ]  The app comes up in the correct English LTR mode when clearing all memory off the app

PR #1896, opened 2022-05-17
Issue #1850 create saga to clear user data

shivangi-PG:


src/application/locale_effects.ts
```diff
@@ -32,3 +32,7 @@ export async function saveCurrentLocaleCode(code: string): Promise<void> {
 export async function loadCurrentLocaleCode(): Promise<string> {
     return await AsyncStorage.getItem(PREFERENCES_LOCALE_CODE);
 }
+
+export async function removeLocaleAsync(): Promise<void> {
+    await AsyncStorage.removeItem(PREFERENCES_LOCALE_CODE);
+}
```
rasmus-storjohann-PG, line 38: There should be a `return` here ... the function signature shows a return type, so there must be a return.

rasmus-storjohann-PG (changes requested):
Just one small change, and let's also discuss how to make commits smaller and more self contained.

rasmus-storjohann-PG:
I see some issues when QAing this change, including the app restarting twice. Need to probably remove the code from other reducers that handle this action, since these are now redundant. That makes for a more extensive, and hence high risk change that could introduce bugs. We'll wait until the production release is done, then we will revisit this change and look to merge it then, with modifications.

src/stores/locale/index.ts
```diff
@@ -34,8 +34,6 @@ export const reducer = (store: LocaleStore = buildDefaultStore(), action?: actio
             return { ...store, loading: false };
         case constants.RESET_LOCALE:
             // intended fall through
-        case constants.CLEAR_ALL_USER_DATA:
-            return { ...store, code: undefined };
```
rasmus-storjohann-PG, line 38: Fix fall throughs

shivangi-PG: Fixed in [Issue](https://github.com/pg-irc/pathways-frontend/pull/1896/commits/47c691656f78062bcac87dd0f004aed5cb83cef6) https://github.com/pg-irc/pathways-frontend/issues/1850 [remove fall through from switch statements](https://github.com/pg-irc/pathways-frontend/pull/1896/commits/47c691656f78062bcac87dd0f004aed5cb83cef6)

src/sagas/user_data.ts
```diff
@@ -97,6 +97,10 @@ export async function loadRouteAsync(): Promise<string> {
     return await AsyncStorage.getItem(ROUTE_STORAGE_KEY);
```
rasmus-storjohann-PG, line 97: Need to remove the ROUTE_STORAGE_KEY data item too

src/sagas/user_data.ts
```diff
@@ -150,3 +154,17 @@ const setDefaultReviewProperty = (servicesData: any): ServiceMap => {
...
+
+export function* watchClearAllUserData(): IterableIterator<ForkEffect> {
+    yield takeLatest(constants.CLEAR_ALL_USER_DATA, clearAppData);
+};
+
+export function* clearAppData(): IterableIterator<CallEffect > { 
+    try {
+        yield call(removeUserDataAsync);
+        yield call(removeLocaleAsync);
+        yield call(setTextDirection, undefined);
```
rasmus-storjohann-PG, line 166: set text direction here has the side effect of rebooting the app? Might want to rename that function to make that clear

src/stores/user_experience/index.ts
```diff
@@ -65,13 +65,21 @@ export const reducer = (store: UserExperienceStore = buildDefaultStore(), action
...
+            return headerMenuReducer(store, action);
         case constants.CLOSE_DISCLAIMER_MODAL:
+            return headerMenuReducer(store, action);
         case constants.OPEN_DISCLAIMER_MODAL:
+            return headerMenuReducer(store, action);
         case constants.CLOSE_CREATOR_MODAL:
+            return headerMenuReducer(store, action);
         case constants.OPEN_CREATOR_MODAL:
+            return headerMenuReducer(store, action);
         case constants.SAVE_LOCALE_REQUEST:
```
rasmus-storjohann-PG, line 83: hmmmm, so every rule has exceptions, and this is a place where I thought fall through actually makes the code easier to understand. Because there are so many cases falling through, it is much less likely to be misssed, and you would also have an easier time understanding that all these cases do the same thing.

src/stores/__tests__/services.test.ts
```diff
@@ -250,18 +249,6 @@ describe('services reducer', () => {
...
-        const store = buildNormalizedServices([serviceBuilder], []);
-        const action: ClearAllUserDataAction = {
-            type: constants.CLEAR_ALL_USER_DATA,
-        };
-        const storeState = reducer(store, action);
-        it('removes services from service map', () => {
-            expect(storeState.services).toEqual({});
-        });
-    });
-
```
rasmus-storjohann-PG, line 264: Note that some of these commits do less than one thing ... in particular, you made changes to the code (removing CASE statemtns) and to the tests (as seen here). These changes are related, since changing the behaviour of the code would make the tests fail ... and all tests should ideally pass at every commit. So in these cases, the changes to the code and the tests should go in the same commit.

rasmus-storjohann-PG (changes requested):

rasmus-storjohann-PG:
I'll need to spend a bit of time testing this change, doesn't look like the behaviour is quite right. But the code changes are definitely moving us in the right direction

src/sagas/user_data.ts
shivangi-PG: resolved in
[Issue](https://github.com/pg-irc/pathways-frontend/pull/1896/commits/0906acd1b71536fb155829caf23ec806b73e0391) https://github.com/pg-irc/pathways-frontend/issues/1850 [create function removeRouteAsync and use in clearAppData](https://github.com/pg-irc/pathways-frontend/pull/1896/commits/0906acd1b71536fb155829caf23ec806b73e0391)

src/sagas/user_data.ts
shivangi-PG: changed in [Issue](https://github.com/pg-irc/pathways-frontend/pull/1896/commits/4100e0df98821729f2b9cd02ade59e20e50fc6a7) https://github.com/pg-irc/pathways-frontend/issues/1850 [change function name](https://github.com/pg-irc/pathways-frontend/pull/1896/commits/4100e0df98821729f2b9cd02ade59e20e50fc6a7)

